PR #11, opened 2017-01-05
[RFR] dev 1 : unit test exisiting code

mozhgan-kch:
- [x]  Compile the code write some tests that check the model objects (model folder) are correct (MKC)
- [x]  Tests for population objects. Check that the data is correct. (MKC)
- [x] Test the GPU memory mapping. Fix any issues with the memory mapping to CUDA. (MKC and PR)
- [x] Develop code for function execution order. Remove any behaviour from agent functions so that they just printf. We should be able to see that the agent functions are correctly called by the layers. (MKC)
- [x] Add to the above read and writing of agent variables. Develop a test to check that a simple agent function that adds a constant to each agent variable works correctly. (MKC)
- [x] type checking for get/set variable in FLAMEGPU api class . The checking should be done as part of curve
- [x] Simultaneous launch of functions using curve namespace
      To test this, we need to have a new model that has two agents
- [x] a new model that has multiple agents to test above
- [x] fix the problem with CUDA boost test that is not supported on VS when compiled with NVCC to solve #24
- [x] bind function, agent, var names , commit 4e608eb


mondus:
Just realised. We cant use a device singleton (https://devtalk.nvidia.com/default/topic/573203/why-is-there-no-support-for-static-class-members-/?offset=4). The FLAME_API class will have to be a member of CUDAModel.h.

PR #93, opened 2019-10-25
General CMake Improvements

### LINKED ISSUE #96: Remove explicit visual studio files
Switching to Cmake for windows and linux builds removes duplication of project configuration for development. 

Appveyor will also need updating to account for this. 

@mondus @mozhgan-kch are you both happy with this?

### LINKED ISSUE #100: Cmake General
Addressed in Pull Request: #93 

New/Current Cmake rules set up filters for Visual studio projects in the examples folder.
Examples split into header or source, depending on file type. This might want modifying so that include (i.e. public library headers) go somewhere else. 

For the library, filters just mirror disk. 

This may need further adjustment when the library headers are split into public / private headers (by moving private headers to the `src` directory. 

------------------
This issue replaces #91, as a general CMake thing
Work can be found in #93 pull request.

Ideally want all of this solved before Monday 4th Nov
TODO:
+ [x] ~Adjust CMake to auto detect files in given directories~
+ [x] Ensure filtering is as intended e.g. `include` vs `header`
+ [x] Delete all `.vcxproj`/`.sln` (#103)
+ [x] Relocate `.lib` build path
+ [x] Update AppVeyor to use CMake (#103)
+ [x] Does Paul want example binaries to build to same dir? ("just need to add `../../` to line `20` in each example/test `CMakeLists.txt`)
+ [x] Update readme to only talk about cmake

ptheywood:
> Having reviewed 'Files changed', I'm comfortable with this pull request being squashed, so long as the full changelog is included within the message. The changes are quite clear from reading the changed files. 
> 
> This would correct the fact that half the commits fail to mention CMake.
>
> I think the compile warnings branch will survive (it's a branch of this). If not, I can cherrypick it into a new branch and fix conflicts. It mostly doesn't touch CMake (it only adds the forced warning stuff), so conflicts shouldn't be too bad.
> 
> Rob

This pull request significantly improves CMake, such that it works appropriately for both Visual Studio and g++. Hence, the `.sln` and `.vcxproj` files have been removed. Intuitive messages and options for users of both `cmake` and `cmake-gui` have been added. The readme has been updated to explain how to use CMake, on either platform, for those unfamiliar with it.

Closes #100 
Closes #96 

In future, when we have working examples, CMake can also be used to automatically configure the debugging settings (e.g. default working directory/args) within Visual Studio.

**Full Changelog:**
* Updates minimum cmake version to 3.12 (there was a comment saying <3.12 works with linker issues)
* Updates readme to provide platform agnostic `cmake` guidance.
* Adds delimiting message when individual projects are configured
* Unify warning level across C/CXX/CUDA
* Corrects handling of build configs between single/multi config generators.
* Reduces repeated messages when `common.cmake` is used by multiple targets.
* Set output dir OS specific (e.g. `linux-x64`, `windows-x64`)
* Fixes `SMS` variable to work via `cmake-gui`
* Disables generation of installation scripts
  * Hide `CMAKE_INSTALL_PREFIX` from `cmake-gui`
* Relocates binaries to within the build directory
  * `root-build-dir/bin/<os>/<config>`
  * Tests goto subdirectory `tests` of this path
  * `flamegpu2.lib` goes to `root-build-dir/lib/<os>/<config>`
  * This behaviour is mimicked if an example's `cmakelists.txt` is instead used.
* Renames `flamegpu2` sub-build dir from `src` to `FLAMEGPU2`
* Visual Studio:
  * Filters:
    * Builds visual studio filter hierarchy, mirroring directory structure
    * Groups targets into different filters
  * Reduce MSVC warning to W4
    * Suppress `unreferenced local function has been removed` warnings (on Windows)
  * Remove explicit Visual studio files and directories in favour of `cmake`
  * Sets startup project
    * Defaults to the top level target
    * For the root `CMakeLists.txt`, it sets to `flamegpu2` target.
* Lint:
  * Rename `lint` target to `all_lint`
  * Exclude lint targets from `ALL_BUILD` target
  * Bugfix: Only first file from list was being passed to cpplint
  * Prevents files in `external` dir from being linted
  * Adds check for `cpplint` to CMake
    * Gives warning if unavailable
    * Uses this path for calling cpplint (allows user to custom specify path if they so desire)
    * Doesn't create lint targets if cpplint not available
* CI:
  * Updates Travis to use recent Ubuntu, GCC, CUDA and CMake. 
  * Update Appveyor to use `cmake` (and CUDA 10)
  * Builds docs on Travis, but not appveyor *
* Docs:
  * Builds documentation using doxygen through cmake
  * Doxygen is an optional dependency - the docs target will only be generated if doxygen is found. 
* Misc:
  * Typos in comments
  * *Accidentally rebased in 9838db7 at some point 🤷‍♂️*


Robadob:
I can test it tonight/over the weekend or something.

Robadob:
Testing this on my office machine, [this](https://gist.github.com/Robadob/a2e3f446afed79a12dde660c3b2701d9) is the output from build sln (debug), all succeeded.

Also noticed that the binaries directory is set to `linux-x64` (intermediate files goto `x64`)

ptheywood:
https://docs.microsoft.com/en-us/dotnet/visual-basic/reference/command-line-compiler/nowarn

Looks like we could suppress individual warngins in msvc through this comand line option. 

Warnings we probbaly want to suppress are: 
+ warning C4505: '__cudaUnregisterBinaryUtil': unreferenced local function has been removed

Unsure about
+ warning C4100: 'abort': unreferenced formal parameter

Warnings which should be kept on (and resolved)
+ warning C4267: 'argument': conversion from 'size_t' to 'unsigned int', possible loss of data
+ warning C4458: declaration of 'name' hides class member
+ warning C4715: 'ReaderFactory::createReader': not all control paths return a value
+ warning C4189: 'd_ptr': local variable is initialized but not referenced

Robadob:
> Unsure about
> * warning C4100: 'abort': unreferenced formal parameter

That can be corrected, it's the fact you're never using the `abort` arg on the cuda error checking fn due to some changes.

Next commit addresses C4505, by which point I think I'm happy with CMAKE windows.
Delete vcx projs?

Robadob:
CI currently failing because cmake.org is/was down.
```
# Install redent CMAKE
CMAKE_SH=https://cmake.org/files/v3.12/cmake-3.12.2-Linux-x86_64.sh
wget -O cmake.sh $CMAKE_SH
--2019-10-30 15:13:57--  https://cmake.org/files/v3.12/cmake-3.12.2-Linux-x86_64.sh
Resolving cmake.org (cmake.org)... 66.194.253.19
Connecting to cmake.org (cmake.org)|66.194.253.19|:443... failed: No route to host.
The command "sudo -E ./scripts/travis_install.sh" failed and exited with 4 during .
```

image

ptheywood:
Cmake is available on pip - which may be easier than wgetting from cmake.org?

https://pypi.org/project/cmake/

Robadob:
> Cmake is available on pip - which may be easier than wgetting from cmake.org?
> 
> The pip package is maintained by cmake developer(s)
> 
> https://pypi.org/project/cmake/

Upto you whether you want to change that, I'm not upto speed on configuring CI

cmake/common.cmake
```diff
@@ -189,7 +189,7 @@ else()
     message(WARNING "cpplint script: NOT FOUND! "
                     "Lint projects will not be generated"
                     "Please install cpplint as described on https://pypi.python.org/pypi/cpplint. "
-			        "In most cases command '(sudo )pip install cpplint' should be sufficient.")
+			        "In most cases command '(sudo) pip install cpplint' should be sufficient.")
```
ptheywood: pip install --user cpplint is probably recommended over `sudo pip install cpplint`

ptheywood: Resolved by 471f300

PR #105, opened 2019-11-01
Compilation Warning Fixes

### LINKED ISSUE #101: Fix compiler warnings
Currently there are many compiler warnings on both windows and linux. We should try to fix these. 

Addressed in Pull Request: #105

Robadob:
This is a branch of msvc cmake one, but will be merged after, hence pull req to base message branch.

Closes #101

**Full Changelog:**
* Fixes all warnings in VS2015 and Travis.
* Adds CMake option to enable warnings as errors
  * This has been enabled on both CIs
* Future support for NVCC link stage warnings as errors
  * CMake doesn't yet support this, but it's setup with the [variable](https://gitlab.kitware.com/cmake/cmake/issues/18265#note_600354) proposed by the developer  implementing it (see #106)
* Turns on some NVCC options for even more aggressive warnings
  * Fixed order of all initialiser lists
* **Fixes bug in** `AgentFunctionDescription::setIntialState()` 

**Warning Note (unused arg):**
Unused args should not be present in the code. You should comment out the args identifier if it is unused to confirm that it's intentional. Doing this will fix the warning, allowing CI to pass, without needing to physically remove the unused arg.

e.g.
```c
void foo(int /*bar*/)
{
    //do something
}
```

**Warning Note (arg hides member var):**
This is bad practice, as it can easily lead to misusing vars due to typos, e.g. the bug I fixed as part of this pull request.

I've addressed these by naming args `_identifier` rather than `identifier`, alternate options would be to enforce the practice whereby member variables are all written `mIdentifier` or `m_identifier`, this is common practice in some environments for the same reason (with secondary benefit of allowing accessors to use the variables actual name).

**Warning Note (implicit down conversion):**
There was a warning due to mixing `unsigned int` and `size_t` within `MessageList` or similar. I have replaced this with an internal `typedef size_type`, this should be used throughout for anything to do with length or indexing a data structure. Ideally all data structures should inherit their `size_type` from a central collections `allocator` template as is the case with `std::vector`, `std:map` etc, however going that far is a 'maybe'.

Avoid using `size_t`  as `size_t` if you are not referring to memory size (bytes), in reality our `size_type` has no reason to be greater than a `uint32`, flamegpu has no vectors of `char`.

**Note:**
There are still a large number of (the same) warnings produced by the nvcc linker stage under visual studio. Due to complexities of nvcc build process, these warnings do not however prevent compilation when warnings are upgraded to errors. There is a way to suppress this warning only during the linker stage (as its useful elsewhere), however CMake does not yet support it (hence the future support item in changelog).


ptheywood:
As of 231b2c5

```
make lint_flamegpu2 2>&1 | tee lint.log

grep "^/" lint.log | cut -d"[" -f2 | cut -d"]" -f1 | sort | uniq -c
     39 build/header_guard
      1 build/include
     48 build/include_order
      3 build/include_subdir
     10 build/include_what_you_use
      5 build/namespaces
     13 readability/braces
     23 readability/casting
      8 readability/todo
     13 runtime/explicit
     22 runtime/references
    106 whitespace/blank_line
    309 whitespace/braces
      2 whitespace/comma
    327 whitespace/comments
     24 whitespace/end_of_line
      1 whitespace/forcolon
     50 whitespace/indent
      8 whitespace/line_length
      6 whitespace/newline
      7 whitespace/operators
      8 whitespace/parens
      1 whitespace/semicolon
```

PR #118, opened 2019-11-06
Updates gitignore.

### LINKED ISSUE #99: .gitignore rules
.gitignore needs configuring to ignore cmake generated files on both windows and linux. 

Currently if users generate in a `build` directory all files will be ignored, but not if the user builds in a different directory. 

+ https://github.com/github/gitignore/blob/master/CMake.gitignore
+ https://github.com/github/gitignore/blob/master/VisualStudio.gitignore

Robadob:
Updates github/visual_studio, adds github/cmake

Includes current commit id, so changes can be tracked if necessary in future.

Closes #99


ptheywood:
As of f3147b under linux, building in an alternative directory to `build` (i.e. `build2`) does not result in any un-ignored files. 

This should also be tested and adjusted under windows (@Robadob) and then this will be ready to merge.

PR #119, opened 2019-11-06
Suggested fix for segfault in simple_example.

### LINKED ISSUE #113: simple_example segfault
The simple example currently (c9932b6) segfaults under linux:

```
./bin/linux-x64/debug/simple_example 
step: 0
inp msg name: location
Segmentation fault (core dumped)

```

Robadob:
Closes #113 

Implicitly calls `CUDAMessage::setInitialMessageList()` at construction of `CUDAMessage`.

Messages are not filled on the host, so initialisation should be implicit. Only concern is whether the list size is allocated to match agent size, I didn't check that far.

Comments out `CUDAAgentModel::setMessageData(MessageDescription&)` as it is now redundant.

It should be removed before this is merged (hence draft PR).

Also adds a check/exception to the offending method incase pointer is empty.

Now executes to exit without segfault (on Windows).

### DIFF
--- a/examples/mas/src/main.cu
+++ b/examples/mas/src/main.cu
@@ -223,7 +223,7 @@ int main(int argc, char* argv[]) {
     cuda_model.setInitialPopulationData(population1);
     cuda_model.setInitialPopulationData(population2);
 
-    cuda_model.setMessageData(location1_message);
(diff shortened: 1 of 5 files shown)


Robadob:
This is ready for merge once Travis passes.

PR #125, opened 2019-11-07
Feature: SubModels

### LINKED ISSUE #81: Feature: SubModels - Iterative Blocks
Primage model in two instances has a need to run multiple agent functions for an unknown number of iterations until an optimisation has reached an end point.

It was agreed that this could be modelled as a sub-model, which inherits from a parents model, executing until an exit condition is satisfied.

This would require:
* agent variable mappings
    * Optional read only access
* Read only message access
* Environment bound mappings? (spatial messages need env bounds for best case perf)
    * This could be performed by doing scaling inside spatial message handling?
* Optional override of central model fn?
* Exit condition

```xml
<submodels>
    <submodel>
        <name>BondingModel</name>
        <modelfile>bonding.xml</modelfile>
        <mapping>
            <environmentMapping>
                <!-- Perhaps an env variable needs to be shared between models. -->
            </environmentMapping>
            <agentMapping>
                <name>Boid</name>
                <subName>BondingAgent</subName>
                <variable>
                    <name>x</name>
                    <subName>x</subName>
                <variable>
                <variable>
                    <name>y</name>
                    <subName>y</subName>
                <variable>
                <variable>
                    <name>z</name>
                    <subName>z</subName>
                <variable>
                <!--Agent force variables can be wholly contained within bonding model.-->
            </agentMapping>
            <messageMapping>
                <!--Perhaps read only access to some messages is required-->
            </messageMapping>
        <mapping>
        <bounds>
            <!--Bonding submodel has its own spatial partitioning, somehow we need to configure the env bounds (and radius)-->
            <fromMessage>
                <name>location_message</name>
            </fromMessage>
        <bounds>
        <function>
            <!--Optional replacement for bonding condition, e.g. protein level check + rng-->
        </function>
        <exitCondition>
            <reduction>
                <agentVariable>
                    <agent>BondingAgent</agent>
                    <variable>ForceMagintude</variable>
                </agentVariable>
                <operator>+</operator>
                <condition>&lt;0.01</condition>
            </reduction>
        </exitCondition>
    </submodel>
</submodels>
....
<layers>
    <layer>
        <gpu:layerFunction>
            <name>doSomething</name>
        </gpu:layerFunction>
    </layer>
    <layer>
        <subModel>
            <name>BondingModel</name>
        <subModel>
    </layer>
    <layer>
        <gpu:layerFunction>
            <name>doSomethingElse</name>
        </gpu:layerFunction>
    </layer>
</layers>
```

### LINKED ISSUE #238: CUDAAgent refactor
As discussed on slack, considering a big refactor to cleanup the internals of `CUDAAgent`/`CUDAAgentStateList`, and better align them with SubModels.

Strategy would be for all agents/states which are mapped/bound, to share a common 'Fat' Agent or Statelist. All operations are handled via their 'Fat' object, by passing it a selector, which means only a particular view of the fat agent is accessed.

I started working on what it might look like here, but it needs quite a bit more work: https://gist.github.com/Robadob/3258809fdbf3aa1001427361e4d9c3ec

### LINKED ISSUE #320: CUB Error Checking
All (most?) CUB functions return `cudaError_t`, which should be checked.

This is currently not the case, potentially leading to the errors being reported at a misleading location.

We should make sure that all `cub::` methods are wrapped in `gpuErrchk` macros.

I've not made a PR to fix this where I noticed (and used) this to identify the hardware specific issue in the swig branch, as I believe CUDAAgent has/is being refactored ?.

Robadob:
Closes #81 (SubModels)
Closes #320 (CUDA error check cub inside CUDAScatter)
Closes #238 (CUDAAgent refactor)

SubModels are a new feature, not present in FLAMEGPU1. They are intended to replace FLAMEGPU1s global conditions, providing a more intuitive interface for implementing iterative operations within models.

A submodel must exist by itself within a layer. When this layer executes, the submodel executes as though `CUDAAgentModel::simulate()` has been called, exiting when the exit condition returns `EXIT`. 

Agents and environment properties can be mapped between model and submodel, with the submodel instance of an agent able to carry extra variables not available within the main model. These variables will be maintained between submodel executions.

Submodels may also contain their own agents, messages and environment properties which are not mapped to a parent model, these will all be cleared between each submodel execution. Unmapped environment properties will return to their initial value.

Multiple submodels can be nested to form a complex hierarchy, the only limitation is that a model may not exist within it's own submodel hierarchy.

These changes have been tested on Linux (by Pete) and Windows.

An example model will be provided in a future PR.


Robadob:
SubAgent behaviour (besides state transitions) is all tested and working. Before testing transitions I am refactoring `CUDAAgent` in a separate branch. Due to the size of this refactor it's not yet been pushed to remote.

ptheywood:
All 550 tests pass under linux (CUDA 10.2, GCC 8, GTX 1070) (1344d48)

```
[==========] 550 tests from 43 test suites ran. (58134 ms total)
[  PASSED  ] 550 tests.
```

Issue #136, opened 2019-11-13
CUDAAgentModel Should Notice Dependency Changes.


Robadob:
**Issue:**
```c++
    const unsigned int AGENT_COUNT = 5;
    ModelDescription model("model");
    AgentDescription agent("agent");
    AgentPopulation population(agent, AGENT_COUNT);
    Simulation simulation(model);
    SimulationLayer hostfn_layer(simulation, "hostfn_layer");
    CUDAAgentModel cuda_model(ms->model);
    hostfn_layer.addHostFunction(&host_function);
    simulation.addSimulationLayer(hostfn_layer);
    for (unsigned int i = 0; i < AGENT_COUNT; i++) {
        AgentInstance instance = population.getNextInstance();
    }
    model.addAgent(agent);
    cuda_model.setInitialPopulationData(ms->population);
    cuda_model.simulate(ms->simulation); //Exception InvalidCudaAgentMapSize("CUDA agent map size is zero")!
```
If `CUDAAgentModel` is constructed, before all calls to `ModelDescription::addAgent()` have occurred, no agents are allocated within `CUDAAgentModel`, so an exception is thrown when `CUDAAgentModel::simulate()` is called.

**Preferred Behaviour:**
`CUDAAgentModel` notices that it's dependency `ModelDescription` has been changed and automatically reallocates itself/memories.

**Alternate (less desirable) Behaviour:**
* `CUDAAgentModel` performs the allocation the first time `simulate()` is called.
* The exception is changed to stating that `ModelDescription` has changed.

**Fix:**
Description classes have a hash function, or store a timestamp when changed, this can then be tracked by anything dependent on them, to automatically trigger updates.

**Concern:**
If `CUDAAgentModel::setInitialPopulationData()` has already been called, reallocation might necessitate losing this data. How should this be handled?

Robadob:
Same applies to `AgentPopulation` when variables are added to an `Agent` after population has been created.

Robadob:
The description refactor merge has changed behaviour around this, but there is still no explicit blocking.

Creating a `Simulation` now takes a copy of the `ModelData` hierarchy, so the original can still be changed as it is nolonger attached.

Unclear if this needs further work.

ptheywood:
No longer relevant due to copy consruction from refactor?

Issue #56, opened 2018-10-01
Example: Circles Model


ptheywood:
Implement the Circles model. In flame gpu 1 we have the following 4 versions:

+ [ ] CirclesBruteForce_double
+ [ ] CirclesBruteForce_float
+ [ ] CirclesPartitioning_double
+ [ ] CirclesPartitioning_float

Robadob:
When this model is completed the `CMakeLists.txt` should be updated, so that it autoconfigures the default debugger settings for Visual Studio (e.g. working directory, args).

This is managed similarly to how `VS_STARTUP_PROJECT` is used to set startup project, however it's at a target level, rather than directory level.

```
VS_CONFIGURATION_TYPE
VS_DEBUGGER_COMMAND
VS_DEBUGGER_COMMAND_ARGUMENTS
VS_DEBUGGER_ENVIRONMENT
VS_DEBUGGER_WORKING_DIRECTORY
```

More info: [CMake properties](https://cmake.org/cmake/help/latest/manual/cmake-properties.7.html), [set_property()](https://cmake.org/cmake/help/latest/command/set_property.html)

Robadob:
With brute force messages working, I've started working on this in branch `circles_example`.

I'm simply doing a generic float brute force version (I see no reason in having a double clone, can typedef float to double if really necessary.

a5f4082 Provides the existing FGPU1 circles model. I will then follow that commit with the updated circles model from my thesis.

Robadob:
> When this model is completed the `CMakeLists.txt` should be updated, so that it autoconfigures the default debugger settings for Visual Studio (e.g. working directory, args).
> 
> This is managed similarly to how `VS_STARTUP_PROJECT` is used to set startup project, however it's at a target level, rather than directory level.
> 
> ```
> VS_CONFIGURATION_TYPE
> VS_DEBUGGER_COMMAND
> VS_DEBUGGER_COMMAND_ARGUMENTS
> VS_DEBUGGER_ENVIRONMENT
> VS_DEBUGGER_WORKING_DIRECTORY
> ```
> 
> More info: [CMake properties](https://cmake.org/cmake/help/latest/manual/cmake-properties.7.html), [set_property()](https://cmake.org/cmake/help/latest/command/set_property.html)

d983ed4 addresses this

Robadob:
As of 51fcd39 the model works as intended (tested by visualising with the PRIMAGE binary format visualiser).

The question is now how best to implement validation.

[...]

PR #165, opened 2019-12-16
Description refactor

### LINKED ISSUE #49: Feature: Agent Death
Death of agents from device agent functions.

### LINKED ISSUE #124: Inconsistency: getMemoryMap() vs getVariableMap()
These serve the same purpose for Agents and Messages, they should share the same name.

```c++
AgentDescription::getMemoryMap()
MessageDescription::getVariableMap()
```

I suggest updating `Agent`'s use of `MemoryMap` to `VaraibleMap`.

This was probably not done in the first place, because the typedefs would conflict as they are declared outside the class, move them inside!!!!

### LINKED ISSUE #132: xDescription Refactor
**Started working on this in branch** `description_refactor`

Proposition to refactor the description classes:

Description Hierarchy

1. The 3 items marked in grey: `AgentStateDescription`, `AgentFunctionInput`, `AgentFunctionOutput` should all be removed, they are redundant as they only hold a single item (state name, message name) **(Done)**
2. The `SimulationLayer` hierarchy, should be part of the model description, it's not clear why it's instead part of Simulation. This also creates additional complexity in testing (have to define agents/agent fns before you can create simulation object). There should be a clear separation of 'design' and 'execution', this layer thing sits in the middle adding additional (unnecessary) complexity. **(Done)**
3. `ModelDescription` is the only class that the modeller should be able to create directly, all other description classes should be created and owned by their parent (protected constructor/destructor, public r-value assignment operator), returning the modeller a reference on request. This stops them passing a reference to an item which can later go out of scope. **(Done)**
4. Proper copy constructors; as users cannot create these classes themselves, it needs to be possible for them to 'import' a reference from a different `ModelDescription` hierarchy, (which clones as much as feasible?).
    a. These would also come in use when creating the Simulation, so we can hold a copy internally?
5. Once the `Simulation`/`CUDAAgentModel` has been created from the `ModelDescription`, any attempt to change the description (or it's children) should throw an exception unless the modeller explicitly acknowledges that it isn't linked to the created object. This would make it clear when changes will have no impact on the simulation. **(Done)**
6. Standardisation of naming, if only 1 exists/can be added, it should be treated as though it always exists (only has a get method), if multiple can be added there should be a create/new method, and a get method. **(Done)**
7. Move all typedefs inside the class definitions, they shouldn't be at global scope if they're only used by the class. **(Done)**
8. I 'm undecided as to whether using `std::reference_wrapper` to put references in a collection is a hack (based on ['references should not be reseated'](https://isocpp.org/wiki/faq/references#refs-vs-ptrs)), I suppose it depends on whether we can ensure the parent containing the message/agent function will not be deleted when the function is referenced elsewhere. **(Redundant)**

This will involve rewriting the class affected by #145, making that issue redundant.

### LINKED ISSUE #140: Replace agent function macros
The heavy use of preproccessor macros is ugly and harmful to code parsing.

I have a feeling it might be possible to replace it with templated code on the library side, similar to how thrust and cub require a struct containing `operator()` implementation to be passed to their device reductions.

* User defines a struct with a function of known prototype
* They pass an instance of that struct to the addAgentFunction method, which is templated with a typename for the arg which takes the function name
* The agent function `__global__` kernel wrapper is also templated to take this agent function method
* The pointer to the `__global__` is acquired (which should forcibly instantiate the templated version of it)
* This is then treated much the same onwards, however when launching the device functions you don't have to pass a device function pointer.

-----

Secondary benefit, it removes the calling of a device function via a device pointer. The compiler explicitly cannot optimise across such boundaries, so it might improve performance.

### LINKED ISSUE #145: Redundancy in AgentFunctionDescription.
This class has very few comments, so it's not too clear. 

3 concerns

## Point 1:
It uses the term 'input/output' to refer to message, which is quite unclear.

## Point 2:
It has a set of functions `setInput()`, `setOutput()`, `getInput()`, `getOutput()`, which allow a single `AgentFunctionInput`, `AgentFunctionOutput` to be stored in a `InputsMap` or `OutputsMap` respectively (see #132 for why references are not appropriate).

It also has `setInputChild()`, `setOutputChild()`, `getInputChild()`, `getOutputChild()`, which similarly store a single `AgentFunctionInput`, `AgentFunctionOutput` respectively, however this time as pointers.

Most of the other functions refer to the first (map) version.

It doesn't appear the InputChild variants are used by any examples

## Point 3:
The classes `AgentFunctionInput`, and `AgentFunctionOutput` seem redundant, they only contain a message name, why can't the user just pass the message name directly?

# TODO:
* Refactor so that `AgentFunctionInput`, and `AgentFunctionOutput` don't exist.
* Remove InputChild variants
* *Stop abusing references, behind the scenes use shared/weak ptrs and don't let modellers create most objects themselves (so that we retain ownership): (https://en.cppreference.com/w/cpp/memory/enable_shared_from_this)*

### LINKED ISSUE #168: Optional (Brute Force) Messages
Currently looking at this. Two points of query/concern.

1.
`d_scan_flag` is used to mark when a message has been written to (and is later exclusive summed, to manage the scatter of all written messages into a contiguous block). 

Should this array exist once globally? Or once per Message type?

Should this array be some global define? Or passed as a variable?

I'm on the side of a single global, similar to how device random is handled.

2.
Sorting/scattering messages requires a movement for each variable. Without a significant rework, this will require launching a kernel per variable. However, I expect this is something we will want to avoid.

### LINKED ISSUE #170: Bug: Curve hash collision causes variable to not be unregistered.
Curve uses linear probing, so when a hash collides it looks for the next free slot and stores the variable there. This has the effect of leaking the colliding variables each iteration (until Curve overflows).

This creates the issue that when unregister a collided variable, if the variable it collided with has already been unregistered, the linear probe fails.

The traditional solution to this is to have a deleted flag, in addition to the empty flag. So that linear probing will also continue over deleted flags (except for when inserting).

Robadob:
Fixes #132 
Fixes #140 
Closes #145 (now redundant)
Closes #124 (now redundant)
Fixes #170
Closes #168 
Closes #49 

Kind of affects #136 (When Simulation is defined, it takes a copy of ModelDesc data)
Affects #56 (still need spatial partitioning version)

**Changelog:**
* `ModelDescription` is the only description class which can be created directly by the modeller, child classes are now created by their parent instance. (#132)
* Changes to how agent functions are represented internally, this prevents passing of device function pointers to kernels. (#140)
* `Simulation` is now an abstract class, `CUDAAgentModel` extends this.
* Removes `AgentPopulation::setInitialPopulation()`, `AgentPopulation::setPopulation()` now also handles this behaviour.
* Changes input file to an optional argument
* Promotes `TinyXML` error codes to exceptions.
* **Fixes a bug in AgentInstance::setVariable()** (By casting to std::vector and using insert, it would sometimes trigger resize and break any subsequent calls to get or set).
* **Fixes a bug in Curve**, hash collision lead to a leak in the hash table on deletion. (#170)
* Fixes/Tests brute force messaging.
* Adds brute-force Circles example (based on my thesis version). (#56)
* Able to define spatial messaging type (though they will work as brute force currently), **pending discussion on how message access will be handled in agent fn.**
* `Simulation` now tracks how many model steps have occured.
* Optional [brute-force] messages now work (#168)
* Agent Death: Adds agent death, ability to resize CUDAAgent and unifies scatter operation also used by optional messages (and agent birth) (#49)


Robadob:
Christmas TODO:
* Test
  * CUDAAgent/AgentInterface?
  * CUDAMessage/MessageInterface? (not even sure latter exists)
  * AgentPopulation
* Document
  * CUDAAgentModel/Simulation
  * CUDAAgent/AgentInterface
  * AgentPopulation/AgentInstance

(and fix messages)

Issue #179, opened 2020-02-13
Feature: Runtime Compilation of Agent Functions


mondus:
```cpp
static const char* agent_func_str = "\
	FLAMEGPU_AGENT_FUNCTION(output_message) { \
		FLAMEGPU->addMessage<int>("id", FLAMEGPU->getVariable<float>("id"));\
		FLAMEGPU->addMessage<float>("x", FLAMEGPU->getVariable<float>("x"));\
		FLAMEGPU->addMessage<float>("y", FLAMEGPU->getVariable<float>("y"));\
		FLAMEGPU->addMessage<float>("z", FLAMEGPU->getVariable<float>("z"));\
		return ALIVE;\
	}"
							
LayerDescription &layer = model.newLayer();
layer.addRTAgentFunction(agent_func_str);
```

PR #180, opened 2020-02-13
Feature: Runtime Compilation

mondus:
This pull request is to add run time compilation which is a necessity to having language bindings (e.g. python) for FGPU.

See issue #179 

Consider use of https://github.com/NVIDIA/jitify

TODO
- [x] - Basic Runtime compilation function executing in place of agent function
- [x] - Move RTC to jitify.
- [ ] - Basic RTC function executing in place of agent function but with flamegpu_device_api.h include.
- [ ] - Basic FLAME_GPU_AGENT_FUNC function.
- [ ] - Restructure compilation as part of CUDAAgent Model invocation to avoid forcing context. Store function string in description upto this point.
- [ ] - Tidy up cmake and add option for runtime compilation which would impact use of dlfcn-win32. Shoudl be able to build FGPU2 within runtime compilation support.
- [ ] - Merge new messaging in


src/flamegpu/gpu/CUDAAgentModel.cu
```diff
@@ -81,104 +83,136 @@ bool CUDAAgentModel::step() {
         /*! for each func function - Loop through to do all mapping of agent and message variables */
         for (const std::shared_ptr<AgentFunctionData> &func_des : functions) {
             auto func_agent = func_des->parent.lock()->description;
-            if (!func_agent) {
```
ptheywood: @mondus Your visual studio settings seem to be murdering indentation (replacing spaces with tabs, won't pass linting)

Robadob: Yeah I think it defaults to tab.

mondus: Done there will be a few more linting errors to still fix

mondus:
This pull request is rejected as the approach to RTC has been improved

Issue #55, opened 2018-10-01
Feature: Instrumentation / Profiling tools


ptheywood:
Add instrumentation tools for simple performance analysis, along with NVTX based markers for more advanced profiling. Introduce a profile build configuration which enables this

+ [ ] Instrumentation
+ [ ] NVTX

PR #183, opened 2020-02-14
Initial NVTX marker support.

ptheywood:
Adds NVTX markers if enabled via CMake to improve profiling experience. 

Currently demonstrated in `CUDAAgentModel.cu`

No unit tests as not viable (no visible effects to the user)
Not documented very well with doxygen  due to use of macros. Will resolve in the future. 

See #55


Robadob:
> Will resolve in future

Will resolve before merge?

ptheywood:
Underwindows nvToolsExt.h cannot be found by default, so need to resolve that before merge.

ptheywood:
As of db0b766 nvtx markers now build and work on windows and linux, with a rought cmake mechanism set up which will be useful for @mondus to find and add nvrtc too. 

Unfortunately Cmake < 3.17 doesn't set all CUDA related parameters propperly for visual studio generators, so a slightly awkwared workaround was required. 

In the future if we move to require cmake >= 3.17 (unlikely any time soon) it should be possible to use the new features in 3.17 to replace most of the cmake changes made for this.

PR #209, opened 2020-03-05
Example: Boids

ptheywood:
Boids example implementation(s)

3 variants
+ [ ] Brute force
+ [ ] Spatially Partitioned messaging
+ [ ] Vector types

+ [ ] Initialisation 
    + It is unclear how the FLAMEGPU 1 boids model initialisation is designed. 
+ [ ] Validation
    + How can the boids example be validated


Robadob:
Vector types?

ptheywood:
There's a boids example in FLAMEGPU 1 which uses glm::vec3 to encode position and velocity as agent variables, rather than x/y/z and mapping. 

https://github.com/FLAMEGPU/FLAMEGPU/tree/master/examples/Boids_PartitioningVecTypes

Robadob:
Thinking about this, you might want to move the requirement for boids with vec types to the GLM issue (#217). Adding GLM will require tweaking quite alot of the codebase which currently only supports int types.

Whereas, porting an existing model to use vectors should be quite fast to do.

Issue #211, opened 2020-03-06
Fix Spatial messaging tests


Robadob:
Agent functions currenty only evaluate count (the rest is commented out for some reason).

Bonus: Make a test use > 1024 agents.

Issue #82, opened 2019-10-10
Messaging: Generic Sorted Messages


ptheywood:
Messages sorted by a specific key, either sorting all possible values or optimised to an upper limit of bins.

Could be specialised for special cases :
 
+ 1 message per key
+ along with multiple messages per key

Access would then be based on the user requesting messages from a specific key.

Robadob:
* `MsgArray`
    * 1 message per key (enforced with atomics?)
    * Keys form a contiguous integer block 
    * User sets min/max index* `ArrayMsg`
    * Could add a convenience function for interpolating messages variables? (for 2D/3D might be possible to back this with texture memory intrinsics.
* `MsgArray2D`
    * Same as above, 2D indexes, provides classic FGPU discrete functionality
    * Maybe a Moore neighbourhood iteration option.
* `MsgArray3D`
    * Same as above, 3D indexes
    * Maybe a Moore neighbourhood iteration option.

Little unsure about these ones:

* `MapMsg`:
   * Msg per key
   * User defined hash function
* `MultiMapMsg`: 
    * Multiple messages per key
    * User provides a hash function?
    * Question is over whether we would want it to be a real hash map, whereby we don't return messages with different key but same hash.
    * Basically has Spatial PBM at back end, could possibly rewrite spatial messages to inherit from

mondus:
Don't call it map call it bucket?

PR #213, opened 2020-03-06
Array/Array2D/Array3D Message Types

Robadob:
Addresses #82 (unsure if it counts as closing or not. vague issue)
**Changelog:**
* Adds `MsgArray` type with Moore iteration, and mandatory/optional output, exception unit tests
* Adds `MsgArray2D` type with Moore iteration, and mandatory/optional output, exception unit tests
* Adds `MsgArray3D` type with Moore iteration, and mandatory/optional output, exception unit tests
* Remove a redundant branch from `CUDAMessage::swap(bool, const unsigned int &, const unsigned int&)` (this relates to earlier bugfixes.
* Fixes spatial messaging tests (#211)
* Removes redundant 2nd template arg from `MsgBruteForce::Description::newVariable()` that would have allowed array vars to be added to messages (but not used).
* `MsgSpatial2D::Description` exception unit tests
* `MsgSpatial3D::Description` exception unit tests
* Example `Game of Life`
* Adds postfix `operator++` and `operator->` to all Msg iterators.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,6 +30,7 @@ option(BUILD_ALL_EXAMPLES "Enable building examples" ON)
 
 # Options to enable building individual examples, if BUILD_ALL_EXAMPLES is off.
 option(BUILD_EXAMPLE_HOST_FUNCTIONS "Enable building examples/host_functions" OFF)
+option(BUILD_EXAMPLE_GAME_OF_LIFE "Enable building examples/game_of_life" OFF)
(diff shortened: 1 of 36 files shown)


Robadob:
Undecided on interface for Moore neighbourhood iteration.
Challenge is that it requires radius and origin, so which order should the parameters be?

This seems like most reasonable:
```
inline __device__ Filter operator() (const size_type &x, const size_type &y, const size_type &radius = 1) const {
    return Filter(metadata, combined_hash, x, y, radius);
}
```

PR #214, opened 2020-03-06
Tidy examples

ptheywood:
The older examples (mas and simple_example) no longer really fit the current state of FLAME GPU / were partial implementations of now complete models, so they have been removed.


Robadob:
Merging this before @mondus has responded re: simple_example/mas. I'm going to start working on a new example and i'd rather avoid the conflict.

PR #215, opened 2020-03-06
Replicate standard output from FLAME GPU 1

### LINKED ISSUE #210: FLAME GPU 1 Standard output
FLAME GPU 1 prints various things to stdout, e.g.

```
Processing Simulation Step 1
Processing Simulation Step 2
Total Processing time: 2.320352 (ms)
```

Do we want to do this? It might be useful for progress checking. 

It might be better to hide (some of) this output behind a default flag, such as `-v/--verbose` for the `Processing Simulation Step N` messages, and `-t/--timing` for the timing information?

This may become more complex with submodels.

ptheywood:
+ [x]  Add -v/--verbose flag to Simulation. CUDAAgentModel then prints iteration number to stdout.
+ [x] Add -t/--timing flag to Simulation. CUDAAgentModel then times and outputs the simulation timing.

Closes #210

### DIFF
--- a/src/flamegpu/sim/Simulation.cu
+++ b/src/flamegpu/sim/Simulation.cu
@@ -103,6 +103,12 @@ int Simulation::checkArgs(int argc, const char** argv) {
             config.random_seed = static_cast<unsigned int>(strtoul(argv[++i], nullptr, 0));
             continue;
         }
+        // -v/--verbose, Verbose FLAME GPU output.
+        if (arg.compare("--verbose") == 0 || arg.compare("-v") == 0) {
+            // Reinitialise RandomManager state
+            config.verbose = true;
+            continue;
+        }
         // Test this arg with the derived class
         if (checkArgs_derived(argc, argv, i)) {
             continue;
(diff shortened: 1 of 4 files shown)


src/flamegpu/sim/Simulation.cu
```diff
@@ -118,9 +133,12 @@ void Simulation::printHelp(const char* executable) {
...
     const char *line_fmt = "%-18s %s\n";
-    printf(line_fmt, "-i, --in", "Initial state file (XML)");
-    printf(line_fmt, "-s, --steps", "Number of simulation iterations");
-    printf(line_fmt, "-r, --random", "RandomManager seed");
+    printf(line_fmt, "-i, --in <file.xml>", "Initial state file (XML)");
+    printf(line_fmt, "-s, --steps <steps>", "Number of simulation iterations");
+    printf(line_fmt, "-r, --random <seed>", "RandomManager seed");
+    printf(line_fmt, "-v, --verbose", "Verbose FLAME GPU output");
+    printf(line_fmt, "-t, --timing", "Output timing information to stdout");
+    printf(line_fmt, "    --no-timing", "Do not output timing information to stdout");
```
Robadob: Why have an option of on and a separate option for off? Surely one is redundant.

ptheywood: I was probably thinking about making it opt-out but got distracted (to match current flame).

PR #219, opened 2020-03-19
CMake changes for Visualiser

### LINKED ISSUE #69: Visualisation
FLAME GPU 1 provides a simple visualisation tool, and the ability to implement custom visualisations.

This may be a separate tool for FLAME GPU 2, for real-time visualisation and/or playback visualisation from a collection of output files (or similar)

**Work is ongoing in a separate repo: https://github.com/FLAMEGPU/FLAMEGPU2_visualiser**

This should be pulled in by CMake changes in `visualisation` branch, but a new option must be enabled.

Robadob:
Closes #69

These changes enable `VISUALISATION` flag to be set at CMake time.

This flag pulls in the [visualiser repo](https://github.com/FLAMEGPU/FLAMEGPU2_visualiser) and marks it as a dependency of FGPU2, configures include dirs and copies SDL.dll on compilation success.

Have tested it with the root CMake project, and added it/tested with example ones too.

*This should probably remain draft until the visualiser is ready, and an example has an optional vis.*

**Two possible remaining tasks:**
* Visualiser depends on GLM, FGPU2 will depend on GLM, this could mean the repo is cloned twice independently (and due to size its a fairly slow clone).
* Due to the nature of CMake options, if `VISUALISATION ` is set `ON` all examples/tests will be setup to have visualiser enabled, even if they don't require it. Probably not a big deal but notable.


Robadob:
Now have it working with a simple fixed example, I haven't commit these changes due to the large amount of dependency hell I need to cleanup (to save polluting the vis repo).

Brief video demo here: https://youtu.be/-W7fkSgg3IM

**Jobs Todo:**
* Removal of non core C/C++ from interface (this will reduce the need to share includes, it's kind of recommended behaviour for libraries)
* Bring the visualiser code upto FGPU2 standards (lint, vs2015+ compilation warnings)
* Improve CMake (remove include sharing, handle freetype, handle glew)

ptheywood:
I'm happy to test / fix this on linux once it's worth doing. The SDL installation wiki suggests using the distribution provided copy rather than building your own copy. This should just be looking for SDL2, and if not found but requested building it at that point in time. 

As an aside we might want to think about offline builds. I.e. what if you have a local copy of gtest but due to lack of internet the download fails? This is probably worthy of an isse?

Robadob:
> I'm happy to test / fix this on linux once it's worth doing.

I'd wait until the todo list is completed, some of that cleanup should simplify any issues you'll need to address.

>This should just be looking for SDL2, and if not found but requested building it at that point in time.

SDL already works like that, I haven't looked into freetype or glew yet.

> As an aside we might want to think about offline builds. ...

That's not too impactful, I've already done that for the FGPU2_vis include, so I can dev it not in a weird nested folder.

Robadob:
It should now be possible to build the visualisation on windows via the main FGPU2 cmake. Running might be a challenge due to the resource paths for shader/model, which are currently hardcoded to paths on my machine.

Robadob:
Should start build a list of configuration options to add support for: (feel free to edit this comment with suggestions)

**Model/Window Level:**
- [ ] Window title
- [ ] Window dimensions
- [ ] Environment scale (something to allow us to scale the view frustum, we might be able to auto do this based on spatial partitioning dims)

**Agent/AgentState Level:**
- [ ] Model file
- [ ] Model scale
- [ ] TBD Shader customisation stuff

** Misc:**
- [ ] Add static models (set position, scale, rotation, shader options)

Robadob:
Started working on agent customisation, model path and scale (1d or 3d) can be chosen. Need to also allow this to be customised on per agent-state basis.

Scale isn't really worth using until env scale options are in place (shader is still hardcoded to multiply agent locs by some constant atm).

Robadob:
From today's meeting Paul wants a minimal functional version, so that I can move onto submodels.

I think the key things remaining for this are:

* Environment scale
* Replace use of assert/fprintf
* Display step numberf on HUD

Should all be quite trivial to complete

Robadob:
I'm happy for this to be merged now.

PR #221, opened 2020-03-20
Reserved names

### LINKED ISSUE #206: Reserved variable names
To prevent users from specifying variable names which would result in curve conflicts for flamegpu compulsory features, it might be best to prevent variables being declared with certain names, or names which follow a pattern.

~E.g. following unix file name conventions, if the variable string name begins with `.` it could be reserved. This would then only require that at variable definition time the string is checked to not begin with `.`.~

Variables that begin with `_` shall be reserved. A single character check should be cheap and work for variable names of any length.

However, setting the variable may still be achievable if it exists by default without adding (potentially) expensive additional runtime checks on the host and the device to prevent users from setting these values (while allowing internal properties to set them?) 

-----------

~~**As of #218 it has also become necessary to prevent users creating agent's with variables `name` and `state` (due to the xml output format).**~~ *(This was completed within the PR)*

Robadob:
Closes #206
*This is forked from xml_export, so should be merged after (hence keeping it draft).*

Update Agent, Message, Env vars to not allow variables that begin with `_` to be created or set by the user (haven't bothered disabling read-only methods, majority of cases variable doesn't exist will trigger).

Affected Methods:
* `AgentDescription::newVariable()`
* `FLAMEGPU_DEVICE_API::setVariable()`
* `FLAMEGPU_DEVICE_API::AgentOut::setVariable()`
* `FLAMEGPU_HOST_NEW_AGENT_API::setVariable()`
* `MsgBruteForce::Description::newVariable()` (Inherited by all? message types)
* `MsgBruteForce::Out::setVariable()` (Inherited by most message types)
* `MsgArray::Out::setVariable()`
* `MsgArray2D::Out::setVariable()`
* `MsgArray3D::Out::setVariable()`
* `EnvironmentDescription::add()`
* `EnvironmentDescription::set()`
* `EnvironmentDescription::setConst()`
* `HostEnvironment::add()`
* `HostEnvironment::set()`

### DIFF
--- a/include/flamegpu/model/AgentDescription.h
+++ b/include/flamegpu/model/AgentDescription.h
@@ -222,6 +222,10 @@ class AgentDescription {
  */
 template <typename T, ModelData::size_type N>
 void AgentDescription::newVariable(const std::string &variable_name, const std::array<T, N> &default_value) {
+    if (!variable_name.empty() && variable_name[0] == '_') {
+        THROW ReservedName("Agent variable names cannot begin with '_', this is reserved for internal usage, "
+            "in AgentDescription::newVariable().");
+    }
     std::string lower_variable_name = variable_name;
     for (auto& c : lower_variable_name)
         c = static_cast<char>(tolower(c));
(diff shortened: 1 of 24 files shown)


Robadob:
TODO: prevent agents from having variable 'name' or 'state', this conflicts with xml io.

PR #225, opened 2020-04-15
Rtc feature

mondus:
This pull request is to add run time compilation which is a necessity to having language bindings (e.g. python) for FGPU.

See issue #179

- [X] - Support and tests for agent variables
- [x] - Support and tests for environment variables (requires refactoring)
- [x] - Support and tests for messaging
- [x] - Support and testing for random (and any other memory allocated inside a `flamegpu_internal` namespace)
- [x] - Support and tests for agent function conditions
- [x] - Support and testing for agent outputs

Support for environment variables requires either

1) Merging all RTC agent functions into a single compilation unit. In which case agent functions could be compiled individually when added to the model (so that errors can be correctly reported) and then compiled as a single unit with a single definition of the required environment `__constant__` memory.
2) Pre-build a library file with the environment `__constant__ `defined and link this with each compilation unit. This is probably not an option as each cuda module will probably end up with its own version of the `__constant__`.

*Note: That agent variable support has been added by creating a run-time version of curve which maps a variable name to a memory address at compile time. This requires a dynamically created template function. For this to be extended to support additional namespaces (e.g. environments etc.) The namespace integer must be used. The only way to do this is to switch based on the namespace at run-time as the namespace is not a compile time variable (e.g. not a template parameter).*


mondus:
Curve RTC has now got a namespace switch which can be used for registering different variables (e.g. env variables). This adds to code size but and a very small branch instruction overhead at runtime (which is still faster than runtime curve).

Robadob:
> Curve RTC has now got a namespace switch which can be used for registering different variables (e.g. env variables). This adds to code size but and a very small branch instruction overhead at runtime (which is still faster than runtime curve).

Is it possible to put that behind the `__cudacc_rtc__` macro?

mondus:
> > Curve RTC has now got a namespace switch which can be used for registering different variables (e.g. env variables). This adds to code size but and a very small branch instruction overhead at runtime (which is still faster than runtime curve).
> 
> Is it possible to put that behind the `__cudacc_rtc__` macro?
> 
> _Though on a second read i might be misunderstanding, I guess you meant faster than compile time curve? and this is a purely RTC thing._

Yes the RTC version will be faster than the non RTC version as `getVariable ` etc. does not need to look anything up from constant memory it just returns the device pointer directly.

Robadob:
First if you haven't already you may want to change your git editor.

e.g. I use `git config --global core.editor "'C:/Program Files (x86)/Notepad++/notepad++.exe' -multiInst -notabbar -nosession -noPlugin"`

This saves messing about with vim.

First checkout to a spare branch

`git checkout -b rtc_feature_rebase`

Then you want to squash your branch.

`git rebase --interactive HEAD~49`

This should open up your text editor with a list of all the commits, change any commits to be squashed into previous commit to say squash. After closing the editor, it will allow you to change the message for each new commit.

There shouldn't be any conflicts during this process. The less eventual commits the better, as the next rebase will work by reapplying you commits over master.

Next check master is current
`git checkout master`
`git pull`
`git checkout rtc_feature_rebase`

Then do the big rebase

`git rebase master`

It will stall whenever it hits a conflict, do `git status` check the conflicting file, fix the conflicting file.

`git add <file>`
`git rebase --continue`
When it stops telling you stuff it's done.

ptheywood:
This is how I locally squashed all of the commits in this branch (without using rebase so I didn't need to change pick to squash 40 times). The last common parent between this branch and master is `c763bc2`.
Note that i'm super keen on using `git reset --hard` which makes this kinda dangerous.

This is not the only way to do this and loses your commit history. 

### Make sure local branches are up to date

```bash
git fetch origin
git checkout master
git reset --hard origin/master
git checkout rtc_feature
git reset --hard origin/rtc_feature
```

### Backup your branch locally now it is up to date

```
git checkout rtc_feature
git checkout -b rtc_feature_backup
git checkout rtc_feature
````


### Roll back in time to the last common ancestor of rtc_feature and master

```bash
git reset --hard c763bc2
```

### Squash all of the RTC commits together 

```bash
git merge --squash rtc_feature
Resolve conflicts here
git commit -a -m "RTC Implementation"
```

### Rebase onto master

```bash
git rebase master
```

PR #240, opened 2020-04-24
 Automatic Agent ID

### LINKED ISSUE #199: Agent ID generation (host and device)
FLAME GPU 1 has methods to generate (probably) unique IDs for agents from the host or the device. 

General process is:

1. After any agents have been loaded from disk (i.e. already have Ids) the maximum id value is found and stored on the host and the device.
2. host method to generate new IDs in init/step functions, which increments the host-value and sets a flag indicating that the device copy is out of date
3. If the flag was set, copy the new value to the device before any kernels which generate new agents (or kernels in general?).
4. Device function which generates new id (atomically). If this value has been updated it needs copying back to the host

It might be easeir to just use managed memory.

ptheywood:
Automatically provide an `id` for all agents.

+ [x] Typedef for the type to simplify changing this (`AgentData::IDtype`)
+ [ ] define / static constexpr for a default / uniinitalised value (`AgentData::IDMAX`/`AgentData::IDUnassigned` equal to `UINT_MAX` or `-1`)  
+ [x] Automatically add agents a hidden variable `_id` - prevents the user from changing this.
+ [ ] Implement  
+ [ ] `const` accessor methods where appropriate in the user facing API
    + [ ] Device code via `FLAMEGPU->id()` or `FLAMEGPU->agent_id()` or similar
    + [ ] Host access via some iterator - requires host access of agent data.
        + For agent birth in step functions + agent iteration in step functions.
    + [ ] Host access during population creation? - via `AgentntInstance::id()`?
+ [ ] Assign unique value per agent population. 
    + [ ] `AgentDescription`/`AgentData` stores a value for the maximum assigned so far / next to assign, initialised to 0 / -1
    + [ ] New populations are allocated values based on the max so far.
    + [ ] When xml io is used, the value should be reset (as currently this resets all populations)
    + [ ] New agents from host functions within simulation should use the current value
    + [ ] New agents from device functions should use and update the new value. This can be assigned after the agent function has executed, 

Also need to:
+ [ ] Remove explicit `id` from examples, to make use of the new feature.

### Tests

+ [ ] Access on host
+ [ ] Access on device
+ [ ] Check the property exists
+ [ ] Check the user cannot request the property (this may be covered by other tests already)
+ [ ] Test auto generation values
    + [ ] Problematically generated population (no existing pop)
    + [ ] Generation from multiple pops, how will this work.
    + [ ] Loading from disk
    + [ ] Combination of population + disk access - order is important
    + [ ] Generation on device for new agents
    + [ ] Generation on host for new agents (existing population)
+ [ ] ...

### Documentation:

+ [ ] Document that `_id` / `agent.id()` exists
+ [ ] Describe it's behaviour + example usage.
+ [ ] Describe the limitations.

### Questions:

+ [ ] Reading/Writing from disk
    + Should this be written and read from disk? This increases complexity.
    + Should `_` props be read/written in general? (no?)
+ [ ] Multiple host populations
    + When creating host populations, how should id be handled. Should there be a population relative ID which are then mutated when the pop is added to a cuda agent model? This will need thinking about.

+ [ ] do we want this to be optionally disabled to save memory? (probably not)
    + Via preprocessor macro? 
        + otherwise this will leave pointless methods available to the user - i.e. the id() method would have to just return some arbitrary value. 
        + but this would break the static library for all models that do use the feature.

+ [ ] Per agent-type unique Ids, or globally unique Ids?
    + Global uniqueness might make life a little easier, although lead to potential confusion.

+ [ ] index vs ID.
    + This distinction probably needs to be clearly distinguished in a few places.

This will close #199


ptheywood:
After rebasing this doesn't pass all tests which previosuly passed, but this is somewhat on hold for now anyway.

PR #248, opened 2020-05-04
Disable MSVC warning 4100 - unreferenced formal parameter.

ptheywood:
This warning is hard to avoid with agent/host functions which do not refernece the FLAMEGPU object.

Untested on windows


ptheywood:
ready to merge after rtc

PR #251, opened 2020-05-06
Improve finding of nvrtc/nvtx via cmake

### LINKED ISSUE #249: findNVRTC edge case.
On Systems with multiple CUDA installs the find_nvrtc cmake module I wrote does not always find matching nvrtc to cuda version found. 

I.e. on a ubuntu 16.04 system with the ubuntu distributed version of cuda installed + more recent versions manually installed (via apt, or runfiles) it is possible for a mismatch to occur.

The issue is that the system install nvrtc can be higher on the path than the user installed with the current search order in the module. 

+ [ ] Fix the edge case search path
    + Finding nvrtc.h in `/usr/include` over `/usr/local/cuda-*/...`
    + `NO_DEFAULT_PATH` ?
    + Need to ensure the module is using the cuda path  as higher priority than `CMAKE_INCLUDE_PATH` 
+ [ ] Add instructions to use `-DNVRTC_LIBRARIES` and `-DNVRTC_INCLUDE_DIR` as overides somewhere.

ptheywood:
Improve finding of nvrtc/nvtx via cmake

Also formats those files to use spaces not tabs.

Resolves #249. 

This should not be merged without a manual cherry pick of the commit post - rtc_feature merge.


Robadob:
Tested it on mav?

ptheywood:
> Tested it on mav?

I did this on mav as it's the only place with the bug. 

Untested on windows, but as I've only added HINTS directives and kept the existing PATHS directives it should have no effect

Robadob:
> I did not test the RTC_test compilation however.

Don't believe that would be required.

Robadob:
This commit is now in rtc_feature branch via cherry-pick.

PR #258, opened 2020-05-11
Doxygen Tweaks

ptheywood:
This PR is for making tweaks to the doxygen configuraiton.

So far: 
+ [x] Strips absolute paths to be `FLAMEGPU_ROOT` relative i.e. in diagrams `/home/user/path/to/flamegpu/include/flamegpu/model/AgentFunctionDescription.h` becomes `include/flamegpu/model/AgentFunctionDescription.h`

I'd also suggest:

+ [x] restricting the number of layers to which diagrams are generated to `1` via `set(DOXYGEN_MAX_DOT_GRAPH_DEPTH   1)`. 

This greatly simplifies the include dependency diagrams so they are actually legible, but does lose some information on dependency chains. 
Links are generated so that you can easily follow dependency chains through (red boxes).

I.e. for `AgentFunctionDescription.h`:

With `DOXYGEN_MAX_DOT_GRAPH_DEPTH = 0`

image

And with `DOXYGEN_MAX_DOT_GRAPH_DEPTH = 1`

image

And with `DOXYGEN_MAX_DOT_GRAPH_DEPTH = 2`

image

### DIFF
--- a/cmake/doxygen.cmake
+++ b/cmake/doxygen.cmake
@@ -46,8 +46,9 @@ function(create_doxygen_target FLAMEGPU_ROOT DOXY_OUT_DIR XML_PATH)
         # Limit diagram graph node count / depth for simply diagrams.
         set(DOXYGEN_DOT_GRAPH_MAX_NODES   100)
         set(DOXYGEN_MAX_DOT_GRAPH_DEPTH   0)
-        # Select diagram output format {png, YES/NO}, {svg, YES}, {svg, NO} 
+        # Select diagram output format i.e png or svg
         set(DOXYGEN_DOT_IMAGE_FORMAT      png)
+        # If using svg the interactivity can be enabled if desired.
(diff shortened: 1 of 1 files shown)


ptheywood:
> Less sure about the depth limit, a depth limit of 1 seems little better than just checking the includes on the files directly.

That's partially why i included the `2` option as an example. Any choice is a compromise between the diagram being readable and  how much informationit conveys. Another option would be to just disable the dependency graphs, as full graphs are illegible and small graphs are (somewhat) pointless.

 
> Not too sure what this comment is supposed to mean, could be clearer
> `# Select diagram output format {png, YES/NO}, {svg, YES}, {svg, NO} `

a quick way of suggesting pairs of parameters for the related options. I'll be more verbose.

ptheywood:
Set to MaxDepth 1 and then OK to merge.

PR #259, opened 2020-05-11
Remove redundant args sent to runtime compilation.

Robadob:
`__CUDACC_RTC__` is defined by the compiler, need not be specified.

`-rdc=true` is not used (RTC does not support linking with executing assembly), and carries a runtime overhead to performance (e.g. ~10% slower).

Tested this in Visual Studio 2015

Longer term, we might want to think about if we can remove `rdc=true` from the main library, we'd simply need to find a convenient way of managing our use of constant memory.

--------------
Unrelated note, most of the RTC tests have this compilation warning (with or without the change), unsure if its visual studio 2015 specific.

```
--- JIT compile log for rtc_env_func_program ---
dynamic/curve_rtc_dynamic.h(45): warning: extern declaration of the entity Curve::UNKNOWN_VARIABLE is treated as a static definition

flamegpu/exception/FGPUStaticAssert.h(14): warning: extern declaration of the entity FGPU_SA::integral_constant<bool, (bool)0> ::value is treated as a static definition

flamegpu/exception/FGPUStaticAssert.h(14): warning: extern declaration of the entity FGPU_SA::integral_constant<bool, (bool)1> ::value is treated as a static definition

flamegpu/gpu/CUDAScanCompaction.h(77): warning: extern declaration of the entity flamegpu_internal::CUDAScanCompaction::hd_configs is treated as a static definition

flamegpu/runtime/utility/DeviceEnvironment.cuh(16): warning: extern declaration of the entity flamegpu_internal::c_deviceEnvErrorPattern is treated as a static definition
```

### DIFF
--- a/src/flamegpu/gpu/CUDAAgent.cu
+++ b/src/flamegpu/gpu/CUDAAgent.cu
@@ -561,25 +561,11 @@ void CUDAAgent::addInstantitateRTCFunction(const AgentFunctionData& func, bool f
     std::string include_fgpu;
     include_fgpu = "-I" + std::string(env_inc_fgp2);
     options.push_back(include_fgpu);
-    // std::cout << "fgpu include option is " << include_fgpu << '\n';     // TODO: Remove DEBUG
 
     // cuda path
     std::string include_cuda;
     include_cuda = "--include-path=" + std::string(env_cuda_path) + "/include";
     options.push_back(include_cuda);
-    // std::cout << "cuda include option is " << include_cuda << '\n';     // TODO: Remove DEBUG
-
-    // add __CUDACC_RTC__ symbol
-    std::string rtc_symbol;
-    rtc_symbol = "-D__CUDACC_RTC__";
-    options.push_back(rtc_symbol);
-    // std::cout << "RTC pre-processor symbol is " << rtc_symbol << '\n';     // TODO: Remove DEBUG
-
-    // rdc
-    std::string rdc;
-    rdc = "-rdc=true";
-    options.push_back(rdc);
-    // std::cout << "rdc option is " << rdc << '\n';                       // TODO: Remove DEBUG
 
     // curve rtc header
     CurveRTCHost curve_header;


ptheywood:
> Longer term, we might want to think about if we can remove rdc=true from the main library, we'd simply need to find a convenient way of managing our use of constant memory.

I'm not sure how this would be achievable with multiple .cu files, without just making all cuda code header only (even then i'm not sure it would work if those functions are used in separate compilation units)? 
Aren't we taking the rdc performance hit to allow the much greater flexibility comapred to FLAME GPU 1? 
Admitidly the RTC mode needs all the performance improvements it can get given compilation time is factored in. 

> Unrelated note, most of the RTC tests have this compilation warning (with or without the change), unsure if its visual studio 2015 specific.

I get similar warnings running the test suite under linux (CUDA 10.2, GCC 7.5)

```
--- JIT compile log for rtc_env_func_program ---
dynamic/curve_rtc_dynamic.h(166): warning: statement is unreachable

dynamic/curve_rtc_dynamic.h(200): warning: statement is unreachable

dynamic/curve_rtc_dynamic.h(183): warning: statement is unreachable

dynamic/curve_rtc_dynamic.h(95): warning: statement is unreachable
          detected during instantiation of "T FLAMEGPU_READ_ONLY_DEVICE_API::getVariable<T,N>(const char (&)[N]) [with T=int, N=7U]" 
rtc_env_func_program(11): here

dynamic/curve_rtc_dynamic.h(45): warning: extern declaration of the entity Curve::UNKNOWN_VARIABLE is treated as a static definition

flamegpu/exception/FGPUStaticAssert.h(14): warning: extern declaration of the entity FGPU_SA::integral_constant<bool, (bool)0> ::value is treated as a static definition

flamegpu/exception/FGPUStaticAssert.h(14): warning: extern declaration of the entity FGPU_SA::integral_constant<bool, (bool)1> ::value is treated as a static definition

flamegpu/gpu/CUDAScanCompaction.h(77): warning: extern declaration of the entity flamegpu_internal::CUDAScanCompaction::hd_configs is treated as a static definition

flamegpu/runtime/utility/DeviceEnvironment.cuh(16): warning: extern declaration of the entity flamegpu_internal::c_deviceEnvErrorPattern is treated as a static definition
```

Robadob:
> I'm not sure how this would be achievable with multiple .cu files, without just making all cuda code header only (even then i'm not sure it would work if those functions are used in separate compilation units)?

You don't need RDC=true to share pointers, only to share `__constant__`, `__device__`.

All device functions are header side, so don't require rdc.

All use of `__device__` memory is to replaced by pointers when we move singletons into `CUDAAgentModel`.

This only leaves use of `__constant__` memory, a possible solution to this would be to `getAddressOfSymbol()` within the compilation unit, and use that in place of the symbol throughout management of the simulation.

This is just a longer term hypothesis.

PR #263, opened 2020-05-13
Update issue templates

Robadob:


Robadob:
This branch exposes that issue templates are in the wrong directory need to be moved into an `ISSUE_TEMPLATES` sub directory.

PR #272, opened 2020-05-15
Implement getStepCounter in Host and Device APIs

### LINKED ISSUE #266: Expose step_count in agent/step/host functions
The current step count should be exposed to the user via the various FLAMEGPU api objects. 

I.e. 
+ in agent functions, the user should be able to access the number of steps that have occured (device)
+ in step/host/exit functions, the user should be able to access the number of steps that have occured (host)
+ init functions (shouldn't) need this - they *should* only ever run when `step_count == 0`

`step_count` should be 0 indexed, and refer to the start of the current iteration.

Within submodels, `step_count()` will be relative to the submodel. If it needs to know about the parent models step, this could be calculated in the parent model and stored in an environment variable, mapped into the submodel by the user.

If `--steps 3`:

| function_type   | Iteration 0 | Iteration 1 | Iteration 2 |
|-----------------|-------------|-------------|-------------|
| Init function   |           0 |         N/A |         N/A |
| agent condition |           0 |           1 |           2 |
| agent function  |           0 |           1 |           2 |
| host function   |           0 |           1 |           2 |
| step function   |           0 |           1 |           2 |
| exit condition  |           0|           1 |           2 |
| exit function   |         N/A |         N/A |           3 |

### Todo

+ [x] Increment `step_count` as late as possible in `CUDAAgentModel->getStepCounter()`.
+ [x] Expose `step_count` on the host `flamegpu_host_api`
+ [x] Expose `step_count` on the device via `flamegpu_deivce_api.h`
+ [x] Call methods `getStepCounter` to be consistent with `Simulation::getStepCounter()`
+ [x] Tests
    + [x] Test device getStepCounter() in agent function
    + [x] Test device getStepCounter() in agent func condition
    + [x] Test device getStepCounter() in RTC agent function
    + [x] Test host getStepCounter() in init func (==0)
    + [x] Test host getStepCounter() in host layer func (not incremented yet)
    + [x] Test host getStepCounter() in step func (not incremented yet
    + [x] Test host getStepCounter() in exit func
    + [x] Test host getStepCounter() in exit condition

ptheywood:
Closes #266 

+ Accessible in all HOST api functions (Init, step, host, exit, exit condition)
+ Accessible in device api functions (agent functions, agent function conditions, RTC)
+ Tested

Changes had to be made to how RTC environment stuff works.

History could be squashed, though it does include a few useful standalone commits.

### DIFF
--- a/include/flamegpu/gpu/CUDAAgentModel.h
+++ b/include/flamegpu/gpu/CUDAAgentModel.h
@@ -166,6 +166,10 @@ class CUDAAgentModel : public Simulation {
      */
     unsigned int step_count;
     /**
+     * Update the step counter for host and device.
+     */
+    void incrementStepCounter();
+    /**
      * Map of agent storage 
      */
     CUDAAgentMap agent_map;
--- a/include/flamegpu/runtime/flamegpu_device_api.h
+++ b/include/flamegpu/runtime/flamegpu_device_api.h
@@ -65,6 +65,14 @@ class FLAMEGPU_READ_ONLY_DEVICE_API {
      */
     const DeviceEnvironment environment;
 
+    /**
+     * Access the current stepCount
+     * @return the current step count, 0 indexed unsigned.
+     */
+    __forceinline__ __device__ unsigned int getStepCounter() const {
+        return environment.get<unsigned int>("_stepCount");
+    }
+
  protected:
     Curve::NamespaceHash agent_func_name_hash;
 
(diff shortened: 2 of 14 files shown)


ptheywood:
The ComputeCapability errors being reported in BruteForce don't make any sense to me as the `getComputeCapability` test doesn't appear to touch CUDAAgentModels at all. Probably worthy of it's own issue so we don't lose it in the PR.

Robadob:
I've created a PR which fixes it. It's because they were being thrown
without my macro which reports line number. So they were reporting line
number of previously thrown error.

PR #277, opened 2020-05-19
Better fix for Jitify on Mavericks as suggested by Jitify Dev

### LINKED ISSUE #262: Improved fix for mavericks/rtc
See https://github.com/NVIDIA/jitify/issues/65#issuecomment-627277428

Current workaround of passing `--include-path=` will be patched in future, therefore it's suggested we instead use `--pre-include=/usr/local/cuda/include/cuda.h`.

Need to make this change and test it when I'm next playing with mavericks

Robadob:
closes #262
Tested on Mav


Robadob:
This might conflict with #259

PR #282, opened 2020-05-19
Remove HostEnvironment::add/remove()

### LINKED ISSUE #244: HostEnvironment::add should be removed
This should be removed. It should not be possible to create new environment variables in step functions as this is impossible to map to RTC which requires a fixed set of variables in the model description at compile time (e.g. first call to step).

There are also no modelling use cases for this outside of the text_host_environment.cu tests.

Robadob:
Closes #244


Robadob:
Affected tests build and pass.

Noticed RTC is broken on master again though. Need to look into this.

PR #283, opened 2020-05-20
Enable swig on travis

ptheywood:
This should result in a CI failure currently

Wait for travis to run before mergin (it should fail, but only when building and not during configuration.


mondus:
Requires a fix for function pointer: E.g. https://travis-ci.org/github/FLAMEGPU/FLAMEGPU2_dev/builds/689306462#L911

PR #291, opened 2020-05-26
Swig to enable Python models

mondus:
Outstanding

- [x] Basic library building
- [x] Python package building post event
- [x] Python virtual env building post event
- [x] Host agent includes (currently has an issue with Thrust which may be fixable with refactoring header)
- [x] Template instantiation
- [x] Host function pointers in layers
- [x] PyTest tests
- [x] CI build test integration
- [x] doc strings

Outstanding Desirable

- [ ] Generate Python docs
- [ ] CMakeTest integration (See #114 )


mondus:
Nested templates in swig are not straight forwards. This is required for messaging. The following gist provides a solution for making this work: 

https://gist.github.com/mondus/f142ca4c2ebc94eac7d070b521a367cc

There will however need to be some refactoring of the message headers to split device and host code.

mondus:
Major refactoring of messages has been completed. This separates host and device functionality of message implementations into separate headers. From swigs perspective it means that the host only header can be included to do code generation from. It also removes the requirement of having RTC guards as a separate device header hides any host code from jitify.

PR #304, opened 2020-06-08
Revert "Fix header order issues newly detected by cpplint 1.5.0"

ptheywood:
This reverts commit 02d26d1f38cb269ed282f3fc72366d1ea0faba48.

cpplint 1.5.1 once gain breaks linting due to order of includes. Reverting to the previous configuration seems to be the simple solution, as `<cuda.h>` is once again correctly detected as a c system header, and expected to be before c++ system headers such as `<string>`.


ptheywood:
CI failure is just appveyor crashing on reboot which I could not re-issue.

PR #311, opened 2020-06-16
This is probably how step functions should work with Swig.

Robadob:
I've added support for init/step/exit/host-layer functions, this works exactly the same as C, so multiple can be added.

This builds, and the tests in `test_cuda_agent_model.py` pass. (Not sure why appveyor is giving a linker error, I addressed that and can build with/without SWIG locally)

I had to add `SWIG` preprocessor macro to hide the functions which are implemented by swig. I couldn't use `%extend` as that produces an external method which can't access private members. (But this is better than adding public accessors which a user has no reason to call).

Obviously, same thing needs to be repeated for `ExitCondition` too.


Robadob:
@mondus Can this be closed if you've now taken what you wanted from it?

P.S. I noticed you didn't add host-layer functions in your commit to SWIG branch.

PR #314, opened 2020-06-18
set CMAKE_CUDA_ARCHITECTURES to OFF for CMake 3.18+

### LINKED ISSUE #313: CMake: Address CMP0104
Generating project with CMake 3.18.0+ causes this warning for all cuda targets:

```
CMake Warning (dev) in src/CMakeLists.txt:
  Policy CMP0104 is not set: CMAKE_CUDA_ARCHITECTURES now detected for NVCC,
  empty CUDA_ARCHITECTURES not allowed.  Run "cmake --help-policy CMP0104"
  for policy details.  Use the cmake_policy command to set the policy and
  suppress this warning.
  CUDA_ARCHITECTURES is empty for target "flamegpu2".
This warning is for project developers.  Use -Wno-dev to suppress it.
```

Presumably, renaming `CUDA_ARCH` to `CUDA_ARCHITECTURES` will fix it?

(This warning can be ignored, however the intention of policy warnings is that you fix behaviour as old behaviour is deprecated).

It can be suppressed with `set_policy(CMP0104 OLD)` at the top of `common.cmake`.

ptheywood:
Disabled developer warnings which are genertated for each target otherwise.

Don't want to use the new property, as would restrict us to Cmake 3.18+, and it's not clearly documented what --gencodes it generates.

Also would not support CUDA 11 LTO (which may or may not help performance)

closes #313

### DIFF
--- a/cmake/cuda_arch.cmake
+++ b/cmake/cuda_arch.cmake
@@ -3,6 +3,11 @@
 
 # @todo - split setting of values out of the project so it's onyl done once before include_directory?
 
+# CMAKE > 3.18 introduces CUDA_ARCHITECTURES as a cmake-native way of generating gencodes (Policy CMP0104). Set the value to OFF to prevent errors for it being not provided.
+if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.18")
+    set(CMAKE_CUDA_ARCHITECTURES "OFF")
+endif()
+
 # Check if any have been provided by the users
 string(LENGTH "${CUDA_ARCH}" CUDA_ARCH_LENGTH)
 
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -5,6 +5,13 @@ cmake_minimum_required(VERSION VERSION 3.12 FATAL_ERROR)
 # Tests require GTest
 #enable_testing()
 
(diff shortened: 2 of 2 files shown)


Robadob:
This doesn't work.

The warnings occur at generate time, rather than configure time. So alot of bits of code are not touched.

ptheywood:
I didn't think they enabled CUDA as a target language though? Didn't seem to get those warnigns for me, though it may have already built gtest so didnt' try again? 

Didn't try the visualisation either. 

Should be able to change the scope of the variable (which might work depending on order), though might have to add it to the visualisation CMake separately.

Robadob:
> I didn't think they enabled CUDA as a target language though? 
They don't, I'm assuming it's a consequence (cmake bug?) of them being children of a CUDA project.

ptheywood:
> They don't, I'm assuming it's a consequence (cmake bug?) of them being children of a CUDA project.

Ahh yeah we enable the language before they are included I suppose (i also might not have had tests enabled when i tried this locally)

ptheywood:
@Robadob - I can't reproduce the warnings for gtest etc under linux using `3.18.0-rc2`, so seems limited to the visual studio generators (I'm guessing the vcxproj uses the CUDA props once CUDA is enabled, where as makefile generator is using the correct compiler). 

Will have to debug/fix this from Windows.

PR #316, opened 2020-06-19
Download external dependencies via CMake at configure time

ptheywood:
Uses Cmake FetchContent to download external dependencies, rather than them being tracked within the repo.

This removes a lot of code (Thrust/CUB) which  we don't need to track, just the version we want to use.

Jitify and Tinyxml2 aren't packaged in a super friendly way for this, so some awkwardness.

Thrust/CUB packaging from `1.9.10` is very cmake friendly, but they have a workaround in place that causes issues, so I found a workaround for their workaround to avoid initialisation order warnings being generated. 

There is room for further improvements / polish, which i'll add to an issue for future cleaning (better logging for end users, documentation for how to provide your own versions, cmake option to disable downloading for meterred connections). 

This upgrades the version of Thrust/CUB being used to get a bugdix for ExclusiveSum which was triggering a swig related test failure on some machines.

Also downloads googletest with fetch content (much simpler)

~This may need some adjustments for Windows (shouldn't, but might).~ 

**Edit** - this does need some adjusting for windows :cry:

Closes #284

PR #303, opened 2020-06-08
CUDA 11.0RC 

ptheywood:
Closes #302

+ Fixes the test suite for CUDA 11.0.167 under linux
+ Upgrades jitify to the latest version (commit 2a015bb) which works with CUDA 11.0 RC
+ Adds CUDA 11.0RC to the Ubuntu and Windows github action worflows as build matrix options.

Tested generated executables locally under linux. 
Windows executables untested (will likely wait until a non RC to install CUDA 11.0 on windows locally).

PR #321, opened 2020-06-22
CUDAScanCompaction/CUDAScatter are nolonger singletons

### LINKED ISSUE #232: Move CUDA Singletons into CUDAAgentModel
It was decided in discussion that the various singletons which own CUDA data should be moved into the CUDAAgentModel, in part to support having multiple CUDAAgentModels in the same application (although this will require further changes / threading too). 

This will also resolve the cudaFree in static destructors issue, which results in cuda errors, as the cuda free can instead be handeled by the CudaAgentModel which is not static. 

Cuda calls in destructors for non-static instances appears to work without issue, although various stack overflow posts broadly state that cuda command should not be issued in destructors.

Robadob:
They are now part of `CUDAAgentModel::singletons` for convenience, however they aren't implemented as global singletons.

Due to how widely `CUDAScatter` is used, many functions now require the arguments `CUDAScatter &scatter, const unsigned int &streamId`. In future, when streams are being used properly, it may be worth restructuring all this so that they're all packaged within a single `StreamData` object/structure, given `CUDAScatter` and `CUDAScanCompaction` all operate using stream specific data.

Tests all pass on Windows.

Closes #232

### DIFF
--- a/include/flamegpu/gpu/CUDAAgent.h
+++ b/include/flamegpu/gpu/CUDAAgent.h
@@ -22,6 +22,7 @@
 #include "jitify/jitify.hpp"
 #endif
 
+class CUDAScatter;
 class CUDAFatAgent;
 struct VarOffsetStruct;
 /**
@@ -75,9 +76,12 @@ class CUDAAgent : public AgentInterface {
      * Copies population data from the provided host object
      * To the device buffers held by this object (overwriting any existing agent data)
      * Also updates population size, clears disabled agents
+     * @param scatter Scatter instance and scan arrays to be used (CUDAAgentModel::singletons->scatter)
+     * @param streamId The index of the agent function within the current layer
      * @param population An AgentPopulation object with the same internal AgentData description, to provide the input data
+     * @note Scatter is required for initialising submodel vars
      */
-    void setPopulationData(const AgentPopulation& population);
+    void setPopulationData(const AgentPopulation& population, CUDAScatter &scatter, const unsigned int &streamId);
     /**
      * Copies population data the device buffers held by this object
      * To the hosts object (overwriting any existing agent data)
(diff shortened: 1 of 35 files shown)


Robadob:
@ptheywood 

Cannot build flamegpu2 lib in release on Visual Studio 2019.

This is coming from `Spatial2D.cu`, whereby the only cub usage is `cub::DeviceScan::ExclusiveSum`.

Could this be related to #303 which fixed the earlier cub issue?
Why do you think Windows CI is passing. (I'm on CUDA 10.1 this machine).

```
1>C:\Users\Robadob\fgpu2_dev\externals\cub\util_type.cuh(891): error C2993: 'T': is not a valid type for non-type template parameter '__formal'
1>C:\Users\Robadob\fgpu2_dev\externals\cub\util_type.cuh(909): note: see reference to class template instantiation 'cub::BinaryOpHasIdxParam<T,BinaryOp>' being compiled
1>C:\Users\Robadob\fgpu2_dev\externals\cub\util_type.cuh(891): error C2065: '__T0': undeclared identifier
1>C:\Users\Robadob\fgpu2_dev\externals\cub\util_type.cuh(891): error C2923: 'std::_Select<__formal>::_Apply': '__T0' is not a valid template type argument for parameter '<unnamed-symbol>'
```

ptheywood:
Can't see how #303 is related, it had nothing to do with CUB. 

No idea why you're getting that error subject to includes and only in this branch(?) for one file. Might go away once #316 is merged and CUB is updated regardless. 

Also its not obvious why that error would only occur in release, unless there is a release only define that is breaking it somehow?

Robadob:
> Can't see how #303 is related, it had nothing to do with CUB.

I thought that was the fix for Paul's `cub::exclusiveSum` bug, that's what i was trying to refer to.

> No idea why you're getting that error subject to includes and only in this branch(?) for one file. Might go away once #316 is merged and CUB is updated regardless.

It's only in release in my local configuration (CI was passing fine). I don't normally build release, so master is possible affected too.

PR #322, opened 2020-06-23
Example: Sugarscape

### LINKED ISSUE #67: Example: Sugarscape
Implement the sugarscape model. 

Ideally this will require a discrete environment / agents.

Robadob:
Porting the Sugarscape example from FLAMEGPU1 to demonstrate SubModels.

As we don't have proper dicrete visualisation support, it visualise tightly packed cubes with colors based on visualisation rules.

Notable changes:
* Global function condition has been replaced with a submodel.
* Agent variable `state` has been renamed `status` (similar named throughout have also been updated for clarity).
* Agent's now explicitly hold their cell coordinates as variables (array var `pos`)
* A random initialisation function is provided. This is based off a combination of the FGPU1 init file and an understanding of how the environment exists as hotspots, rather than fully RNG.
* A bug from the original FLAMEGPU1 version, causing bias inside `movement_request`, has been corrected.
* A commented out random value from the original FLAMEGPU1 version, causing bias inside `movement_response`, has been reverted.
* Cells now have individualised `env_max_sugar_level` values, this allows hotspots to persist through the simulation.
* Various configuration parameters have been tuned to provide a more compelling visualisation.
* Occupied cells have their `env_sugar_level` set to -1, to facilitate visualising occupied cells over sugar level. (Previously they would be 0)

Note:
* The model still lacks birth, so it is not possible for more agents to be born. Due to agents (farmers?) not being represented by FLAMEGPU agents, birth would necessitate agent ids being inherited from the location whenever birth occurs (this includes existing agents changing their id to location on movement).

Closes #67 

*Having two discrete examples, may be time to start considering how we will visualise discrete models. properly*

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,7 +56,8 @@ option(BUILD_EXAMPLE_CIRCLES_SPATIAL3D "Enable building examples/circles_spatial
 option(BUILD_EXAMPLE_GAME_OF_LIFE "Enable building examples/game_of_life" OFF)
 option(BUILD_EXAMPLE_HOST_FUNCTIONS "Enable building examples/host_functions" OFF)
 option(BUILD_EXAMPLE_ENSEMBLE "Enable building examples/ensemble" OFF)
-                       
+option(BUILD_EXAMPLE_SUGARSCAPE "Enable building examples/sugarscape" OFF)
+
 option(BUILD_SWIG_PYTHON "Enable python bindings via SWIG" OFF)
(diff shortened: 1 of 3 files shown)


Robadob:
Quick video of the model here: https://youtu.be/qSTfFCgXd8E

ptheywood (approved):
Vis works well after tweaking near/far plane values. 

Can't really comment on whether or not the implementation of sugardscape is correct, but the vis looks sensible.

PR #325, opened 2020-06-26
Support for multiple CUDAAgentModel instances of same ModelDescription.

### LINKED ISSUE #324: Multiple CUDAAgentModel instances with the same ModelDescription
Currently, you can't have two active `CUDAAgentModel` which share the same model description.

This limitation is due to `EnvironmentManager`, as properties are attached to their model name. Hence, two models with same name would share properties.

The ideal solution to this would be to assign `CUDAAgentModel's` instance index's (a function static counter), which are used in place of the model's name.

Robadob:
Closes #324 

**Edit:** This resolves the limiting factor with `EnvironmentManager`, however on closer inspection Curve is not setup to work with two instances of `CUDAAgentModel`, let alone two instances with the same `ModelDescription`.

Whilst it is setup with  the concept of namespace, I don't believe that's being used during agent function execution. This isn't really a problem at the moment, given the calls to `CUDAAgentModel::step()` will be synchronous, but needs fixing in future (although `Curve` singleton would actually need some internal changes to make it thread safe too).

### DIFF
--- a/include/flamegpu/io/factory.h
+++ b/include/flamegpu/io/factory.h
@@ -36,11 +36,25 @@ inline std::string getFileExt(const std::string& s) {
 */
 class ReaderFactory {
  public:
-    static StateReader *createReader(const std::string &model_name, const std::unordered_map<std::string, std::shared_ptr<AgentPopulation>> &model_state, const std::string &input) {
-        std::string extension = getFileExt(input);
+    /**
+     * Returns a reader capable of reading 'input'
+     * Environment properties will be read into the Simulation instance pointed to by 'sim_instance_id'
+     * Agent data will be read into 'model_state'
+     * @param model_name Name from the model description hierarchy of the model to be loaded
+     * @param sim_instance_id Instance is from the Simulation instance to load the environment properties into
+     * @param model_state Map of AgentPopulation to load the agent data into per agent, key should be agent name
+     * @param input Filename of the input file (This will be used to determine which reader to return)
+     * @throws UnsupportedFileType If the file extension does not match an appropriate reader
+     */
+    static StateReader *createReader(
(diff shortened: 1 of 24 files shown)


include/flamegpu/io/factory.h
```diff
@@ -36,11 +36,16 @@ inline std::string getFileExt(const std::string& s) {
 */
 class ReaderFactory {
  public:
-    static StateReader *createReader(const std::string &model_name, const std::unordered_map<std::string, std::shared_ptr<AgentPopulation>> &model_state, const std::string &input) {
+    static StateReader *createReader(
+        const std::string &model_name,
+        const unsigned int &instance_id,
```
ptheywood, line 41: Is `instance_id` the instance ID of the StateReader, or of the Simulation?

This probably applies to every use of `instance_id`.

Robadob: Always `Simulation` instance, this can be resolved with comments. I updated all the comments that already existed. But some places didn't have comments to begin with.

I'll add comments to the IO classes.

Robadob: (It's required because that's used to access the Environment from within the IO method.)

Robadob: I wasn't sure if Model name was strictly necessary there, so I left it in.

ptheywood: would prefer it to be `simulation_instance_id` in that case tbh, Comments work so fine without, just not keen on the ambiguity of it.

Robadob: I've updated IO to `sim_instance_id` throughout, i'm less concerned about the rest. Just waiting for tests to build incase i broke something with refactor tools.

PR #327, opened 2020-06-26
Remove Appveyor and Travis as CI providers

ptheywood:
Removes Appveyor and Travis as CI providers, as Github Actions provides both Linux and Windows builds, with much faster turnaround due to concurrency. 

+ [x] Github integration with these services will need to be removed from the repository by @mondus (Appveyor still runs even if you delete appveyor.yml)
    + **Important** that this will disable appveyor for all branches, not just this one - i.e. swig (it's how appveyor works unfortunately). Other branches will want rebasing at this point (which they probably should be) 
    + The swig branch once rebased will need the GHA workflows adjusting to enable swig building.


mondus:
Wait until #291 can be rebased (after pytests are completed).

PR #328, opened 2020-06-27
Automatic cudaDeviceReset() on last CUDAAgentModel destruction.

### LINKED ISSUE #150: cudaDeviceReset
`cudaDeviceReset` is useful to flush data and clear up memory.

Ideally this would need to be in a destructor of the constant singleton, or similar, however order of destrcutors and the cuda context is not defined, so it is unsafe to issue any cuda calls in a destructor 

This will also invalidate CuRVE, so we need to think about this.

Also required for `cuda-memcheck`

Robadob:
Updates `CUDAAgentModel` to track active instances, when last one destructs `cudaDeviceReset()` is triggered.

This feature is disabled during tests, because it makes them take approx 15x longer (because every test that uses cuda requires it's own cuda init).

Closes #150

**TODO (in order):**
- [x] Test on Linux
- [x] Modify `tests/helpers/main.cu:11` to `false`
- [ ] Potentially change how `CUDAAgentModel::AUTO_CUDA_DEVICE_RESET` flag is managed. e.g. would we prefer a single static struct with all global flags?
- [x] If this is merged after #321, this won't build until `CUDAAgentModel.cu:141` is deleted. (don't think anything else will break)
- [x] Rebase
- [x] Squash merge (copy everything above *TODO** as message)


Robadob:
So running tests through debugger, first exception that gets caught occurs in `DeviceRTCAPITest.AgentFunction_getset`.

[`cudaGetGlobalPtr()`](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MODULE.html#group__CUDA__MODULE_1gf3e43672e26073b1081476dbf47a86ab) causes an access violation reading `0xFFFFFF` within `nvcuda.dll`.

This is triggered via `CUDAAgentModel::setRTCEnvironmentVariable()`), in particular it's coming from `initRTC()`, whereby environment properties are being set. The only environment property being `_stepCount`.

So the cause for this isn't too clear.

Preceding tests for RTC all pass, however the directly preceding test concerns `_stepCount` so it could be a hangover from failing to purge/reset something related to RTC environment.

image

Robadob:
> Before this, all tests within `AgentEnvironmentTest` fail, with the environment reads all returning 0.

I've now traced the cause of this, and the test is to blame. The value read from env is stored in an external `__device__` variable. This is then checked **AFTER** the `CUDAAgentModel` has been destroyed (due to scoping and the testing fixture). Shouldn't be too hard to correct. Doesn't explain the RTC test failure.

Robadob:
Still not fixed the initial RTC error, but I've got a different one that I've reduced down to tests.

`DeviceRTCAPITest.AgentFunction_cond_non_rtc` runs first and passes
`DeviceRTCAPITest.AgentFunction_cond_rtc` then runs and fails at `CUDAAgentModel::RTCSetEnvironmentVariable()` (i think this is same above).

I'm using `tests_dev` with all the other tests from that suite disabled.

ptheywood:
Tests all pass under linux, with an expected increase in runtime in line with mentioned in #298 

Release on linux each test takes ~ 130ms rather than <~ 2ms:

```
[==========] 550 tests from 43 test suites ran. (110046 ms total)
[  PASSED  ] 550 tests.
```

Compared to current `master`

```
499 tests in [==========] 499 tests from 39 test suites ran. (49793 ms total) in master.
```

Of which most of the time is in the RTC tests

```
[----------] 13 tests from DeviceRTCAPITest (48994 ms total) 
```

PR #347, opened 2020-07-06
Extends readme to mention standalone project

### LINKED ISSUE #261: Standalone Models
Users should not be expected to have thier examples in the examples directory (although this should still be an option for anyone wishing to contribute a new example).

Instead, users code should be able to be tracked in it's own repository, with a simple method of defining their example (CmakeLists.txt) with the main flamegpu2 library either:

+ installed on the system globally (/per user)
+ built from a local directory
    + via submodule?
    + via cmake download

This will probably need to take the form of a new repository as a bare-bones example which users can copy as a starting point.

i.e. the following files should be sufficient - essentiallly and example.

+ `LICENCE`
+ `Readme.md`
+ `CMakeLists.txt`
+ `src/`
    + `main.cu`
+ `tests/`
    + `...`

The contents of `CMakeLists.txt` will be the difficult bit / main difference between this and examples in the main repository - as flamegpu2 will need to be found for `common.cmake` to be included and usable.

We will likely want several versions of this for the different ways to work with flamegpu2, although ideally with minimal differences.

+ C++ 
+ Python
+ **Maybe**  C++ RTC, but probably not (no one should use this tbh)

Robadob:
Closes #261

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,3 +200,6 @@ FLAME GPU 2 supports runtime compilation of agent functions. E.g. specification 
 
 1. `FLAMEGPU2_INC_DIR` This should be set to the `\include` directory of the main FLAMEGPU2 directory.
 2. `CUDA_PATH` This may be set when installing CUDA, however if it does not exist or has been changed this then it **must** point to the CUDA installation folder of the CUDA version used to compile FLAMEGPU (e.g. the one detected by CMake).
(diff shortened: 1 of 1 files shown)


ptheywood:
Merging pre-CI as doesn't effect it.

PR #355, opened 2020-07-13
Increase curve max size to 1024

Robadob:
Tests pass on Windows x Release
Relates to #356


Robadob:
Where an agent function inputs and outputs message of same name, there is a collision. This needs addressing.

ptheywood:
> Where an agent function inputs and outputs message of same name, there is a collision. This needs addressing.

We could disable that functionality for now? I don't think that it's supported in Flamegpu 1 (although no reason it shouldn't be supported IMO).

Robadob:
The spatial messaging tests need adjusting, they currently have the
warning.

Robadob:
I fixed the spatial message warning issue, so I'll merge this now.

PR #361, opened 2020-07-14
Json IO support

### LINKED ISSUE #357: JSON IO Support
Needs adding before Alpha release, should be an easy job.

Which JSON library to use? why?

Began working in branch `json_io`, cmake is setup with rapidjson but no code has been written.

Robadob:
JSON IO handled by RapidJSON SAX interface.
Added option to toggle export format between 'pretty' and 'compact'.
Adds a config block to the IO formats, this can be used to pass config arguments (except for input file).
Updates tests to cover JSON and config block.
The ability to load config from file, necessitated changing `Curve`/`EnvironmentManager`, so that they copy data to the device as late as possible. This change makes `Curve` and `EnvironmentManager` copy all data to device on demand via respective `updateDevice()` methods. Further work in future, may take this further, by unifying their block of constant cache memory so only a single memcpy is required.
Fixed a bug, whereby mapped environment properties within a submodel would not be available within RTC agent functions.
Note, none of these changes have been tested with RTC (as SWIG is not yet merged).

Closes #357 

Example output:
*(The new config and stats blocks added to XML follow the same pattern, however nested inside the main `states` block)*
```json
{
	"config": {
		"simulation": {
			"input_file": "",
			"steps": 0,
			"timing": false,
			"random_seed": 1594808855,
			"verbose": false
		},
		"cuda": {
			"device_id": 0
		}
	},
	"stats": {
		"step_count": 0
	},
	"environment": {
		"uint64_t": 15,
		"int32_t_a": [
			16,
			0,
			1
		],
		"_stepCount": 0
	},
	"agents": {
		"a": {
			"default": [
				{
					"uint64_t": 4,
					"uint8_t": 10
				},
				{
					"uint64_t": 5,
					"uint8_t": 11
				}
			]
		},
		"b": {
			"2": [
				{
					"double": [
						2.0,
						0.0,
						1.0
					],
					"float": [
						1.0,
						0.0,
						1.0
					]
				},
				{
					"double": [
						2.0,
						1.0,
						1.0
					],
					"float": [
						1.0,
						1.0,
						1.0
					]
				}
			]
		}
	}
}
```


Robadob:
> Assuming that the config options in the input file are over-ridden by extra cli args?

Yes iirc

PR #364, opened 2020-07-17
Add runtime arg '--console', this will disable the visualisation.

### LINKED ISSUE #231: Visualisation target in visual studio
In visual studio (and maybe makefiles), add a custom target for visualisation_release, visualisation_debug builds, if cmake was configured with VISUALISATION=ON (for dependencies).

This would need setting in a similar vein to the profile generator target.

Whether or not we want this on linux is questionable, having sepearate cmake build directories is trivial, although the main library is built multiple times in this case.

Robadob:
Adds command line argument `-c` or `--console` to disable visualisation.
This is represented by the `Simulation::Config::console_mode` variable, which can also be managed programatically and via file import.
This is locked to true when not building with `VISUALISATION` enabled.

Closes #231 

-----

Slight issue, if a user activates visualisation before parse args the vis will be open and then close when parse args is called. Only fix for this would be to make parse args part of the `CUDAAgentModel` constructor, so it was always executed at the start. This would however prevent users from overriding the default config as they can currently do.

Compromise could be to allow args to be passed to constructor, but leave the manual call there too. This would hopefully encourage use via the constructor to be the default behaviour.

### DIFF
--- a/examples/circles_bruteforce/src/main.cu
+++ b/examples/circles_bruteforce/src/main.cu
@@ -125,6 +125,7 @@ int main(int argc, const char ** argv) {
      */
     NVTX_PUSH("CUDAAgentModel creation");
     CUDAAgentModel cuda_model(model);
+    cuda_model.initialise(argc, argv);
     NVTX_POP();
 
     /**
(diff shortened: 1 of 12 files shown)


Robadob:
I've now applied the change, whereby initialise can be called directly for constructor.

@ptheywood Please can you check this works on Linux pete (just run any example and doesnt crash when passed runtime args, and maybe check that `CUDAAgentModel::applyConfig_derived()` is called).

The rule for calling virtual functions from constructors is that they will be `statically resolved`. I understand, how this works if I were to call a virtual function in the constructor of the base class, but I'm not sure if that means calling non-static base class method, which calls virtual will be fine (i googled and couldn't find a clear matching example). So best to check it's not a visual studio quirk that it works.

PR #365, opened 2020-07-21
Bucket Messaging format.

### LINKED ISSUE #82: Messaging: Generic Sorted Messages

Robadob:
Messages are assigned a key which is an integer from a fixed range (This is currently `typedef int IntT`, as I was unsure if it was worth fixing it to a specific type or not).

Multiple messages may share a key, similarly keys may have no messages.

This is based off master, so will require reformating in order to work with Swig.

Closes #82

### DIFF
--- a/include/flamegpu/runtime/flamegpu_host_agent_api.h
+++ b/include/flamegpu/runtime/flamegpu_host_agent_api.h
@@ -93,7 +93,7 @@ class HostAgentInstance {
     OutT sum(const std::string &variable) const;
     /**
      * Wraps cub::DeviceReduce::Min()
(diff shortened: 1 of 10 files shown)


mondus:
Todo: Add bucket lookup ranges

Robadob:
Have just made and squashed the required changes. This can be merged once it passes CI.

PR #369, opened 2020-07-31
Remove unused doxygen template

ptheywood:
Removes the Atomineer Doxygen templates, which are not being used (and don't appear to have ever been used?)


Robadob:
I wasn't even aware of them, when i ported doxygen i just converted the doxygen script which Mozghan had created to the CMAKE rules.

PR #371, opened 2020-08-24
rtc_boids and other profiling changes

Robadob:
*This branch started out as adding an rtc version of boids but has grown, so I've compiled a changelog.*

**Changelog:**
* Added `boids_rtc_bruteforce` and `boids_rtc_spatial3D` examples, these models are identical to their non-rtc variants.
* Removed `rtc_example`
* Removed `new` and `delete` from `agent_function_wrapper` and `agent_function_condition_wrapper`, api object is now allocated in local not global.
* Added `getVariableByHash_ldg()` and `getVariable_ldg()` to CURVE. These are used by message bruteforce (also affects RTC curve header).
* Improve the arguments passed to RTC, so that they match the non-rtc compilation settings.
* Add CMake option `EXPORT_RTC_SOURCES`. If enabled, this sets the macro `OUTPUT_RTC_DYNAMIC_FILES` causing RTC dynamic files to be exported to the working directory at runtime.
* Refactor CURVE, to improve RTC variable access performance (removal of a redundant branch).
* Optimise `end()` iterator for spatial messaging to reduce instructions on comparison.

Normal tests pass on Windows (seatbelts on).
Normal tests pass on Linux (Petes machine, seatbelts on/off).
Swig tests pass on Windows (seatbelts on)
Normal tests fail on Mavericks. All tests failing are environment, not sure why.


Robadob:
So much got added to this branch than intended. Not sure if we want to keep the __ldg change (but I think removing it will also require reverting 48b863).

Otherwise it seems mergeable.

CMakeLists.txt
```diff
@@ -44,12 +44,14 @@ option(BUILD_FLAMEGPU2 "Enable building FLAMEGPU2 library" ON)
 option(BUILD_ALL_EXAMPLES "Enable building examples" ON)
 
 # Options to enable building individual examples, if BUILD_ALL_EXAMPLES is off.
-option(BUILD_EXAMPLE_BOIDS_BRUTE_FORCE "Enable building examples/boids_brute_force" OFF)
-option(BUILD_EXAMPLE_CIRCLES_BRUTE_FORCE "Enable building examples/circles_brute_force" OFF)
-option(BUILD_EXAMPLE_CIRCLES_SPATIAL_3D "Enable building examples/circles_spatial_3d" OFF)
+option(BUILD_EXAMPLE_BOIDS_BRUTEFORCE "Enable building examples/boids_bruteforce" OFF)
+option(BUILD_EXAMPLE_BOIDS_SPATIAL3D "Enable building examples/boids_spatial3D" OFF)
+option(BUILD_EXAMPLE_BOIDS_RTC_BRUTEFORCE "Enable building examples/boids_rtc_bruteforce" OFF)
+option(BUILD_EXAMPLE_BOIDS_RTC_SPATIAL3DE "Enable building examples/boids_rtc_bruteforce" OFF)
```
ptheywood, line 50: Rogue trailing E

Robadob: and the description is wrong..

PR #374, opened 2020-08-26
DeviceEnvironment now uses offsets from symbol rather than direct ptr.

Robadob:


ptheywood (approved):
Builds and passes tests under linux

Issue #375, opened 2020-09-09
Missing Swig Functionality


ptheywood:
Although Swig is now in master, some things are still missing

+ [ ] Alternate Array apporach
+ [ ] Missing Tests
+ [ ] General tidying
+ [ ] Ctest (#267)
+ [ ]

Robadob:
IO tests and BucketMsg are working in this branch: https://github.com/FLAMEGPU/FLAMEGPU2_dev/tree/swig_tests_a

ptheywood:
Generated lists of tests via gtest and pytest, with some manipulation to get the listings relatively close to each other. 

https://gist.github.com/ptheywood/cb1d2d20d4c446d2fb135d4ffb400f7c

Commands used:

```
./bin/linux-x64/Release/tests --gtest_list_tests
```

```
pytest --collect-only | sed -e 's/TestCaseFunction//g' | sed -e 's/UnitTestCase //g' | sed -e 's/[<>]//g' | sed -e 's/   test_/   /g'
```

Robadob:
Full list of tests missing from each suite (also accounting for bucket/IO tests not currently on master)
```
DeviceExceptionTest::71
RTCDeviceExceptionTest::17
DeviceRTCAPITest::14
TestCUDAAgentModel/TestSimulation::5
TestCUDASubAgent::29
SubEnvironmentManagerTest::10
SubEnvironmentDescriptionTest::5
SubAgentDescriptionTest::5
HostFunctionTest::18
DeviceAPITest::2
EnvironmentManagerTest/1::-1
HostAgentSort::6 (not possible in Py?)
HostEnvironmentTest::6
HostRandomTest::4 (char methods)
HostReductionTest::37 (20 not possible, 17 histogram/exception)
HostMiscTest::9 (not valid for python?)
Spatial2DMsgTest::1 (read empty)
Spatial3DMsgTest::1 (read empty)
TestMessage_BruteForce::1 (read empty)
TestMessage_Array::1 (optional none)
TestMessage_Array3D::2 (optional none [appears twice in C list?])
ModelDescriptionTest::1 (submodel...)
LayerDescriptionTest::1 (submodel...)
TestDependencyVersions::2
TestUtilNVTX::1
TestUtilComputeCapability::3
TestMessage::1
```

Robadob:
Was agreed in meeting that further SubModel test duplication is unnecessary.

Robadob:
[...]

PR #376, opened 2020-09-10
Swig support for MsgBucket and IO tests

Robadob:
Relevant to #375


Robadob:
TODO: Add in swig tests for SubAgentDescription etc?

Robadob:
I will merge this as soon as CI passes (as I think that was agreed in Weds meeting).

PR #377, opened 2020-09-11
Rename CUDAAgentModel to CUDASimulation

ptheywood:
Renames `CUDAAgentModel` to the more accurate/appropriate `CUDASimulation`.

No functional changes, just renaming.

This can then subsequently be split in the future to `CUDASimulator` and `CUDASimulation` (the same for `Simulation` abstract classes), to then avoid changing the interface (much) to enable ensembles.  

C and python tests all pass under linux.


Robadob (approved):
Not ran tests, but can't see any reason they would fail.

Probably need to update userguide too as that will mention `CUDAAgentModel` in a bunch of places.

mondus:
Todo before merge
- [ ] Run python swig generators and check python tests
- [X] Rename the TestCudaAgentModel test

ptheywood:
Double checked that all tests (c++ and python) pass under linux, so ready to merge this although it will confict /  cuase compilation issues with other active PRs.

PR #379, opened 2020-09-11
In-Layer Concurrency

### LINKED ISSUE #189: Concurrency within Layers
With FLAME GPU 2 we should be able to achieve concurrency using streams in various places. 

E.g. concurrent agent functions within layers. 

building a DAG with good dependency analysis rather than explicit layers would allow this to be taken further. 

This will need to account for which agent functions can be executed concurrently. 

I.e. a layer with 2 agent functions, for 1 agent type with a common state as a common input/output. This should not be executed concurrently / be allowed within a single layer.

CuRVE will need adjusting to make it non-blocking, streams will be needed everywhere appropriate, and maybe events. Care will need to be taken with `__constant__` cache. 

### Rough plan:

+ Restructure `CUDAAgentModel::step()` into smaller parts
    + See below for possible outline 
+ Construct streams upfront, based on maximum required in any given layer.
+ Ensure all operations within a layer are async with streams.
+ Make as  much async as possible with streams for pre/post-layer data management?
+ ...

### `step` refactor outline
Details the current logical structure, how a serial refactor could look and how a parallel refactor could look

#### Current step function 
effectively, some difference between logic + impl

```cuda
current_step_effective(){

    // init singletons

    // step value resetting (message lists etc.)

    // Iterate layers
    for(auto & layer: layers){

        // reset flags/counters/singletons

        // Iterate functions
        for(auto & func: layer->functions){

            if(condition){
                // register variables for condition

                // do condition (sync), map of functions that need executing

                // unmap condition variables
            }

            if(!condition or condition_passed){
                // map agent

                if(in_message){
                    // map in_message
                }
                if(out_mesasge){
                    // map out_message
                }

                // exec agent function

                // post-message stuff
                if(in_message){
                    // un map in_message + post process
                }

                if(out_mesasge){
                    // un map out_message + post process
                }

                // death

                // agent state transitions

                // birth

                // unmap
            }
        }

        // host layer functions

        // process host agent birth

    }

    // step functions

    // process host agent birth

    // exit conditions

    // step_counter++

    // tidy up

    // return success of exit conditions

}
```

#### Sequential refactored version strucutre

Aproximate. Synchronous ops (memcpy) will block parallelism even if streams used correctly.

```cuda
step_restructure_sync(){

    step_initialisation() // init singletons + reset any values 

    // Iterate layers
    for(auto & layer: layers){

        layer_pre(layer) // reset any flags our counters on a per layer basis

        // Iterate functions
        for(auto & func: layer->functions){


            do_function_condition(func) // register, do, unregister. Store /return success

            if(run_function){
                do_function(func); // map agent/messages, exec, unmap/postprocess/death/birth 
            }
        }

        do_host_functions(layer);// host functions
        processHostAgentCreation(); // Process host agent birth from layer

    }


    do_step_functions() // step functions
    processHostAgentCreation(); // Process host agent birth from step

    exit_required = do_exit_conditions() // exit conditions

    post_step() // tidy up, step_counter++

    return !exit_required;

}
```

#### Parallel refactored which might enable concurrency

To get concurrency on windows need to issue async commands wide rather than tall. 

Note that async memcpy will require mallocManaged(pinned=true) host memory for data being copied (to avoid paging). This might have to be a secondary step to kernel concurrency.
Might be better to use events in the default stream for sync than device - in case any longer running async ops? (long term opt.)

```cuda
step_restructure_async(){

    step_initialisation() // init singletons + reset any values 

    // Iterate layers
    for(auto & layer: layers){

        layer_pre(layer) // reset any flags our counters on a per layer basis

        // do_function_condition_conc
        {
            map_conditions_sync(layer->functions) // s
            function_conditions_async(layer->functions, streams)
            conditon_results_async(layer->functions, streams) // issue async memcpies in each stream together for wddm concurrency. Async might not help with 4B copies, as latency costs more than data transfer. probably better to just do a 4b*function copy.
            unmap_conditions_sync() // This has to be synchrnonous currently
        }

        // container of functions to exec / store internally 
        funcs_to_exec = layer->functions if not function_condtion_fail
        // do_agent_functions_conc
        {
            map_agent_functions_sync(funcs_to_exec)
            message_prep_sync(funcs_to_exec) // This might be able to be async, need to look at mesage impl.
            agent_func_async(funcs_to_exec, streams)
            message_post_sync(funcs_to_exec); // probably includes memcpy, do them together after kernels have been launched.
            unmap_agent_functions_sync(funcs_to_exec)
            death_sync(funcs_to_exec) // Might get some concurrency in here, but the func needs to sync. wrt streams involved /
            agent_state_transition_sync(funcs_to_exec) // Might get some concurrency in here, but the func needs to sync. wrt streams involved /
            agent_state_transition_sync(funcs_to_exec) // Might get some concurrency in here, but the func needs to sync. wrt streams involved /
            birth_sync(funcs_to_exec) // Might get some concurrency in here, but the func needs to sync. wrt streams involved /
        }

        do_host_functions(layer);// host layer functions

        processHostAgentCreation(); // Handle any agent output from host layer functions, could run in parallel no guarantee  user isn't creating race conditions.

    }


    do_step_functions() // step functions
    processHostAgentCreation(); // Handle any agent output from step functions


    exit_required = do_exit_conditions()

    post_step() // tidy up, step_counter++

    return !exit_required;

} 
```

## Testing Concurrency
Currently unsure how we could test concurrency programatically. 

1.Define a test model with multiple agent types / multiple states (all cases which should have concurrency
2. Need to add cuda events before and after the agent func in appropriate streams
3. Afterwarrds, using the start and end times of the kernels, there *should* be some overlap in the windows. 
4. Alt add a way of running without concurrency (might be useful for debugging anyway?)

## RTC vs cuda runtime api.
There is likely to be a decent bit of duplication in here for rtc support. It might be better to abstract this logic out into a separate class. 

I.e. have a `CUDAAgentFunction` class which takes a `functionDesctiption`, instanciated on creation of the `CUDAAgentModel`? This would do the if normal else rtc kidn of logic and jsut present methods to `map()`, `pre()`, `run(stream)`, `unmap()`, `post()`, etc?

Or maybe make `CUDAAgentFunction` abstract, with concrete specialisations for normal and rtc versions - this would possibly be a way of preventing RTC from being left behind - new features must go into the base class, leading to RTC not being usable until all features implemented.

ptheywood:
Implement and detect in-layer concurrency via streams. 

This is now ready for review. It is not compelte, but is most of the way there and provides the majority of the intended benefit, although things can be improved.

Need to check that it has worked as intended on Windows (i.e. do the tests pass) in Release mode.

## Issues

Closes #189

Outstanding improvements detailed in #449

## Completed bits 

+ [x] add `CUDAConfig::inLayerConcurrency` property to prevent stream-based concurrency
    + required for testing, and potentially useful for debugging.
+ [x] Programatic Step Timing, init, exit and RTC timing.
    + [x] Plus output via `-t/--timing`?
    + [x] tests
+ [x] Simulate() NVTX Range
+ [x] NVTX Improvements `CUDASimulation`
+ [x] Refactor Step() into smaller chunks.
    + [x] Make Init functions user-callable
    + [x] Make Exit functions user-callable
+ [x] Better CUDASteam management/re-use
+ [x] Concurrent post-processing via streams (PBM construction, Scattering, CUB/Thrust calls...)
+ Add tests which detect concurrency, through the average runtime of steps:
    + [x] Naive agent function, different agent types
    + [x] Message Output to independant lists
    + [x] Message Input from independant lists
    + [x] Message Input from the same message list
    + [x] Optional Message Output
    + [x] Agent Birth (independant agent types)
    + [x] Agent Death
    + [x] Device Exception check (only runs with `SEATBELTS` on)
+ Add RTC tests which detect concurrency, through the average runtime of steps:
    + [x] Naive agent function, different agent types
    + Others (see issue)
+ [x] Check synchronisation is still performed as appropriate
+ [x] Disable performance related tests in debug mode.

This has primarily been tested / implemented under Linux. 
+ [x] Validate on Linux 
+ [x] Validate on Windows (WDDM), may be discrepancy.

Ideally the problem sizes should be small enough that individual agent functions will not fill the device, but this requires blocksize information to accurately target. Rather than exposing this, just use very small populations.

### Simple test case
4 agent types, 1 function each in the same layer, with small populations (4k?) each doing lots of pointless global memory accesses and floating point ops to ensure larger run time so that concurrency is (almost) guaranteed to occur. 
Time each function N times, and compare average with and without streams enabled.  

nvprof output for the simple test case, with `NO_SEATBELTS=OFF` showing that concurrency is already being achieved. My test correctly detects this! ~ 70ms with streams, ~ 200ms without
image

### DIFF
--- a/src/flamegpu/model/ModelData.cpp
+++ b/src/flamegpu/model/ModelData.cpp
@@ -1,4 +1,5 @@
 #include <iostream>
+#include <algorithm>
 
 #include "flamegpu/model/ModelData.h"
 
@@ -172,3 +173,10 @@ bool ModelData::hasSubModelRecursive(const std::shared_ptr<const ModelData> &sub
     return false;
 }
 
+ModelData::size_type ModelData::getMaxLayerWidth() const {
+    unsigned int maxWidth = 0u;
+    for (auto &layer : layers) {
+        maxWidth = (std::max)(maxWidth, static_cast<ModelData::size_type>(layer->agent_functions.size()));
+    }
+    return maxWidth;
+}
--- a/include/flamegpu/exception/FGPUDeviceException.h
+++ b/include/flamegpu/exception/FGPUDeviceException.h
@@ -22,8 +22,8 @@ class DeviceExceptionManager {
      * Free all device memory
      */
     ~DeviceExceptionManager();
-    DeviceExceptionBuffer *getDevicePtr(const unsigned int &streamId);
-    void checkError(const std::string &function, const unsigned int &streamId);
+    DeviceExceptionBuffer *getDevicePtr(const unsigned int &streamId, const cudaStream_t &stream);
+    void checkError(const std::string &function, const unsigned int &streamId, const cudaStream_t &stream);
 
  private:
     /**
--- a/include/flamegpu/exception/FGPUException.h
+++ b/include/flamegpu/exception/FGPUException.h
@@ -419,5 +419,9 @@ DERIVED_FGPUException(VersionMismatch, "Versions do not match");
  * Defines an error reported when the expect input/output file path does not exist
  */
 DERIVED_FGPUException(InvalidFilePath, "File does not exist.");
+/*
+ * Defines an error indicating that a CUDAEventTimer was queried without being synced.
+ */
+DERIVED_FGPUException(UnsycnedCUDAEventTimer, "Elapsed time requested for Un-synced CUDAEventTimer");
 
 #endif  // INCLUDE_FLAMEGPU_EXCEPTION_FGPUEXCEPTION_H_
--- a/include/flamegpu/gpu/CUDAAgent.h
+++ b/include/flamegpu/gpu/CUDAAgent.h
@@ -78,7 +78,7 @@ class CUDAAgent : public AgentInterface {
      * @param population An AgentPopulation object with the same internal AgentData description, to provide the input data
      * @note Scatter is required for initialising submodel vars
      */
-    void setPopulationData(const AgentPopulation& population, CUDAScatter &scatter, const unsigned int &streamId);
+    void setPopulationData(const AgentPopulation& population, CUDAScatter &scatter, const unsigned int &streamId, const cudaStream_t &stream);
     /**
      * Copies population data the device buffers held by this object
      * To the hosts object (overwriting any existing agent data)
@@ -113,7 +113,7 @@ class CUDAAgent : public AgentInterface {
      * @param streamId The index of the agent function within the current layer
      * @see CUDAFatAgent::processDeath(const unsigned int &, const std::string &, const unsigned int &)
      */
-    void processDeath(const AgentFunctionData& func, CUDAScatter &scatter, const unsigned int &streamId);
+    void processDeath(const AgentFunctionData& func, CUDAScatter &scatter, const unsigned int &streamId, const cudaStream_t &stream);
     /**
      * Transitions all active agents from the source state to the destination state
      * @param _src The source state
@@ -122,7 +122,7 @@ class CUDAAgent : public AgentInterface {
      * @param streamId The index of the agent function within the current layer
      * @see CUDAFatAgent::transitionState(const unsigned int &, const std::string &, const std::string &, const unsigned int &)
      */
-    void transitionState(const std::string &_src, const std::string &_dest, CUDAScatter &scatter, const unsigned int &streamId);
+    void transitionState(const std::string &_src, const std::string &_dest, CUDAScatter &scatter, const unsigned int &streamId, const cudaStream_t &stream);
     /**
      * Scatters agents based on their output of the agent function condition
      * Agents which failed the condition are scattered to the front and marked as disabled
@@ -134,7 +134,7 @@ class CUDAAgent : public AgentInterface {
      * @note Named state must not already contain disabled agents
      * @note The disabled agents are re-enabled using clearFunctionCondition(const std::string &)
      */
-    void processFunctionCondition(const AgentFunctionData& func, CUDAScatter &scatter, const unsigned int &streamId);
+    void processFunctionCondition(const AgentFunctionData& func, CUDAScatter &scatter, const unsigned int &streamId, const cudaStream_t &stream);
     /**
      * Scatters agents from the provided device buffer, this is used for host agent creation
      * The device buffer must be packed according to the param offsets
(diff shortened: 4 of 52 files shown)


ptheywood:
The offending `memset` blockign `NO_SEATBELTS=OFF` is in `DeviceExceptionManager::getDevicePtr`. 

The reset needs moving before outside the loop. Probably need to make sure that device exceptions work with concurrency too.

Robadob:
No reason they shouldn't. Each stream should have its own output buffer.

ptheywood:
detecting concurrency within the post agent function steps is going to be very difficult, as I can't artificially slow doen these methods (not user defined) so can't get reliable timing info, even if I add extra timers to measure the cost of the post-step phase rather than simulation/layers themselves

I.e. for 512 agents on a 1070, the pbm reconstruction kernel takes ~ 14us, vs ~200us for a message iteration kernel which only contains return true (with seatbelts on). I can't reliably time such a short running kernel, and can't (currently) separate the timing out without (potentailyl) adding runtime costs to the entire library through excessive timing.

src/flamegpu/model/ModelData.cpp
```diff
@@ -172,3 +172,10 @@ bool ModelData::hasSubModelRecursive(const std::shared_ptr<const ModelData> &sub
     return false;
 }
 
+ModelData::size_type ModelData::getMaxLayerWidth() const {
+    unsigned int maxWidth = 0u;
+    for (auto &layer : layers) {
+        maxWidth = std::max(maxWidth, static_cast<ModelData::size_type>(layer->agent_functions.size()));
```
Robadob, line 178: Wrap `std::max` in brackets e.g. `(std::max)`.

VC++ doesn't like it because `windows.h` must be included somewhere, and this declared `min` and `max` as pre-processor macros.

ptheywood:
Now rebased onto current master (d9c3fc534aa8302c02fb2ad37686b4430eefd5c2).

Currently failing the 12 concurreny tests, due to additional syncs added during the rebase that need moving. 
1 additional test failure which also occurs in master on linux (#438)

ptheywood:
Currently getting 2 test failures on windows with `SEATBELTS=OFF`: 

```
[  FAILED  ] 2 tests, listed below:
[  FAILED  ] TestCUDASimulationConcurrency.ConditionConcurrencyAllDisabled
[  FAILED  ] TestUtilCUDAEventTimer.CUDAEventTimer
```

This requires some investigation (but the other new tests passing is promising)

ptheywood:
The test issue has been resolved under WDDM, by a `cudaDeviceSync` prior to the thread sleep (for wddm devices). 

Long term we should check this further, and consider not using cudaEvent_t timers under windows on wddm devices, instead using a lower precision `chrono::steady_clock` timer (util class already provided and used for ensemble timing). 
This should be precise enough for total simulation timing, but may not be too useful on the per step or per layer per step timing levels. 
I'll open a new very low priority issue. 

Otherwise this should be good to merge/review and merge.

PR #383, opened 2020-10-06
Fix ambiguity when specifying agent functions by name.

### LINKED ISSUE #378: Agent Function Names should not need to be globally unique
With the current layer implementation(s) supporting `void addAgentFunction(const std::string &name);` as a way of adding an agent function to a layer, different agents cannot have the same agent function name. 

This is the behaviour in FLAME GPU 1, but it prevents (easy/clean) re-use of a common agent function for multiple agent types. 

IMO it would be nice if this requirement was not required. 

This also appears to hold be the case when the first argument of the `FLAMEGPU_AGENT_FUNCTION` definition is used to add to a layer, as that object has no understanding of which agent type is going to call it.

Robadob:
Also extend `AgentFunctDescription::operator==` to also consider parents (in a minimal fashion, to prevent recursion loop).
Closes #378

### DIFF
--- a/include/flamegpu/model/LayerDescription.h
+++ b/include/flamegpu/model/LayerDescription.h
@@ -81,22 +81,24 @@ class LayerDescription {
     /**
      * Adds an agent function to this layer
      * The agent function will be called during this stage of model execution
-     * @param name Name of the agent function description to execute during this layer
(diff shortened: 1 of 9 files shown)


Robadob:
I haven't ran the Python tests, some of those may also require updating.

PR #386, opened 2020-10-12
Bugfix: NVTX marker was missing semicolon.

Robadob:
Compilation was failing when USE_NVTX was enabled.


ptheywood (approved):
CI won't notice a difference, but will still wait for it before merging.

Issue #253, opened 2020-05-06
Environmental variables for Jitify/RTC


ptheywood:
Currently 2 environment variables are required to jitify/rtc compilation:

+ `CUDA_PATH` pointing to the root of the relevant CUDA version
+ `FLAMEGPU2_INC_DIR` pointint to `flamegpu2_dev/include`

This could be improved by:
+ [ ] also attempting other common environmental variables (`CUDA_HOME`/`CUDA_ROOT` are semi-common alternatives to `CUDA_PATH`)
+ [ ] fallback to trying the baked in path to the CUDA version used to compile the executable (if running on the same machine)
+ [ ] try he location of `include` from compile time (abs path, and / or relative to executable expected location)
+ [ ] Possibly pack the include headers into the  executable? (this seems daft)
+ [ ] set FLAMEGPU2_ROOT instead, and infer `/include` from that.

This needs to be
+ [ ] Clearly documented

Robadob:
> Possibly pack the include headers into the executable? (this seems daft)

It's certainly unusual, however current headers dir is 503kb, size of the smallest debug executable on windows is 20mb.

Bearing in mind, we plan to reduce the size of header dir further in future, it's not the most insane idea.

It's also the only portable method (assuming no weird machines without write access).

ptheywood:
The recent patch searches relative to the current working directory, rather than the current binary's directory, so calling from outside can lead to the path not being found. 
Both would probably be a wise choice.

Robadob:
That's a good shout, I didn't consider it tbh. I'd just need to check the cross platform way of getting binary location. Trivial addition.

ptheywood:
Some of this should also be conisdered pre binary release #514

Robadob:
Worth noting, the Python wheel now includes the debug headers. Though C RTC binaries don't.

ptheywood:
support `FLAMEGPU2_INC_DIR` (deprecated but works, lwoer priotiy) and `FLAMEGPU_INC_DIR`

Robadob:
#1074 Adds GLM headers to Python wheels (as this seems compliant with the license as that's packaged too) and setups up `GLM_INC_DIR` environment variable to override path.

PR #387, opened 2020-10-14
RTC include directory detection.

Robadob:
If `FLAMEGPU2_INC_DIR` env var is not set, it will check the working directory and 4 layers of parent directory for the include directory.

Currently doesn't include a mechanism to check version, will attempt to add that in next commit.

#253 (Not sure at this stage whether it is merely relevant to or closes the issue).


ptheywood (approved):
Verified that it works as expected on linux. 
+ No env var finds the correct value `export FLAMEGPU2_INC_DIR=`
+ The correct path in `FLAMEGPU2_INC_DIR` works
+ A bad path in `FLAMEGPU2_INC_DIR` errors with an appropriate error message saying to set the variable correctly.

PR #391, opened 2020-10-14
Moves further device side error checks behind NO_SEATBELTS.

Robadob:
These (getVariable() for each message type) were all missing for some reason.


Robadob:
Tests have passed with and without seatbelts. It's a fairly trivial change so I will merge after CI.

PR #404, opened 2020-10-21
Use __ldg for Spatial3D Messages (non RTC)

ptheywood:
This appears to improve performance of boids 32k on a Titan V GPU by a small amount (~1.5%)

Should check on Turing and possibly at larger size before merging / closing.

spatial-ldg


Robadob (approved):
Probably want to re-base and delete the earlier commit, because it will get confused after that branch is squashed.

I'm happy for it to be merged, but might want to run it by Paul 🤷‍♂️

ptheywood:
Closing for now, as needs more thought/evidence on whether or not this should be done. See #407

PR #405, opened 2020-10-21
Invert the NO_SEATBELTS cmake option for usability.

ptheywood:
`-DSEATBELTS=OFF` makes more sense than `-DNO_SEATBELTS=ON`.

Does not change compile time macro.


Robadob (approved):
Looks good to me, don't think Paul needs to review. Can merge after CI.

PR #408, opened 2020-10-22
Add NO_SEATBELTS checks to device agent birth/death.

### LINKED ISSUE #382: Device Exceptions for Birth/Death if not enabled.
Agent Birth and Death from teh device are opt in (`setAgentOutput`, `setAllowAgentDeath`).

If these features are not opted in to, a (device) exception should be reaised if they are detected. 

This should only be checked for if seatbelts are enabled.

Robadob:
Tweaked death slightly, so it now checks before setting death flag, previously it would always set death flag regardless. This was a touch easier than passing an extra value to kernel (tad more expensive if doing death, tad cheaper if not doing death?). Happy to do the inverse if preferred.

Closes #382

### DIFF
--- a/include/flamegpu/runtime/AgentFunction.h
+++ b/include/flamegpu/runtime/AgentFunction.h
@@ -87,10 +87,14 @@ __global__ void agent_function_wrapper(
         MsgOut::Out(agent_func_name_hash, messagename_outp_hash, out_messagelist_metadata, scanFlag_messageOutput));
 
     // call the user specified device function
-    {
-        FLAME_GPU_AGENT_STATUS flag = AgentFunction()(&api);
+    FLAME_GPU_AGENT_STATUS flag = AgentFunction()(&api);
+    if (scanFlag_agentDeath) {
         // (scan flags will not be processed unless agent death has been requested in model definition)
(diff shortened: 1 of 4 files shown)


ptheywood:
Given that there are typically more kernels without death than with death, having death be the more expensive option probably makes sense. 

Adding the nullptr check (branch) to each kernel launch might have an impact though?, although it should be uniform for all agents, and is only once per thread per agent function unlike the valueptr branching in curve we removed.

Robadob:
It's either a branch or a write to global

ptheywood:
True, ran boids on Titan V compared to `master` and no significant change in total runtime, within usual noise varaition. 

Do we want to add a SEATBELTS only test which detects this device exception is thrown when expected?

Robadob:
> Do we want to add a SEATBELTS only test which detects this device exception is thrown when expected?

Easily done, there's already a test suite full of these tests. I'll get to it when i next reach an impass on my env refactor im doing atm.

Robadob:
ha and it's wrong. Should have tested it. I'm not evaluating the return value. Will error everytime death is disabled currently.

Robadob:
Also only implemented agent out check on setting arrays vars...

All fixed now, tested tests with tests_dev.

ptheywood (approved):
Can confirm new tests pass under linux. Merging.

Issue #330, opened 2020-06-30
Improved Testing for RNG


ptheywood:
@Robadob has potentially found (and fixed?) an issue with the range of random values produced by `AgentRandom::uniform`.

The test case for this is a little weak (i.e. it doesn't test for uniformity, or the bounds. 

Tests need strengthenign to assert that the generated distribution of numbers is aproximately uniform within the expected range, and never generates out of range values.

0. Seed with a sane seed (or do this for multiple seeds)
1. Generate lots and lots of random values
2. Ensure all values are in range (max/min reductions could be used)
    + Ideally also check min/max expected value is produce, although this will potentially require an incredibly large sample size for floating point values, or large-range integers. Limiting the range of values to be relatively small should (i.e. will be much more likely) to produce values of min and max.
3. Generata a histogram, and assert that the values are (approximately) uniform.
    + Simple for integer types, float types will be a bit more complicated.
4. Ensure that all values

Robadob:
Branch: `int_rng_fix`

Robadob:
I wrote a standalone test for the uniform float implementation now (after #411 ) present in `master`.

The code can be found here: https://gist.github.com/Robadob/5e3a068267f8a351991bef16d138dd7c

To my surprise it executes instantly for me locally, producing the below output.

```
Found 256 zeros
Found 0 ones
Min Bin Size 4194304
Max Bin Size 4194304
```

Hence, when mapping every generated float value to one of 1024 bins, the bins are uniformly filled.

PR #411, opened 2020-10-23
Fixes AgentRandom uniform int

Robadob:
Tested 600 ints (from the same thread) in the range [0, 5]

Got this histogram, which seems as you would expect.
```
0: 118
1: 86
2: 89
3: 109
4: 90
5: 110
```

curand rng returns in the range `(0, 1]`, so the math is a bit arkwards.
Furthermore, as we return an inclusive range, we must add an additional 1 to the range (`max-min`).
Perhaps there's a slight argument for the upper bound to be exclusive to simplify the maths.

### DIFF
--- a/include/flamegpu/runtime/random/AgentRandom.cuh
+++ b/include/flamegpu/runtime/random/AgentRandom.cuh
@@ -1,7 +1,7 @@
 #ifndef INCLUDE_FLAMEGPU_RUNTIME_RANDOM_AGENTRANDOM_CUH_
 #define INCLUDE_FLAMEGPU_RUNTIME_RANDOM_AGENTRANDOM_CUH_
 
-#include <cassert>
+#include <limits>
 
 #include "flamegpu/detail/curand.cuh"
 #include "flamegpu/detail/StaticAssert.h"
@@ -24,7 +24,7 @@ class AgentRandom {
     __forceinline__ __device__ AgentRandom(detail::curandState *d_rng);
     /**
      * Returns a float uniformly distributed between 0.0 and 1.0. 
-     * @note It may return from 0.0 to 1.0, where 1.0 is included and 0.0 is excluded.
+     * @note It may return from 0.0 to 1.0, where 0.0 is included and 1.0 is excluded.
      * @note Available as float or double
      */
     template<typename T>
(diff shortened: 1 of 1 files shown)


ptheywood:
Probably worth aiming to be consistent with what users would expect in terms of inclusivity. 

`std::uniform_int_distribution` is `a <= i  <= b`. 
Pythons `random.randint(a, b)` is `a <= N <= b` (although `randrange(start, stop)` also exists as `start <= N < stop`

In terms of floating point:
+ std::uniform_real_distribution` is `a <= x  < b` (which is apparently the "rectangular distribution" of `[a,b)`
+ Python `random.random()` is  `[0.0, 1.0)`

So curand seems to differ from what standard libraries suggest. Would it not be better to modify `AgentRandom::uniform()` so that it returns `[0.0, 1.0)` instead of curands `(0.0, 1.0]`, then the other methods would be cleaner and curand would be more like the rest of the world? Might be a key point when porting from flamegpu 1 though.

Robadob:
Trivial change, I just err on the side of 'that's more ops'.

ptheywood:
I would assume that the cost of an extra floating point subtraction is going to be negligible compared to the cost of actually generating a random number.

Could also expand the testing of RNG (#330) while i'm blocking changes to CUDASimulation

Robadob:
I'm not really sure how #330 should look. I certainly don't think it will fit inside the regular test suite.
Random is random, one trusts it does what it's documented to do.

ptheywood:
No, but we are documenting the inclusivity of the ranges for float and int variants, so atleast verifying that would be good. 
The random integer generation is also done by us (from the curand float) so it's probably worth making sure that distribution is approximately correct (as a common mistake with doing that (in C atleast) is to have too few at the upper bound due to rounding)

Robadob:
The int generation works now, small histogram suggests its got no huge bias, and I can't see it being changed further (besides lopping off the inclusive upper bound), so that doesn't seem like an issue.

ptheywood:
Int generation seems to be inclusive most places so i'd keep it inclusive.

Ranges should certainly be tested though (as changing what uniform returns would potentailyl lead to out of bounds etc) and worth having a test.

Robadob:
I want to make some more changes to this before it's merged (mostly to how double uniform is calculated)

Robadob:
Actually can't do what I thought I could, so won't bother.

Robadob:
Windows/Debug/SEATBELTS=ON Python tests pass (I wanted to check RTC would build the updated header)

```

===================================== 648 passed, 11 skipped in 331.66s (0:05:31) =====================================

```

ptheywood (approved):
Changes look OK at a glance + your testing mentioned on slack sounds OK. Worth getting merged in during 2.0.0-rcX given its potentially a significant change for users.

Still needs more thorough testing eventually so we'll keep #330 open.

PR #412, opened 2020-10-23
Improve doc comments for host functions/conditions.

### LINKED ISSUE #268: Exit Condition Documentation
Exit conditions effectivley run between iterations. I.e. after iteration 0 but before iteration 1 / at the end of iteration 0.

The interaction of this with getStepCounter() should be clarified / well documented.

### LINKED ISSUE #270: Order of exitCondition functions.
Exit conditions are not executed in the order they are added to the model. 

This is unintuivite (although not sure why anyone would actually have two which are order dependent..., i'm abusing it for testing purposes)

This should be **well** documented **or** changed to preserve order of adding to the model.

i.e. 

```
FLAMEGPU_EXIT_CONDITION(exitCondition_a) {
    printf("a\n");
    return CONTINUE;
}
FLAMEGPU_EXIT_CONDITION(exitCondition_b) {
    printf("b\n");
    return CONTINUE;
}
// ...
model.addExitCondition(exitCondition_b);
model.addExitCondition(exitCondition_a);
// ...
```

will produce 

```
a
b
```

I.e. it appears to be order of definition, not order of adding to the model.

**Edit** it does not appear to always be order of deinition either:

```
FLAMEGPU_EXIT_CONDITION(exitCondition_testGetStepCounter) {
printf("a\n");
return CONTINUE;
}
FLAMEGPU_EXIT_CONDITION(exitCondition_increase_expectStepCounter) {
printf("b\n");
return CONTINUE
}
// ...
model.addExitCondition(exitCondition_testGetStepCounter);
model.addExitCondition(exitCondition_increase_expectStepCounter);
// ...
```

```
b
a
```

**Edit**: Currently, regular exit conditions are called prior to any python(maybe RTC?) exit conditions, which would also warrent documentation.

Robadob:
Also put SWIG hostfn callbacks behind macros (to reduce confusion for those who read headers, hence why this depends on alternate_array_interface branch).

Closes #268 
Closes #270

Any further detail will need to go in the userguide.

### DIFF
--- a/include/flamegpu/model/LayerDescription.h
+++ b/include/flamegpu/model/LayerDescription.h
@@ -105,7 +105,7 @@ class LayerDescription {
      * @param func_p Function pointer to the host function declared using FLAMEGPU_HOST_FUNCTION notation
      * @throw InvalidHostFunc If the function has already been added to the layer
      * @throw InvalidLayerMember If the layer already contains a SubModel
-     * @note This version exists because the template overload was preventing implicit cast to std::string
+     * @note There is no guarantee on the order in which multiple host functions in the same layer will be executed
      */
     void addHostFunction(FLAMEGPU_HOST_FUNCTION_POINTER func_p);
     /**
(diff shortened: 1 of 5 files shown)


Robadob:
Currently failing because the ifdef'd away functions aren't being built into the lib.

Need to either move them into the header, or turn on SWIG macro when building lib and pyflamegpu is enabled.

PR #413, opened 2020-10-26
Update Thrust and Jitify Dependencies

ptheywood:
Also adds a note about the potential to use FindThrust from a future CUDA version, to prevent compilation errors with future thrust version mismatches.

Tests all pass under linux.


Robadob:
The windows CI failure is that thrust header one 🙄

Issue #329, opened 2020-06-29
On-device initialisaiton functions (or step/exit functions?)


ptheywood:
Current init functions are sequential host methods. 

In some cases, doing this on the device is probably worthwhile (although it adds complexity). 

General idea is:

+ `intiLayers` as a set of layers which contain host or device initialisation functions (or submodels)
+ the majority of the body of `step` becomes a private method which takes an ordered list of layers and processes them
    + `step` then calls this method for init, agent, step and exit layer sets etc.
+ The method to add a host init (/step/exit) function, either adds to an exisitng layer, or implicitly creates and appends a new layer (if a layer is not provided)
+ Restrictions will have to be imposed on which functions can be in the same layer(s) for the sake of concurrency, as with restrictiosn on layers.

Longer term the layers would become a DAG as with the main layers.

Would need to make sure curand behaves as intended, and potentially provide simple methods for provided non-uniform distributions across the populations (i.e normal distributions) otherwise a sequential host method would still have to be used instead.

Robadob:
Simulate then calls this method...
And not for step functions. That would create unwanted recursion

Robadob:
> a public version of `step` is still required though which doesn't take a set of layers though surely? I.e. where you call step explicilty in the test suite?

Yes, `step()` currently never calls init or exit fns, hence it should probably read: "`simulate` then calls this method for init, agent, step and exit layer sets etc."


> I don't see how it would be recursion, Step functions are just methods executed after the agent function layers, so no reason why they can't also have their own set of layers (though arguably there shoulnt be agent functions in there, but certainly scope for device step functions in some very obscure cases)?

Yeah, I was mistaken on this point whilst reading and walking. I was thinking a vector of layers, representing a step would have step functions.

> Argubaly we don't actually need explicit step functions anymore, as we have host-layer functions, so step functions are just host layer functions which are implicitly at the end of the iteration.

I agree, but they exist as backwards compatibility.

Robadob:
This alludes to device-agent initialisation functions (and they don't have a separate issue).

The easiest way to handle this would be with the existing Device API and agent function wrapper used by regular agent functions. But that would leave unusable message in/out properties (they would be treated as MsgNone) and redundant? read access to agent variables. The alternative being to make a bespoke InitDeviceApi, which then leads to either duplicated code or double inheritance/similar.

Robadob:
[...]

Issue #245, opened 2020-04-30
Ensemble of multiple concurrent (or sequential) simualtions


ptheywood:
ABMs typically need to be run many times as an ensemble to get useful results.

Baking this into FLAME GPU 2 would enable small models on the GPU to be worthwhile, by providing concurrency across the simulations (and within each simulation).

The following is a rough outline of what `main` would look like for an ensemble of simulations, factoring in:

+ The ability to run 1 simulation, or an ensemble of N simulations.
+ Different (environment) parameters per ensemble
    + Model specifies default values
    + Parmeter set(s) specify overriding values, per simulation (very draft)
+ Implements a `Simulator`, individual `Simulation`s grouped as an `Ensemble`
    + Some form of aggregate operations over the results of an `ensemble` (very draft)
+ Populations are produce by the `simulator`, and may be re-used, or copied for differnt simulations, or different population can be use. 
    + Unique ID generation would be moved to the `simulator`, so they are either unique for an agent type across all `simulations`, or can be limited to the simulation.
    + Populations would also have an index alongside the id, which is relative to that population.

Again, this is again very draft / early thoughts.

```cuda


int main(int argc, char*argv[]){

    // Specify the model as usual 
    FLAMEGPU::model::ModelDescription model("model");
    // ... 

    // Create a simulator object, fixing the model
    FLAMEGPU::simulation::CUDASimulator simulator(model);
    // Initialise the simulator with cli. 
    simulator.initialise(argc, argv);

    // Generate a population of agents
    const unsigned int INITAL_POP = 8;
    FLAMEGPU::population::AgentPopulation population = simulator.newPopulation(model.agent("agent"), INITAL_POP);

    for (FLAMEGPU::population::AgentInstance &agentInstance : population) {
        // Set values, can use the index within the population, or the global id from the simulation.
        circleInstance.setVariable<float>("x", agentInstance.index() % 2;
        circleInstance.setVariable<float>("y", agentInstance.id() % 2;
    }


    {
        // Can run a single simulation
        FLAMEGPU::simulation::Simulation simulaton = simulator.simulation();
        simulation.setPopulation(model.agent("agent"), population);
        // Override environment variables 
        simulation.environment.setVariable<int>("seed", 12);
        // run the simulation 
        simulation.simulate();
    }


    {
        // Or could run an ensemble of simulations, including a way of sampling parameters 

        // Get the ensemble
        const unsigned int ENSEMBLE_SIZE = 4;
[...]

PR #415, opened 2020-10-29
Ensemble 1 notes.

Robadob:
This is mostly a place to make notes/thoughts whilst I fiddle with ensembles. Specifically, looking at the class '1', type where multiple instances of the same model run in lock-step, sharing kernel launches (but not variable arrays).

**Concerns**
* We have spoken about memory limitations. How will set/get pop work if memory is limited (e.g. if we want to run 100 ensembles, how can we pre-set their populations if it will outstrip even host memory). Presumably we need towards flat file/init functions being first class citizens (#329). This is less of an issue, if we intend to run multiple copies from the same init, with differing random seeds (similar to primage validation).
* Runs should have different random seeds, this however makes management of random state awkward, so isn't currently a priority for current tests.
* How do methods like get/reset step count work, surely these should be per model instance, rather than whole ensemble, Although i lean to not having them at all and only having `simulate()`.
* How will input files work, 1 per instance, 1 input file + different random seeds?
* If loading config from input file, we will need to have some changes to how loader works.
* Need to move the `Simulation` instance id mechanic out of `Simulation` if we are abandoning shared inheritance. It is required that these Ids are unique per model instance for environment manager. Currently using 0+ as instance ID rather than proper calculating anything (This almost certainly won't work for submodels).
* Host API object requires access to `CUDASimulation`. It uses this for getting model description hierarchy, instance id, and access to CUDAAgent objects.
* Need to pass a flag to RTC to compile the other ensemble wrapper
* `InstanceOffsetData`, `InstanceOffsetData_cdn` need some work, currently consist of 4 memcyps per agent function launch, and use of SoA, means we don't need two separate device buffers as we are currently using.

Relates to #245

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,6 +52,7 @@ option(BUILD_EXAMPLE_CIRCLES_BRUTEFORCE "Enable building examples/circles_brutef
 option(BUILD_EXAMPLE_CIRCLES_SPATIAL3D "Enable building examples/circles_spatial3D" OFF)
 option(BUILD_EXAMPLE_GAME_OF_LIFE "Enable building examples/game_of_life" OFF)
 option(BUILD_EXAMPLE_HOST_FUNCTIONS "Enable building examples/host_functions" OFF)
+option(BUILD_EXAMPLE_ENSEMBLE "Enable building examples/ensemble" OFF)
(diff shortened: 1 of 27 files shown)


Robadob:
Closing this to cleanup the PR page. Might be useful to recover the code if we ever decide to also pursue this route.

Issue #356, opened 2020-07-13
Curve: Collision handling


Robadob:
Currently device side curve accesses only confirm that type size matches. This is clearly flawed.

Temporary solution, also confirm the unmodulated hash matches. This will be a significant improvement, at the cost of an extra memory read.

The long term solutions to this may only be available via RTC, due to the pre-processing which can be performed.

PR #418, opened 2020-11-06
Multi thread/device support

### LINKED ISSUE #395: TestCUDASimulation.AllDeviceIdValues causes test failures on multi-gpu systems
Tests haven't been run recently on mavericks. When running with `NO_SEATBELTS` in either configuration 26 tests fail. This is unusual given all test pass for both @ptheywood and I on our local machines. This was discovered on PR #371, however testing has shown the same to apply to master branch.

Seemingly in common the tests all make use of environment properties, however I haven't dug into it further than that.

Full list of failing tests:
```
[  FAILED  ] 26 tests, listed below:
[  FAILED  ] AgentEnvironmentTest.Get_float
[  FAILED  ] AgentEnvironmentTest.Get_double
[  FAILED  ] AgentEnvironmentTest.Get_int8_t
[  FAILED  ] AgentEnvironmentTest.Get_uint8_t
[  FAILED  ] AgentEnvironmentTest.Get_int16_t
[  FAILED  ] AgentEnvironmentTest.Get_uint16_t
[  FAILED  ] AgentEnvironmentTest.Get_int32_t
[  FAILED  ] AgentEnvironmentTest.Get_uint32_t
[  FAILED  ] AgentEnvironmentTest.Get_int64_t
[  FAILED  ] AgentEnvironmentTest.Get_uint64_t
[  FAILED  ] AgentEnvironmentTest.Get_arrayElement_float
[  FAILED  ] AgentEnvironmentTest.Get_arrayElement_double
[  FAILED  ] AgentEnvironmentTest.Get_arrayElement_int8_t
[  FAILED  ] AgentEnvironmentTest.Get_arrayElement_uint8_t
[  FAILED  ] AgentEnvironmentTest.Get_arrayElement_int16_t
[  FAILED  ] AgentEnvironmentTest.Get_arrayElement_uint16_t
[  FAILED  ] AgentEnvironmentTest.Get_arrayElement_uint32_t
[  FAILED  ] AgentEnvironmentTest.Get_arrayElement_int32_t
[  FAILED  ] AgentEnvironmentTest.Get_arrayElement_uint64_t
[  FAILED  ] AgentEnvironmentTest.Get_arrayElement_int64_t
[  FAILED  ] DeviceAPITest.getStepCounter
[  FAILED  ] DeviceAPITest.getStepCounterFunctionCondition
[  FAILED  ] SubEnvironmentManagerTest.SubDeviceAPIGetDefault
[  FAILED  ] SubEnvironmentManagerTest.SubDeviceAPIGetMasterChange
[  FAILED  ] SubEnvironmentManagerTest.SubSubDeviceAPIGetDefault
[  FAILED  ] SubEnvironmentManagerTest.SubSubDeviceAPIGetMasterChange

```

Robadob:
**ToDo**
- [x] Make `EnvironmentManager` thread-safe
- [x] Make `Curve` thread-safe
- [x] Encode `CudaSimulation` instance id, into Curve hashes.
- [x] Make `EnvironmentManager` multi-device aware
- [x] Make `Curve` multi-device aware
- [x] Make `CUDASimulation` multi-device aware
    * *It contains a runtime check which purges device data if it is detected corrupted, changing device currently breaks this.*
- [x] Add nvcc arg `--default-stream per-thread` to CMake
     * https://developer.nvidia.com/blog/gpu-pro-tip-cuda-7-streams-simplify-concurrency/
* Fix RTC
    - [x] Does the dynamic header need changing (or will it for ensembles)
    * ~I think Pete said jitify complains about compiling in anything other than the main thread,~ *(RTC compile works, doesn't print any warning it's just noticeably slower)*
* Write a bunch of test cases to ensure model integrity isn't harmed during multi device/thread `CUDASimulation` execution.
    * multi-threaded
        - [x] Agent set/getVariable
        - [x] MessageIn getVariable
        - [x] MessageOut setVariable
        - [x] AgentOut setVariable
        - [x] Environment getProperty
        - [x] AgentFunctionCondition Agent getProperty
    * multi-device
        - [x] Agent set/getVariable
        - [x] MessageIn getVariable
        - [x] MessageOut setVariable
        - [x] AgentOut setVariable
        - [x] Environment getProperty
        - [x] AgentFunctionCondition Agent  getProperty
    * RTC multi-device
        - [x] Agent set/getVariable
        - [x] MessageIn getVariable
        - [x] MessageOut setVariable
        - [x] AgentOut setVariable
        - [x] Environment getProperty
        - [x] AgentFunctionCondition Agent  getProperty
- [x] Regular tests work
- [x] RTC tests work
- [x] Reset `tests_dev` cmake

**ChangeLog**
* Make `Curve` singleton thread-safe and unique per-device
* Make `EnvironmentManager` singleton thread-safe and unique per-device
* Add mutex locks around agent function execution, to prevent environment data getting changed by defragment during before execution).
* Add nvcc arg `--default-stream per-thread` to CMake  (and then comment out as it would require far more testing)
* Fix double inclusion of `common.cmake` if configuring with an example as root.
* Modify initialisation of `CUDASimulation` so that `EnvironmentManager` is not used before device has been selected.
* Add tests for multi-threading and multi-device `CUDASimulation` execution.
* Adds a reduced set of multi-thread and multi-device tests for RTC.
* Fix a bug, where rtc_offsets were reset to 0 when `EnvironmentManager:defragment()` was called.

**Known Issues**
*  Curve collision handling (#356) is often triggered if running new tests for multi-device mode. The state of this is dependent on the order of test execution and the number of devices within the machine. This is due to the `CUDASimulation` instance id incrementing each time, and this mutating the hashes used. The test does not collide in isolation, but atleast 1 test from the series always seems to collide if ran as a set. Might be trouble with hashes bunching, as there should only be a max of around 24/1024 hashes in use at any time during these tests.
* `nsys` timeline for one of the multi-device test cases, appears to show the context creation for tertiary devices waiting for CUDA to become idle on already initialised devices. Seems important to initialise CUDA on all devices before launching sims. Calling `cudaFree(nullptr)` on every device at the start of the test appears to work,
* The new compile arg `--default-stream per-thread` removes some implicit device syncs that would otherwise be expected. Very important to sync on either side of agent functions.

Relevant to #245 
Closes #395 (if done properly)

### DIFF
--- a/cmake/common.cmake
+++ b/cmake/common.cmake
@@ -1,4 +1,5 @@
 message(STATUS "-----Configuring Project: ${PROJECT_NAME}-----")
+include_guard(DIRECTORY)
 if(NOT CMAKE_VERSION VERSION_LESS 3.18)
     cmake_policy(SET CMP0105 NEW) # Use separate device link options
 endif()
@@ -152,7 +153,12 @@ set(CMAKE_CUDA_FLAGS_RELEASE "${CMAKE_CUDA_FLAGS_RELEASE} -lineinfo")
 set(CMAKE_CUDA_FLAGS_PROFILE "${CMAKE_CUDA_FLAGS_PROFILE} -lineinfo -DPROFILE -D_PROFILE")
 # Addresses a cub::histogram warning
 set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} --expt-relaxed-constexpr")
-
+# Enable default stream per thread for target, in-case of ensembles
+# This removes implicit syncs in default stream, using it has only been tested for basic models
+# It has not been tested with host functions, agent death, optional messages etc
+# Hence using it is unsafe
+#set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} --default-stream per-thread")    
+    
 # Host Compiler version specific high warnings
 if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
     # Only set W4 for MSVC, WAll is more like Wall, Wextra and Wpedantic
--- a/include/flamegpu/gpu/CUDAAgent.h
+++ b/include/flamegpu/gpu/CUDAAgent.h
@@ -64,16 +64,18 @@ class CUDAAgent : public AgentInterface {
      * Uses the cuRVE runtime to map the variables used by the agent function to the cuRVE
      * library so that can be accessed by name within a n agent function
      * @param func The function.
+     * @param instance_id The CUDASimulation instance_id of the parent instance. This is added to the hash, to differentiate instances
      * @note TODO: This could be improved by iterating the variable list within the state_list, rather than individually looking up vars (the two lists should have all the same vars)
      * @note This should probably be addressed when curve is updated to not use individual memcpys
      */
-    void mapRuntimeVariables(const AgentFunctionData& func) const;
+    void mapRuntimeVariables(const AgentFunctionData& func, const unsigned int &instance_id) const;
     /**
      * Uses the cuRVE runtime to unmap the variables used by the agent function to the cuRVE
      * library so that they are unavailable to be accessed by name within an agent function.
      * @param func The function.
+     * @param instance_id The CUDASimulation instance_id of the parent instance. This is added to the hash, to differentiate instances
      */
-    void unmapRuntimeVariables(const AgentFunctionData& func) const;
+    void unmapRuntimeVariables(const AgentFunctionData& func, const unsigned int &instance_id) const;
     /**
      * Copies population data from the provided host object
      * To the device buffers held by this object (overwriting any existing agent data)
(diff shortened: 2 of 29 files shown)


Robadob:
Current issue is IO tests failing because the Singletons have already been init, when loading frm file. As such the setup env vars loaded from file code is not executed.

Robadob:
Full test suite now passes on Linux.

Will add the multi-thread, multi-device tests tomorrow. Then test RTC.

Robadob:
One side-effect/by-product of multi-device is an issue we have when a thread exits in not device 0, and a new CUDASim is then created in the same thread, without having applyConfig() called, to change device back to device 0 (the default value found in config).

Robadob:
Didn't make much progress today.

* Tweak environment property hashing, this seems to have fixed collisions I was having.
* Modified `CUDASimulation` destructor to do everything on the right device/properly.
* Some other minor changes
    * Moved DeviceExceptionBuffer shared memory copy, at start of agentfunction/agentfunctioncondition, so it's only performed by 1st thread of block.
    * etc.

Spent most the day chasing DeviceException test failures. Had some confusion with Mav, which had different errors. The failures still (only?) occur on my laptop and I'm yet to find a reasonable excuse for why they dissappear as soon as I make a change. Should really debug them with tests_dev to have a much faster build time.

Robadob:
A bunch of message tests are failing/crashing out in release model (on both Windows/Linux), need to investigate.

Robadob:
Initial status of RTC tests, first test passed (message), other 3 failed. The first does not use any environment variables, so this is likely the issue.
Of the test setup with the most detailed failure output, curiously only 1 of 10 tests per device passed. This might hint towards the problem.
Each test also takes several minutes to run, dependent on the number of threads. (and more threads == more chance of hitting a test failure).

PR #421, opened 2020-11-18
Update minimum CUDA version to 10.0

### LINKED ISSUE #406: CUDA 9.X support.
Currenlty we vaguely claim to support `CUDA >= 9.0`. 

In `common.cmake`, we enforce `CUDA > = 7.0`.

While in `CMakeLists.txt` we enforce `CUDA >= 9.0`.

In fact I don't belive it's possible to build with some versions of 9 on some OS's at this point in time (hence CI doesn't try). 
In practice, now that 11.0 and 11.1 are out, it seems sensible that we support the last 2 major releases (at the approx time of flamegpu2 being released) which maps to 10.0+. 

This won't change any support in terms of hardware (Kepler can still be used on 10). 

I think i was the main block to this being the case originally, as at the time 10.x wasn't deployed on various HPC systems. This is still true for 11.x at the time of writing. 

+ [x] Update readme
+ [x] Switch CI to run min/max  (10.0 + 11.1?)
+ [x] Consistent CUDA version checking in common.cmake and CMakeLists.txt
+ [x] Remove mentions / specialisations for older CUDA.

+ [x] If not yet done, upgrade windows CI to 11.1 update 1.

I'm happy to do this if agreed upon.

ptheywood:
Updates minimum supported CUDA version to 10.0, removing mentions of 9.x issues.
Removes Windows CUDA 10.x CI job due to intermittent compiler failures fixed by CUDA 11.0
Simplifies CI workflows with fewer disabled CUDA combinations
Updates Readme with agreed dependency versions
Applies some markdown tweaks to the readme
Adds CUDA 11.1.1 as a download target for windows CI.

Closes #406

### DIFF
--- a/.github/workflows/Lint.yml
+++ b/.github/workflows/Lint.yml
@@ -13,7 +13,7 @@ jobs:
       matrix:
         include:
           - os: ubuntu-18.04
-            cuda: "10.2"
+            cuda: "11.1"
     env:
       build_dir: "build"
       build_tests: "ON"
--- a/.github/workflows/Ubuntu.yml
+++ b/.github/workflows/Ubuntu.yml
@@ -21,31 +21,18 @@ jobs:
           # - os: ubuntu-18.04
           #   cuda: "11.0"
           #   gcc: 9
-          - os: ubuntu-18.04
-            cuda: "10.2"
-            gcc: 8
+          # - os: ubuntu-18.04
+          #   cuda: "10.2"
+          #   gcc: 8
           # - os: ubuntu-18.04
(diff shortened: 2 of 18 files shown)


ptheywood:
`*` is used for loads of things in markdown, I accidentally made stuff italics/bold at one point so switched to `+` to avoid that (I missed a space while trying to add emphasis that i've since removed as it wasn't helpful)

Just tried cmake 3.12, which no longer works due to Thrust's Cmake requiring 3.15+, so will bump it up to that for now as the strictly enforced minimum version.

PR #422, opened 2020-11-19
Ensemble 3

Robadob:
`CUDAEnsemble` (#245) and associated classes/tests implementation based on the [agreed interface example](https://gist.github.com/Robadob/076ce990e560ba9042009ece65a6af99). This will make use of the multi-thread/device support introduced by #418  (Ensemble 1 strategy (#415) was put on hold due to it being a much larger undertaking).

**ToDo**
- [x] Implement initial `RunPlan` classes
- [x] Implement initial `CUDAEnsemble` class
- [ ] Add methods for import/export `RunPlanVec` to disk, and integrate with `CUDAEnsemble` command line args
- [x] Decide how results will be extracted from `CUDAEnsemble` (can't really test properly without this)
- [x] Test with C
- [x] Test with Python
- [ ] Add the memory space estimating code based on max agent populations

**Changelog**
* Added `CUDAEnsemble` class and `Ensemble` example demonstrating it's usage.
* Added `RunPlan` and `RunPlanVec` classes for generating a collection of configurations to execute with the ensemble
* Added swig support for new classes
* Fixed a deadlock caused by recursive shared locks of a mutex (whilst a 2nd thread asked for unique lock in the middle). This was likely unique to windows, recursive locking is undefined behaviour.
* Added `HostRandom::setSeed()`, `HostRandom::getSeed()`
* Added max steps property to `SubModelData`, this allows a submodel to have 0 exit conditions.

**Known Issues/Concerns**
* Still undecided how default steps and random seed will get to `RunPlanVec`.
* Swig gives warnings related to instances of `RunPlanVec` using `std::vector<RunPlan>` e.g. `operator[]` (Stack overflow suggests this might not prevent the methods being wrapped)
* Undecided on whether the LoggingConfigs should be owned by the `CUDASimulation`/`CUDAEnsemble`, or defined separate like `RunPlanVec`.
* The calculation of standard deviation of agent variable log currently uses a `__constant__`, it has a mutex to ensure it's thread safe. But this is shared between all devices/threads. Could potentially cause slowdown in ensemble scenarios with alot of step logging standard deviation.
* Logging IO classes don't share much with the existing IO classes (I've hated those for a while).
* Ensemble exit logs have no attachment to run index, they just exist in a single file in order of completion.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,6 +51,7 @@ option(BUILD_EXAMPLE_CIRCLES_BRUTEFORCE "Enable building examples/circles_brutef
 option(BUILD_EXAMPLE_CIRCLES_SPATIAL3D "Enable building examples/circles_spatial3D" OFF)
 option(BUILD_EXAMPLE_GAME_OF_LIFE "Enable building examples/game_of_life" OFF)
 option(BUILD_EXAMPLE_HOST_FUNCTIONS "Enable building examples/host_functions" OFF)
+option(BUILD_EXAMPLE_ENSEMBLE "Enable building examples/ensemble" OFF)
 option(BUILD_SWIG_PYTHON "Enable python bindings via SWIG" OFF)
 cmake_dependent_option(BUILD_SWIG_PYTHON_VIRTUALENV "Enable building of SWIG Python Virtual Env for Python Testing" OFF
                        "BUILD_SWIG_PYTHON" OFF)
@@ -105,10 +106,6 @@ endif()
 
 # Add each example
 
-if(BUILD_ALL_EXAMPLES OR BUILD_EXAMPLE_HOST_FUNCTIONS)
-    add_subdirectory(examples/host_functions)
-endif()
-
 if(BUILD_ALL_EXAMPLES OR BUILD_EXAMPLE_BOIDS_BRUTEFORCE)
     add_subdirectory(examples/boids_bruteforce)
 endif()
@@ -136,6 +133,15 @@ endif()
 if(BUILD_ALL_EXAMPLES OR BUILD_EXAMPLE_GAME_OF_LIFE)
     add_subdirectory(examples/game_of_life)
 endif()
+
+if(BUILD_ALL_EXAMPLES OR BUILD_EXAMPLE_HOST_FUNCTIONS)
+    add_subdirectory(examples/host_functions)
+endif()
+
+if(BUILD_ALL_EXAMPLES OR BUILD_EXAMPLE_ENSEMBLE)
+    add_subdirectory(examples/ensemble)
+endif()
+
 # Add the tests directory (if required)
 if(BUILD_TESTS)
     add_subdirectory(tests)
--- a/examples/ensemble/CMakeLists.txt
+++ b/examples/ensemble/CMakeLists.txt
@@ -1,0 +1,36 @@
+# Set the minimum cmake version to that which supports cuda natively.
+# 3.10 required for cuda -std=c++14, however 3.12 fixes some device linker errors
+cmake_minimum_required(VERSION VERSION 3.12 FATAL_ERROR)
+
+# Name the project and set languages
+project(ensemble CUDA CXX)
+
+# Set the location of the ROOT flame gpu project relative to this CMakeList.txt
+get_filename_component(FLAMEGPU_ROOT ${CMAKE_CURRENT_SOURCE_DIR}/../.. REALPATH)
+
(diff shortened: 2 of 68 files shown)


Robadob:
Notes from today's meeting.

CUDAEnsemble.config.json_outdir = Directory for ensemble output
Runplan.out_subdir = Subdirectory per RunPlan for output 
Output files with run index filename
Log files at exit time
HostRandom::getSeed()
HostRandom::setSeet()
Log errors to error log files
Step logs go into per run file.
Exit logs go into per simulate()
Add steps to SubModel description, if you want submodel to have fixed/max steps
RunPlanVec::operator*=() ? Clone a vector N times?
How does adding a vector to itself work?

BryanLaura (approved):

Robadob:
Link multiply defined warning about me including the below in a `AgentLoggingConfig.h`, need to address this. Manual expansion to keep the protoype but move the implementation outside is probably the solution, as it's heavily templated, so can't move it out of header.
```c
FLAMEGPU_CUSTOM_REDUCTION(flamegpu_internal_standard_deviation_add, a, b) {
    return a + b;
}
FLAMEGPU_CUSTOM_TRANSFORM(flamegpu_internal_standard_deviation_subtract_mean, a) {
    return a - flamegpu_internal::STANDARD_DEVIATION_MEAN;
}
```

Robadob:
After taking a short break from this trying to reorient remaining tasks in my head.

Available functionality (all untested)

* Define what needs to be logged of a model
* Pass the config to a model and have it collected at runtime
* Specify the output files

Missing functionality
* Swig implementation of log config/result classes
* Ability to extract logs from `CUDASimulation`, `CUDAEnsemble`
* Ability to send logs to file from `CUDASimulation`, `CUDAEnsemble` (only the path is currently parsed from command line args, the triggering after simulate doesn't actually occur).
* Actual xml/json logging export code.

BryanLaura (approved):

Robadob:
Need to test (some of these overlap):
- [ ] Sim logging config, actually logs the right stuff.
- [ ] Ensemble logging config, actually logs the right stuff.
- [ ] Log values are calculated correctly
- [ ] Programatic access to the log is correct
- [ ]  Log to json file is output correct (this will require a custom parser?)
- [ ] Log to xml file is output correct (this will require a custom parser?)

Robadob:
Example of the output format from a `CUDASimulation` log file.
https://gist.github.com/Robadob/e915c15b9cdcee19d69f734af311aa38

Robadob:
Have discussed with Pete re:log parsing. Best idea is to have a standalone parser amongst FGPU2 lib, so that C++/Python users can import logs from disk without needing to write their own json/xml parser.

BryanLaura (approved):

BryanLaura (approved):

Robadob:
I think the C interface is now complete and tested, as far as has been agreed. The only noteworthy point, being that the format of logged json/xml files is not tested. It was decided that a parser for these logs would be independent from the main FGPU (e.g. small header/python lib?), which is considered a low priority at this time.

All that remains now is to fix any issues with the Python interface (there are alot of curious warnings at compile, and Ubuntu 16 CI is failing whilst building pyflamegpu), and test it the same.

Robadob:
Python tests are all writtten.

Just need to chase this access violation now when running host functions inside. I did have the test pass once, so it might be temperamental (as with most multi-threading issues).

Running `py.test test_logging.py` alone simply exits with no detail.

When ran from the tests root with an empty `py.test` more detail is provided:

```
io\test_logging.py Windows fatal exception: access violation

Thread 0x00002880 (most recent call first):
  File "c:\users\robadob\fgpu2\build\lib\windows-x64\python\venv\lib\site-packages\pyflamegpu\pyflamegpu.py", line 5700 in setVariableUInt
  File Windows fatal exception: access violation

```

My hypothesis would be that as host function callbacks are encapsulated inside an object, a proper copy of the callback needs to be made rather than passing a pointer. But I might be wrong. Need to look back into how they actually work.

Robadob:
Besides the CI failure, the core functionality of Ensembles is now functional and tested in both C and Py so I'm happy for this to be merged.

Need to discuss the unticked Items from this PR to decide whether they should be dropped or new issues created.

Robadob:
Need to solve the Ubuntu 16.04 issue (albeit I would presume even master is broken by the same issue), but otherwise I'm happy for this to be merged. Need to review the comments in the thread and to decide if new issues need to be made.

PR #434, opened 2021-02-02
RTC Cache

### LINKED ISSUE #419: RTC AgentFunction Caching
With the completion of #418 and (eventually) `CUDAEnsemble` #245. The premise of executing RTC (or python) models concurrently is a realistic possibility. Currently this means that every instance must compile the same agent functions. Essentially multiplying the already slow runtime compilation times.

The jitify readme.md states:
> * New * "jitify::experimental API provides serialization capabilities to enable user-managed hashing and caching". 

Which links to this example ([header](https://github.com/rapidsai/cudf/blob/v0.12.0/cpp/src/jit/cache.h)|[source](https://github.com/rapidsai/cudf/blob/v0.12.0/cpp/src/jit/cache.cpp)).

This particular example works by caching rtc programs based on (a user assigned?) name. Given how our RTC functions work, it's possible for two functions to have the same name, or a function to be changed since it was first cached. (E.g. if we also want caching between program executions).

Therefore, our hashing will need to take into account:
* Agent function body (some kind of hash of the string could work, or we write it to disk)
* FLAMEGPU2 version id (this is currently based on a header that contains commit hash).
* model name
* agent name
* agent function name
* cuda version?
* sm version?

The alternative would be to compare flamegpu2 version and the two rtc sources for exact match.

~It has also come to my attention that there no guarantee that the rtc_caches for each instance's environmental variables will be stored in the same order. Therefore this should also be remedied.~ (#425 should solve this)

The `RTCMultiThreadDeviceTest` test suite is appropriate way for assessing whether this change works as intended.

Robadob:
Adds in-memory and on-disk caching of RTC code. This vastly improves performance of repeated RTC runs by removing the several second compilation stage per agent function (condition).
On disk cache files are stored to `$tmp/fgpu2/jitifycache(/debug)`, this can be overriden with env var `FLAMEGPU2_TMP_DIR`.
This change required moving to the `jitify::experimental` interface and updating the RTC dynamic header so that multiple instances of the same `ModelDescription` share a common header.

Closes #419 

# Todo
* Ability to auto purge old cache items?


Robadob:
Have attempted running RTC tests with the new commit get an error at compilation. of what appears to be all agent functions.

I have both reordered the compilation code, and switched to `jitify::experimental`. Need to have a look through and see if im doubling up the source code somehow.

```
-- JIT compile log for Fn1_program ---
__nv_name_map(2): error: cannot determine which instance of function template "agent_function_wrapper" is intended

__nv_name_map(2): error: Error in parsing name expression for lowered name lookup. Input name expression was: " agent_function_wrapper < SlowFn_impl >"

2 errors detected in the compilation of "Fn1_program".

```

Robadob:
Unrelated to the issue in the above comment. In the case I'm debugging above the dynamic headers differ due to the hash used for CURVE. This will impact how suitable the caching is.

Robadob:
I have now fixed the issue in the previous commit. I had accidentally removed a branch around agent functions and conditions, and replaced it wrong. Not sure why I thought this wasn't affecting the first function of each test.

In debugging that however I notified that `JitCache` was already acting as an in memory cache for RTC code. However, due to the use of CURVE and the dynamic header, no two instances of a CUDAModelDescription` share the same headers (otherwise they would have Curve collisions!). Therefore it will be necessary to repackage RTC curve so that all instances have the same memory layout. This is not as bad as it may seem,  as currently the RTC compilations are not linked. So they each have a unique copy of the whole Curve hash table.

A further issue I for see is that it's possible we will need to serialise the kernel or kernel instance. At which point they may share memory layout, making the above point moot. I haven't yet reached that stage of testing though.

Robadob:
The functionality now appears to work fully. I just need to make some quality of life changes.

Robadob:
RTC tests (not Python or Seatbelts) all pass on Mavericks and locally.
Also ran a second time to check speedup worked on mav, it did.

All that remains is whether we want to auto remove old cache files. But that requires some discussion as a team.

PR #439, opened 2021-02-10
Swig Visualisation Support

### LINKED ISSUE #436: Python support for Visualisation.
The visualisation components hidden behind the `VISUALISATION` macro and in the external [visualisation repo(?)](https://github.com/FLAMEGPU/FLAMEGPU2_visualiser) are not currently swigified.

Initial assumption is that we only need to swigify the files found in `include/flamegpu/visualiser` and the method `CUDASimulation::getVisualisation()`.

It's unclear why `CUDASimulation::getVisualisation()` isn't already being converted by SWIG (and then leading to a compile error), as `CUDASimulation` is already being swigified. It's possible we simply need to wrap the point at which swig swigifies the vis files with the C include of `#ifdef VISUALISATION`.

Robadob:
Adds visualisation support to SWIG and a SWIG boids example to demonstrate vis. Prior to this work, running pyflamegpu when built with visualisation support would require SDL2 and Glew dlls, despite their functionality not being available.

Closes #436

**Todo:**
- [x] Test the swig boids example
- [x] Test that swig built without vis still works properly
- [x] Test on linux (does that need any libraries packaged?)
- [x] Add `pyflamgpu.VISUALISATION`, to be set `True` if built with vis support (*The code for **this is a bit icky**, as it involves redefining the `VISUALISATION` macro to false when vis is not enabled, reordering the swig file could make this break stuff which checks `#ifndef VISUALISATION`*)
- [x] How to package dlls required by visualiser into python wheel ([e.g. this](https://vinayak.io/2020/10/21/day-51-bundling-dlls-with-windows-wheels-the-package-data-way/), *Note: Until this is done, the dlls from the build directory must be manually copied to the working directory to run any pyflamegpu code built with vis enabled.*)
- [ ] Cleanup CMake so options which set compile definitions are in common place for swig and C? (*Not 100% sure how doable this is, after had to switch SWIG CMake to calling slightly different version of set target property.*)
- [ ] Sort out `Stock::Models` namespace (*It works now as `pyflamegpu.ICOSPHERE` or similar, as the namespace is basically dropped*, [docs](http://www.swig.org/Doc3.0/SWIGDocumentation.html#Python_nn25))


Robadob:
Re: packaging dlls in wheels. There seems to be debate that you shouldn't package a dll with it's common name, as there could be conflicts if 2 packages try to install dlls with same name but different version.
NUMPY solves this by renaming the dll it imports to `libopenblas.NOIJJG62EMASZI6NYURL6JBKM4EVBGM7.gfortran-win_amd64.dll` or similar. I don't understand enough about where the wheel lib is extracted to, to know whether we need to go that far.

Robadob:
The issue with `pyflamgpu.VISUALISATION`  is that SWIG automatically converts all `#define` to properties of the module. Basic defines with no value, evaluate to false and others evaluate to their specified value.

Therefore, to add `pyflamegpu.VISUALISATION` we need to undef the existing define, and then redefine with `true` or `false`. This obviously has the downside of any code included after this statement in the swig file that touches visualisation will see `VISUALISATION` as defined even if defined to `false`.

Options being:
* Repo wide replace `VISUALISATION` macro with something else
* Repo wide change `#ifdef VISUALISATION` to `#if defined(VISUALISATION) && VISUALISATION` (or similar)
* Rename the macro we want to expose via swig.
* Do nothing, and make users instead use the python `isattr()` fn to check whether the attribute exists.

I would probably prefer renaming the  macro, though I haven't got any suggests.

Robadob:
I've had a play with statically linking SDL2 and Glew inside flamegpu2_visualiser.

Glew is merely a case of linking against a different file from download. This does create a link warning, as it wants to mink against the static MS C lib, whereas everything else wants to link against the dynamic. But this can be suppressed with the `\NODEFAULT` commands.

SDL2 however, does not provide a compiled static lib for MSVC. Instead you must build it. On building and testing, I receive a larger number of unresolved external symbols. Even through non-sdl components of flamegpu2lib. Telling SDL2 to use system lib C (`LIBC` option in cmake) fixes many of these but still a handful remain. Googling suggests the solution is to link against some extra libraries from windows toolkit.

Robadob:
Wheel works and testing on windows showed that as expected bad filenames in setup.py fail silently, allowing wheel creation to continue.

Issue #246, opened 2020-04-30
AgentPopulation Refactor


ptheywood:
`AgentPopulation`s are currently quite limited in their use.

+ Create new populations of a fixed initial size
+ Instanciate members of the population on the host by `getNextInstance`
+ Set a population for a type of agent in a `CUDAAgentModel`, overwriting the existing population using `setPopulationData`
+ Get the population data from a `CUDAAgentModel` by `getPopulationData`
    + Directly access members by index
+ Initialise Population(s) from an xml file
+ Output a Population to an xml file

It is not currently possible to:

+ Combine or split populations (if this is desired)
+ Remove individuals from a population on the host
+ Cleanly iterate a population
    + as a whole, or iterate a filtered population representing a agents in a specific state
+ Populations are not available inside host or step functions.

There is also the outstanding issue of ambiguity `getNextInstance` and `getInstanceAt` (issue #117).

Additionally, Populations do not belong to a model or simulation, preventing things like agent-wide unique id generation (for a simulation, or all simulations implementing a model in a given implementation).

The main suggestion here would be to make it so that populations must belong to a `Simulation` (i.e. a `CUDAAgentModel`), or perhaps a `model` (although this might lead to other issues). 
Unfortunately this would make testing Populations more awkward, as they would require a `CUDAAgentModel` to be instanciated.

I.e. 
```cuda
cuda_model.newPopulation(model.agent("agent"), 12);
// or 
mode.newPopulation("agent", 12);
```

The other main improvement would be to implement iterator(s) to iterate over a population of agents (and perhaps a state-specicif population.)

I.e. 

```cuda
AgentPopulation population = //...
// Iterate existing members / pre-allocated members
for (auto &agentInstance : population) {
    agentInstance.setVariable<float>("x", 12.0);
    auto y = agentInstance.getVariable<float>("y");
    auto id = agentInstance.id(); // get the id unique to this agent regardless of simulation/population 
    auto index = agentInstance.index(); // Get the index within the population.
}
```

Or to iterate over members in a specific state within the population

```cuda
AgentStatePopulation stateList = population.statelist("statename");
for (auto &agentInstance : stateList) {
    // ...
}
```

It should also be possible to set properties for all members of a population/statepopulation (i.e. memset)

```cuda
AgentPopulation population = // ...
population.setVariable<float>("x", 12.0);
```

And possibly perform reductions over the population (at least on the device population, maybe the host.)

```cuda
[...]

Issue #237, opened 2020-04-24
Refactoring to improve quality pre-release


ptheywood:
Some general refactoring should be done before release to improve code quality, documentation clarity and the API interace. 

+ [ ] Reduce/remove ambiguous terms
    + I.e. use of `FLAMEGPU` for both host and device functions, even though the type is different.
+ [ ] Directory Names / file structuring
    + i.e. `sim` vs `simulation`
+ [x] C++ namespaces?
    + Should namespaces be used to mirror the filestructre - i.e. `flamegpu::gpu::CUDASimulation`
+ [ ] Ensure Class names are appropriate
    + [x] `CUDAAgentModel` -> `CUDASimulation`
    + [x] `HostAgentInstance` - Does not represent one individual but the whole population.  
+ [ ] Ensure classes are contained within the most appropriate file 
    + Typically `classname.h/cpp/cu`, unless an innerclass or some dependency messs prevents this.
+ [ ] Variable name / interface naming concistency
    + mix of `camelCase` and `snake_case` currently used. This is not the end of the world but we should probably try to be consistent.
    + class names in all caps and so on. (i.e. `FLAMEGPU_READ_ONLY_DEVICE_API`)
    + Concistency between different parts of the API
        + [x] I.e. `FLAMEGPU.setVariable<T>(name, value)` vs `environment.set<T>(name, value)` 
+ ~Type name / qualifier consistency~
    + ~`float * ptr` vs `float* ptr` vs `float *ptr`~
    + ~`float & ref` vs `float& ref` vs `float &ref`~
    + ~`const int` vs `int const` (more important for `const int * const` vs `int const * const`~
    + Can't be enforced via linters, so not viable.
+ [ ] fixed width type
    + Do we want to follow google/cpplint style guid one using fixed widths types for anything other than `int`?
        + This reduces portability, which is a myth with CUDA as the backend anyway so no negatives?
+ [ ] Public/Private headers
    + Not all classes implemented should be accessible in user code - these headers can be included in the `src` directory.
    + Not all methods on some library headers should be available to users - i.e. Adding variables with reserved names to `agent_description`s - this should be accessible within `src` (and `tests?`) but nowhere else. PIMPL?
    + This also needs to be conisderd at CMake time - most include_directories should be marked as `private`, while the `include` direcotry should be public.
[...]

PR #443, opened 2021-02-12
Refactor AgentPopulation/AgentInstance

### LINKED ISSUE #117: AgentPopulation::getNextInstance(char*) is misleading
This function creates a new agent and returns that agent. It's name should imply this, it's current name suggests it's some kind of iterator.

Similarly, it would be worth documenting the class properly, to clarify that construction allocates initial memory, but does not create uninitialised agents.

Robadob:
Replaces `AgentPopulation` with new classes `AgentVector`, `AgentVector::Agent`. These have an interface similar to what a `std::vector` of agents would look like, but store data as SoA internally.
`AgentInstance` has also been created, to pass around standalone data for a single agent.
Both C and Py interfaces have full test suites.
Py interface lacks access to `AgentVector` iterator (and some util) methods.
Related to #246 #237
Closes #117


Robadob:
SWIG version of `AgentVector` doesn't have an `AgentVector::erase()` equivalent. The proper equivalent would be pythons `pop()` function but that returns the erased item Which in this case would be an `AgentElement`.

This same issue applies to `RunPlanVec` (they use the same SWIG hack for making it usable like a python list). Although I haven't checked how `RunPlanVec` handles deletion of items in C.

Robadob:
todo rename `AgentView` to `AgentVector::View`

Robadob:
> SWIG version of AgentVector doesn't have an AgentVector::erase() equivalent. The proper equivalent would be pythons pop() function but that returns the erased item Which in this case would be an AgentElement.

I just gave swig a copy of C insert/erase which take size_type rather than iterator, and have no return value (rather than returning an iterator).

ptheywood (approved):
I've not reviewed all ~10k changes, but both test suites pass under linux

PR #444, opened 2021-02-12
Tell CMake which python executable to use to fix windows CI

ptheywood:
This should resolve the windows GHA recent failures, by avoiding `python3.exe`.

This feels much cleaner than modifying CMake to try and detect the symlink (which doesn't work propperly) or using the not recommended `--symlinks` option for venv creation.

### DIFF
--- a/.github/workflows/Windows.yml
+++ b/.github/workflows/Windows.yml
@@ -70,9 +70,10 @@ jobs:
         }
       shell: powershell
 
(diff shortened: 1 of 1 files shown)


Robadob (commented):
Good solution if it works. I had a look through my CMake gui vars earlier and couldnt spot a python one.

Robadob (approved):
That comment was supposed to be approve

ptheywood:
Works in my very simple test repo. Had to specify Root as well as executable for this to work, otherwise I would have stumbled upon it yesterday as a working solution.

PR #453, opened 2021-02-24
NO_SEATBELTS -> SEATBELTS macro + pytest.SEATBELTS

### LINKED ISSUE #452: Pytest failures with SEATBELTS=OFF
When `SEATBELTS` are set to `OFF` 3 python tests fail, as they expect exceptions which are not thrown by seatbelts. 

The equivalent tests are disabled in the c++ test  suite when `SEATBELTS=OFF`.

Python tests have no knowledge of build configuration(s), so the `pyflamegpu` object will need to expose if it was built with SEATBELTS or not. 

See the exposing of `VISUALISATION` through swig via a `DEFINE`   https://github.com/FLAMEGPU/FLAMEGPU2_dev/blob/490bbcaefd289b7b75529196d2c6c97104ce4eaf/swig/python/flamegpu.i#L561-L565

ptheywood:
Change NO_SEATBELTS to SEATBELTS=0|1 in the macro + expose to python … as pyflamegpu.SEATBELTS

This allows for tests in python to be skipped if seatbelts is not available.
Closes #452

Tested under linux locally in release with and without seatbelts (both test suites).


ptheywood:
Getting some RTC errors on my office machine with SEATBELTS off (invalid argument). Suggets the kernel is being launched as if SEATBELTS is on.

Robadob:
> Getting some RTC errors on my office machine with SEATBELTS off (invalid argument). Suggets the kernel is being launched as if SEATBELTS is on.

This could be due to the jitify cache if you made the changes without committing?

Robadob:
You've updated where NO_SEATBELTS is passed to rtc compilation.

https://github.com/FLAMEGPU/FLAMEGPU2_dev/blob/5556f012a510710b3262ee96e53e643a02afab37/src/flamegpu/util/JitifyCache.cu#L182


and where the agent function template is defined.

https://github.com/FLAMEGPU/FLAMEGPU2_dev/blob/no-seatbelts-rename/include/flamegpu/runtime/AgentFunction.h

This suggests, that if it isn't due to the cache that the rtc might not be happy with `SEATBELTS=1` being passed to `--define-macro`. The double = does look a bit weird.

ptheywood:
Yep you were right its the jitify cache, which doesn't realise that `SEATBELTS` is different.
Probably need to consider the compiler flags passed to jitify in the cache?

Robadob:
Yup never considered that, should be trivial to add.

Robadob:
https://github.com/FLAMEGPU/FLAMEGPU2_dev/blob/5556f012a510710b3262ee96e53e643a02afab37/src/flamegpu/util/JitifyCache.cu#L224

You just need to make short/long reference different when SEATBELTS vs NO_SEATBELTS. I suggest appending `_S`, `_NS` respectively. But anything will work.

Robadob:
Actually, just update short reference. Redundant to add it to long too.

ptheywood:
Added SEATBELTS and the various debug flags to the short reference which seems to have done the job.

Now tested on two machines with and without seatbelts.

src/flamegpu/util/JitifyCache.cu
```diff
@@ -218,13 +220,28 @@ std::unique_ptr<KernelInstantiation> JitifyCache::loadKernel(const std::string &
     const std::string arch = std::to_string((status == cudaSuccess) ? util::compute_capability::getComputeCapability(currentDeviceIdx) : 0);
     status = cudaRuntimeGetVersion(&currentDeviceIdx);
     const std::string cuda_version = std::to_string((status == cudaSuccess) ? currentDeviceIdx : 0);
+    const std::string seatbelts = std::to_string(SEATBELTS);
+    std::string debug = std::string();
+#ifdef __CUDACC_DEBUG__
+    debug += std::to_string(__CUDACC_DEBUG__);
```
Robadob, line 226: What do these macros actually evaluate to?

If they all evaluate to `1` when set otherwise not written. You end up with the issue where different combinations could lead to the same  short referenced.

I'm a bit reluctant with regards to adding all this debug trash to short reference (e.g. if it inflates the file name/path too long for certain OS/s). If they're really necessary add them to the long reference. This will mean if the long reference doesn't match the cash state is overwritten. But there's no reason that DEBUG/NDEBUG should differ.

Robadob (changes requested):

src/flamegpu/util/JitifyCache.cu
Robadob: tldr: different short reference means different cache file, different long reference means cache file is overwritten.

ptheywood: They probably don't actually have a value thinking about it. 
The debug/release state is important though. If your release build uses a cache that was build with debug you're not going to be happy about runtime. 

I just chucked them in short ref as suggested for the SEATBELTS value. I'm not familiar enough with the caching to know the difference.
Will sort it tomorrow.

Robadob: It would be a bit more work, but tweaking it so that the entire options vector is added to the long cache would make most sense.

Robadob: This way, changes to options in future will always be considered.

Robadob: Yeah I only specified short_ref for seatbelts because we switch between those two options that having seperate cache files would be useful. The debug flags are internal stuff which shouldn't change ever/often within a specific cache file. So they're better off in long reference as there would be not reason to differentiate cache files.

Robadob (approved):

PR #457, opened 2021-03-02
Hostfn Agent Access

### LINKED ISSUE #51: Feature: Read Agent data on the host
Read (potentially modified) agent data on the host in init/exit/step functions.

### LINKED ISSUE #71: Feature: Modify agent data in step/init functions from the host.
Long term feature.

It might be nice to make `AgentPopulation`, or an extension to `AgentPopulation` available for this.

Robadob:
Requires squash before rebase/merge, haven't pre-squashed because I don't like some components (e.g. automatically performing host agent birth), easier to remove if i can revert some commits.

`HostAgentAPI` direct access to agent data.

**TODO**
- [x] Rename `HostAgentInstance` to `HostAgentAPI` (and all the `FLAMEGPU_` runtime classes too)
- [x] Add `HostAgentTools::getPopulation()`/`setPopulation()`
- [x] Add internal change tracking to `AgentVector` (it doesn't need to be publicly user accessible)
- [x] Make `DeviceAgentVector` use the change tracking.
- [x] Perform auto sync on step function exit
- [x] Perform auto sync on host reductions 
- [x] Download variable data on-demand?
- [x] Do something about Host Agent Birth 🤷‍♂️
- [x] Decide if we want to hide any `AgentVector` functionality.
- [x] C tests
- [x] Implement SWIG interface
- [x] Py tests

Relates to #246
Closes #51 
Closes #71 

Current interface proposal (Changed from @ptheywood's #246, after brief discussion at meeting).

```c++
FLAMEGPU_STEP_FUNCTION(step_function) {
    // State argument can be ommitted if model does not use states
    HostAgentAPI agent = FLAMEGPU->agent("agent", "default");
    // Get a population (HostAgentVector inherits from AgentVector to allow method hiding etc, may not be necessary)
    HostAgentVector population = agent.getPopulation();

    // Perform a parallel reduction over the whole population
    // Variable hasn't changed within pop so device reduction is fine
    int sum_a = population.sum<int>("a");

    // Iterate the population on the host
    for(auto agentInstance : population){
        // Access variables - with JIT copying
        float x = agentInstance.getVariable<float>("x");
        // Host modification, sets flags to push to device when required
        agentInstance.setVariable<float>("x", x+1);
    }
    
    // Support agent death with normal AgentVector mechanics
    population.erase(4);    

    // Also support setting a common value over the entire population.
    // This is not currently present in AgentVector (do we want to make it generally applicable? would it better to force this to occur via custom reduce/transform?)
    population.setVariableAll<float>("x", 1.0f);

    // The population must be restored if (the particular variable is) changed, before a reduction would be applied
    // SEATBELTS warns about this if it's missing
    agent.setPopulation(population);
    float reduction = agent.reduce<float>("x");

    // Host agent birth needs some thought
    // As it's desirable for it to also affect reductions
    // It should be more efficient than creation via AgentVector, but agent vector makes it a bit redundant.
    auto newInstance = population.newAgent();
    newInstance.setVariable<float>("x", 0.0f);

    // If a user forgets to set a modified population, SEATBELTS will complain
    // We need a means of allowing users to flag they don't wish to return a changed population
}
````
------------

Other notes from discussion:
* Is `FLAMEGPU->agentPopulation("agent");` supposed to return an (equivalent to) `AgentVector`
   *  `FLAMEGPU->agent("agent");` currently returns a `HostAgentInstance`, potentially better renamed `HostAgentTools`? This could then have a method `HostAgentTools::getAgentVector()` (or a better name), which returns the `HostAgentVector`.
* If `population` is equivalent to `AgentVector`, why is it `AgentInstance::die()`, rather than `AgentVector::erase(). I would expect consistency to be preferable.
* Should a host api reduction really cause downloaded agent populations to be upload early?
     * If that's the case, should host agent birth also be applied early?
* The downloaded populations are automatically re-uploaded to device.
* Direct access to agents via an `AgentVector` analogue, essentially makes (the more efficient?) host agent birth redundant. Should this be removed?
* Should the host reductions be accessible via the `AgentVector` analogue or should the analogue be distinct, and named in a manner to discourage use?

It seems as though this `HostAgentVector` should wrap a regular `AgentVector`, providing essentially duplicate/compatible functionality with additional features for creating an effective changelog, to reduce data transfer. It will also need to link pretty deeply with the HostAPI object, to be checked whenever a reduction occurs.

Simplifying these mechanics, at the cost of performance could be preferable, so this requires discussion (or rather clarification from @mondus regarding how 'easy' he wants it to be for people to do very expensive host-agent transactions).

It may also be simpler to make the changelog a core `protected` visibility internal component of `AgentVector`, which is updated within normal usage, but only ever read by the HostAPI.

With regards to mechanics of the changelog. Tracking the first and last modified items for each variable is the obvious way to go. Allowing copies to minimised. Obviously certain methods would savage this, and essentially require all agent data to be updated. However by private overloading `AgentVector`, we could selectively expose the permitted methods.

Rough List of question/answers
> Should the downloaded agent data be automatically updated at the end? (otherwise user must explicitly commit changes, we could detect when a user forgot to upload [with seatbelts?]).

Erring on the side of manual upload/download.

> Should an agent reduction force upload agent data and clear the change log? (otherwise a user must explicitly commit changes before the reduction)

Manual, but give users a seatbelts warning/error.

> Should the host-agent reductions be moved inside the `AgentVector` analogue? (or kept separate, in-order to discourage host-agent access, rather than it be a 'default')

Keep separate

> Should `HostAgentInstance` be renamed `HostAgentTools` (or something else?)

Yes

> Should the changelog be built into `AgentVector` to make dependent class far simpler.

Yes
> How much `AgentVector` functionality should be exposed? (e.g. do we want to go as far as allowing them to insert externally defined `AgentVector`s?

Everything for the time being, can restrict it later

> Should the existing (more efficient?) host agent birth be removed? as host agent access duplicates this behaviour.

No

> How should host agent birth interact with host agent access/reductions (should they be distinct or not)

Yes it should interact with reductions

### DIFF
--- a/examples/ensemble/src/main.cu
+++ b/examples/ensemble/src/main.cu
@@ -9,9 +9,9 @@ FLAMEGPU_INIT_FUNCTION(Init) {
     const unsigned int POPULATION_TO_GENERATE = FLAMEGPU->environment.getProperty<unsigned int>("POPULATION_TO_GENERATE");
     const int init = FLAMEGPU->environment.getProperty<int>("init");
     const int init_offset = FLAMEGPU->environment.getProperty<int>("init_offset");
+    auto agent = FLAMEGPU->agent("Agent");
     for (unsigned int i = 0; i < POPULATION_TO_GENERATE; ++i) {
-        auto agent = FLAMEGPU->newAgent("Agent");
-        agent.setVariable<int>("x", init + i * init_offset);
+        agent.newAgent().setVariable<int>("x", init + i * init_offset);
     }
 }
 std::atomic<unsigned int> atomic_init = {0};
--- a/examples/host_functions/src/main.cu
+++ b/examples/host_functions/src/main.cu
@@ -14,11 +14,12 @@ FLAMEGPU_AGENT_FUNCTION(device_function, MsgNone, MsgNone) {
     return ALIVE;
 }
 FLAMEGPU_INIT_FUNCTION(init_function) {
-    float min_x = FLAMEGPU->agent("agent").min<float>("x");
-    float max_x = FLAMEGPU->agent("agent").max<float>("x");
+    HostAgentAPI agent = FLAMEGPU->agent("agent");
+    float min_x = agent.min<float>("x");
+    float max_x = agent.max<float>("x");
     printf("Init Function! (AgentCount: %u, Min: %g, Max: %g)\n", FLAMEGPU->agent("agent").count(), min_x, max_x);
     for (unsigned int i = AGENT_COUNT / 2; i < AGENT_COUNT; i++) {
-        HostNewAgentAPI instance = FLAMEGPU->newAgent("agent");
+        HostNewAgentAPI instance = agent.newAgent();
         instance.setVariable<float>("x", static_cast<float>(i));
         instance.setVariable<int>("a", i % 2 == 0 ? 1 : 0);
     }
--- a/examples/host_functions/src/main.cu
+++ b/examples/host_functions/src/main.cu
@@ -18,7 +18,7 @@ FLAMEGPU_INIT_FUNCTION(init_function) {
     float max_x = FLAMEGPU->agent("agent").max<float>("x");
     printf("Init Function! (AgentCount: %u, Min: %g, Max: %g)\n", FLAMEGPU->agent("agent").count(), min_x, max_x);
     for (unsigned int i = AGENT_COUNT / 2; i < AGENT_COUNT; i++) {
-        FLAMEGPU_HOST_NEW_AGENT_API instance = FLAMEGPU->newAgent("agent");
+        HostNewAgentAPI instance = FLAMEGPU->newAgent("agent");
         instance.setVariable<float>("x", static_cast<float>(i));
         instance.setVariable<int>("a", i % 2 == 0 ? 1 : 0);
     }
--- a/include/flamegpu/gpu/CUDAAgent.h
+++ b/include/flamegpu/gpu/CUDAAgent.h
@@ -7,6 +7,7 @@
 #include <string>
 #include <mutex>
 #include <unordered_map>
+#include <list>
 
 // include sub classes
 #include "flamegpu/util/JitifyCache.h"
@@ -216,7 +217,16 @@ class CUDAAgent : public AgentInterface {
      * @param scatter Scatter instance and scan arrays to be used (CUDASimulation::singletons->scatter)
      * @param streamId This is required for scan compaction arrays and async
      */
-    void initUnmappedVars(CUDAScatter &scatter, const unsigned int &streamId, const cudaStream_t &stream);
+    void initUnmappedVars(CUDAScatter& scatter, const unsigned int& streamId, const cudaStream_t& stream);
+    /**
+     * Initialises any agent variables within the CUDAFatAgentStateList of state which are not present in the agent-state's CUDAAgentStateList
+     * @param state Affected state
+     * @param count Number of variables to init
+     * @param offset Offset into the buffer of agents to init
+     * @param scatter Scatter instance and scan arrays to be used
+     * @param streamId This is required for scan compaction arrays and async
+     */
+    void initExcludedVars(const std::string& state, const unsigned int& count, const unsigned int& offset, CUDAScatter& scatter, const unsigned int& streamId, const cudaStream_t& stream);
     /**
      * Resets the number of agents in every statelist to 0
      */
@@ -226,6 +236,27 @@ class CUDAAgent : public AgentInterface {
      * They count as unmapped if they are not mapped to a master state, sub mappings will be reset
      */
     void cullUnmappedStates();
+    /**
+     * Resize the state underlying buffers
+     * This is triggered on the CUDAFatAgentStateList, so unmapped agent variables are also resized
+     * @param state The minimum number of agents that must be representable
+     * @param minSize The minimum number of agents that must be representable
+     * @param retainData If true existing buffer data is retained
+     */
+    void resizeState(const std::string &state, const unsigned int& minSize, const bool& retainData);
+    /**
+     * Updates the number of alive agents, does not affect disabled agents or change agent data
+     * @param state The state to affect
+     * @param newSize Number of active agents
+     * @throw InvalidMemoryCapacity If the new number of disabled + active agents would exceed currently allocated buffer capacity
+     */
+    void setStateAgentCount(const std::string& state, const unsigned int &newSize);
+    /**
+     * Returns a list of variable buffers attached to bound agents, not available in this agent
+     * @param state The state affected state
+     * @note This access is only intended for DeviceAgentVector's correctly handling of subagents
+     */
+    std::list<std::shared_ptr<VariableBuffer>> getUnboundVariableBuffers(const std::string& state);
 
  private:
     /**
--- a/include/flamegpu/gpu/CUDAAgent.h
+++ b/include/flamegpu/gpu/CUDAAgent.h
@@ -156,7 +156,7 @@ class CUDAAgent : public AgentInterface {
      * @param state_name The state agents are scattered into
      * @param scatter Scatter instance and scan arrays to be used (CUDASimulation::singletons->scatter)
      * @param streamId The stream in which the corresponding agent function has executed
-     * @see HostAgentInstance::sort(const std::string &, HostAgentInstance::Order, int, int)
+     * @see HostAgentAPI::sort(const std::string &, HostAgentAPI::Order, int, int)
      */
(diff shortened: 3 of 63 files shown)

### CODE REVIEW

**Comment by ptheywood** [comment]
> If population is equivalent to AgentVector, why is it AgentInstance::die(), rather than `AgentVector::erase(). I would expect consistency to be preferable.

I'll have just been thinking about  it from a modelling perspective (i.e. agent death) hence `die()`. I agree that consistency is preferable.

> With regards to mechanics of the changelog. Tracking the first and last modified items for each variable is the obvious way to go. Allowing copies to minimised. Obviously certain methods would savage this, and essentially require all agents to be updated.

For (very/quite) small pops (of relatively skinny agents) in F1 it was cheaper to just copy  the entire SoA at once than do many small copies. 
I think in the end I implemented explicit copying of a requested variable at once. 
Given memcpy latency and higher achieved bandwidth for larger copies first-last tracking should be fine IMO. 


> * [ ]  Should the downloaded agent data be automatically updated at the end? (otherwise user must explicitly commit changes).

Either could work, its a trade off between ease of use and performance. Explicit is probably easier for us, but does shift some extra thought onto the modeller.

> * [ ]  Should an agent reduction force upload agent data (and clear the change log)

If automatic h2d then yes, if not then no?
For small vectors it might even be worth adding host based reductions (cheaper to just do it on the host than doing h2d, reduction, d2h) but that's not a priority.

> * [ ]  Should the host-agent reductions be moved inside the `AgentVector` analogue?

If they are only applicable to agents in a single state then yes.
If agnets in multiple states are contiguous and the reductions can be applied across all agents of a given type thne no (I don't think this is currently the case?)

> * [ ]  Should `HostAgentInstance` be renamed `HostAgentTools` (or something else?)

Yes. It doesn't represent an individual agent anyway in the host api.

> * [ ]  Should the changelog be built into `AgentVector` to make dependent class far simpler.

It's probably a bitfield per variable in the class so shouldn't be too much overhead? I'm indifferent on this really. 


> * [ ]  How much `AgentVector` functionality should be exposed? (e.g. do we want to go as far as allowing them to insert externally defined `AgentVector`s?

Not sure if inserts required. Appending a separate agent vector could be useful (i.e. mass birth)

**Comment by Robadob** [comment]
ffs only linted the lib and forgot to commit the changes to the test suite.

**Comment by Robadob** [comment]
It's now pyfgpu failing, which I haven't looked into (didn't expect it to break, but quite a few reasons it might).

**Comment by Robadob** [comment]
I've had a look into the pyflamegpu build failure.

It's related to the private structure `DeviceAgentVector::VariableBufferPair` not having a copy constructor. For some reason SWIG requires it have to have one, but regular compilation doesn't (perhaps SWIGs replacement for `#include <list>`?).

Not immediatley clear how to fix, so will leave broken until I have the rest of the PR implemented.

**Comment by Robadob** [comment]
Host agent birth seems a bit complicated.

* Host birth only, existing behaviour
* Host birth before DeviceAgentVector (is changed), copy agents from host birth structure to DeviceAgentVector (trigger insert directly after construction?)
* Host birth after DeviceAgentVector (if changed, but before sync), copy agents directly to DeviceAgentVector (essentially trigger insert?).

**Comment by Robadob** [comment]
> Host agent birth seems a bit complicated.
> 
> * Host birth only, existing behaviour
> * Host birth before DeviceAgentVector (is changed), copy agents from host birth structure to DeviceAgentVector (trigger insert directly after construction?)
> * Host birth after DeviceAgentVector (if changed, but before sync), copy agents directly to DeviceAgentVector (essentially trigger insert?).

I think the simplest way to handle this, will be to make the HostAgentBirth buffer/s update the DeviceAgentBirth buffers when a `_requires()` method is called. I think I can tweak those, so the requires is passed pos, before the index is validated.

**Comment by Robadob** [comment]
Test which remain:
- [x] `HostAgentAPI` reductions sync changes from `DeviceAgentVector`
- [ ] `HostAgentAPI` sort sync changes from `DeviceAgentVector` before, and `DeviceAgentVector` syncs after (not recommended, but possible).
- [ ] Host agent birth is noticed by all `DeviceAgentVector` methods (and `HostAgentAPI` reductions?).

**Review by ptheywood — CHANGES_REQUESTED** [review]
Generally looks good, although I've not been too thorough given the number of changes. 

7 tests fail under linux with seatbelts on in release mode:

```
[==========] 764 tests from 55 test suites ran. (345596 ms total)
[  PASSED  ] 757 tests.
[  FAILED  ] 7 tests, listed below:
[  FAILED  ] HostReductionTest.SumException
[  FAILED  ] HostReductionTest.MinException
[  FAILED  ] HostReductionTest.MaxException
[  FAILED  ] HostReductionTest.CustomReductionException
[  FAILED  ] HostReductionTest.HistogramEvenException
[  FAILED  ] HostReductionTest.CustomTransformException
[  FAILED  ] HostReductionTest.CountException
```

```
[ RUN      ] HostReductionTest.SumException
/home/ptheywood/code/flamegpu/FLAMEGPU2/tests/test_cases/runtime/host_reduction/test_misc.cu:15: Failure
Expected: FLAMEGPU->agent("agedddnt") throws an exception of type InvalidCudaAgent.
  Actual: it throws std::out_of_range with description "_Map_base::at".
[  FAILED  ] HostReductionTest.SumException (0 ms)
[ RUN      ] HostReductionTest.MinException
/home/ptheywood/code/flamegpu/FLAMEGPU2/tests/test_cases/runtime/host_reduction/test_misc.cu:22: Failure
Expected: FLAMEGPU->agent("agsssedddnt") throws an exception of type InvalidCudaAgent.
  Actual: it throws std::out_of_range with description "_Map_base::at".
[  FAILED  ] HostReductionTest.MinException (1 ms)
[ RUN      ] HostReductionTest.MaxException
/home/ptheywood/code/flamegpu/FLAMEGPU2/tests/test_cases/runtime/host_reduction/test_misc.cu:29: Failure
Expected: FLAMEGPU->agent("ageaadddnt") throws an exception of type InvalidCudaAgent.
  Actual: it throws std::out_of_range with description "_Map_base::at".
[  FAILED  ] HostReductionTest.MaxException (0 ms)
[ RUN      ] HostReductionTest.CustomReductionException
/home/ptheywood/code/flamegpu/FLAMEGPU2/tests/test_cases/runtime/host_reduction/test_misc.cu:36: Failure
Expected: FLAMEGPU->agent("ageaadddnt") throws an exception of type InvalidCudaAgent.
  Actual: it throws std::out_of_range with description "_Map_base::at".
[  FAILED  ] HostReductionTest.CustomReductionException (1 ms)
[ RUN      ] HostReductionTest.HistogramEvenException
/home/ptheywood/code/flamegpu/FLAMEGPU2/tests/test_cases/runtime/host_reduction/test_misc.cu:43: Failure
Expected: FLAMEGPU->agent("ageaadddnt") throws an exception of type InvalidCudaAgent.
  Actual: it throws std::out_of_range with description "_Map_base::at".
[  FAILED  ] HostReductionTest.HistogramEvenException (1 ms)
[ RUN      ] HostReductionTest.CustomTransformException
/home/ptheywood/code/flamegpu/FLAMEGPU2/tests/test_cases/runtime/host_reduction/test_misc.cu:52: Failure
Expected: FLAMEGPU->agent("ageaadddnt") throws an exception of type InvalidCudaAgent.
  Actual: it throws std::out_of_range with description "_Map_base::at".
[  FAILED  ] HostReductionTest.CustomTransformException (0 ms)
[ RUN      ] HostReductionTest.CountException
/home/ptheywood/code/flamegpu/FLAMEGPU2/tests/test_cases/runtime/host_reduction/test_misc.cu:59: Failure
Expected: FLAMEGPU->agent("ageaadddnt") throws an exception of type InvalidCudaAgent.
  Actual: it throws std::out_of_range with description "_Map_base::at".
[  FAILED  ] HostReductionTest.CountException (1 ms)

```

**Comment by Robadob** [comment]
Fixed all the tests I forgot to check.

Windows/Debug/Seatbelts
```
[==========] 750 tests from 54 test suites ran. (882411 ms total)
[  PASSED  ] 750 tests.

  YOU HAVE 18 DISABLED TESTS
```

```
================================= 445 passed in 106.37s (0:01:46) =================================
```

ptheywood (approved):
Tests all pass under linux now

Robadob:
Todo: query PIMPL in ukrse slack cpp channel (#466).

Robadob:
C and Python tests pass.

The current version has `HostAgentAPI` return a `DeviceAgentVector` which is just a typedef for `DeviceAgentVector_impl&`.

This has the benefit over my previous iteration, of being more maintainable (all the functions aren't manually wrapped. However, creates the problem of us being unable to detect if a user holds onto the reference after the end of the host function it was created in.

PR #458, opened 2021-03-03
Dependency analysis

MILeach:
Automatic generation of layers from dependency specification.

- [x] Create `DependencyNode` base class which agent functions inherit from
- [x] Refactor current code to use `DependencyNode` as base rather than `AgentFunctionDescription`
- [x] Refactor `AgentFunctionDependencyGraph` to `DependencyGraph`
- [x] Extend to support addition of `HostFunctionDescription`
- [x] Extend to support submodels
- [x] Update DOT diagram generation
- [x] Add name field to Submodels
- [x] Update tests to take changes into account
- [x] Build python build
- [x] Add python tests
- [x] Handle HostFunctionCallbacks
- [x] Create example project demonstrating usage
- [x] Review code/check in line with conventions/linting
- [x] Specify multiple dependencies at once
- [x] Invert usage of generateLayers
- [x] Remove user facing raw pointers
- [x] PIMPL - skipping for now
- [x] Update model copy constructors etc.
- [x] Validate added functions belong to same model as graph
- [x] Give warning for unattached functions
- [x] Update python tests
- [x] Check for pre-existing layers in generateLayers()
- [x] Remove printGraph
- [x] Doc comments for DependencyGraph
- [x] Fix LayerDescription.cpp error messages
- [x] Resolve swig ifdefs
- [x] Check tests


Robadob (changes requested):
The changes to ModelDescription hierarchy need to be reflected inside the relevant copy constructors and equality operators.

ptheywood (approved):
Both test suites pass for me. 

Couple of questions/points though I might have missed things while skimming the changes:

+ What happens if a model already has layers and you attemtpt to add a dependency graph to it? 
    + I'm not sure what *should* happen, probably an exception?
+ `dependencyGraph.generateLayers(model);` feels a bit grim, given it mutates `model` and not `dependencyGraph`, but given how this generates layers (which belong to the model) it makes some sence. 
    + An alternative might be that the model provides a dependency graph which it owns (similar to how it provides and owns each new layer), given the graphs need to be associated with a model to be useful anyway? 

Also need to make sure history is squashed when this gets merged, either to a single commit or a couple of commits (one to add all the functionality, one to add the new example)

Robadob:
> * An alternative might be that the model provides a dependency graph which it owns (similar to how it provides and owns each new layer), given the graphs need to be associated with a model to be useful anyway?

I thought this was discussed last week, and came to the conclusion that depedency graph would exist at the root of the ModelDesc hierarchy, but be accessed through AgentFunctionDescs etc.

MILeach:
>     * What happens if a model already has layers and you attemtpt to add a dependency graph to it?
>       
>       * I'm not sure what _should_ happen, probably an exception?
> 
>     * `dependencyGraph.generateLayers(model);` feels a bit grim, given it mutates `model` and not `dependencyGraph`, but given how this generates layers (which belong to the model) it makes some sence.
>       
>       * An alternative might be that the model provides a dependency graph which it owns (similar to how it provides and owns each new layer), given the graphs need to be associated with a model to be useful anyway?

Both good points - I want to have a bit of a discussion regarding what if anything owns the dependency graph, or whether it is simply used to generate the layers. If the latter, the interface could be improved by the model having a loadDependencyGraph method or something which internally calls dependencyGraph.generateLayers(*this). It depends if we want the graph to be persistent or just used to initialise the model.

Robadob (changes requested):
Your tests validate that the graph is structured as expected, but I couldn't spot a test that the graph has actually been converted to layers properly (the main thing users will actually be using this for). This would probably require some kind of test that executes a bunch of host functions in order which manip global or environmental variables, and throws an assert directly from a host function if it's called in the wrong order.

I'm not familiar enough with your dependency stuff to know how complicated that test would need to be.


----------

As Pete has mentioned, `DependencyGraph::generateLayers()` is pretty grim. I would probably have a method something like `ModelDesription::initLayer(DependencyGraph &)` Which takes and stores a copy of the graph internally, and then uses it to parse the graph into layers. As with most of model description hierarchy, you can use `friend` statements to give a particular class or class member function, private access to the class.

-------------

I'm not too bothered about `DependencyGraph` being created out of thin-air, but it would perhaps be better that the `DependencyGraph` takes the `ModelDescription` at construction (or is constructed from the `ModelDescription`), so that it can be validated that the dependency is from the same `ModelDescription` when `dependsOn()` is called. This is pretty major, it's better to report an exception as soon as possible, so the cause can be more clearly identified.

-----------

`dependsOn()` and a bunch of other methods pass raw pointers, this is a big nonono. We don't want users to see pointers (or references, but that's a work in progress).

----------- 

Your new classes don't follow Paul's vision for ~PIMPL.

I'm not sure how strictly important that is here, but it's possible he will want them broken up into interface/data.

Robadob (changes requested):
These changes are all much smaller. Otherwise looks much better

* `DependencyGraph.h` only has a class level doc comment, methods ned them too
* `DependencyNode.h` would it be more correct to include appropriate `@param`, `@throws` in the method doc comments here? they look a bit sparse.
* `DependencyGraph.cpp` you use `#ifdef SWIG`, this will never be true, as the SWIG macro is only visible in headers included by the swig `.cxx` file (SWIG does not rebuild `flamegpu2.lib`, with SWIG=true). This suggests it's something you haven't tested properly,  doesn't appear to you check layers match expected in your py tests anywhere. I would always just handle both in all cases.
* `DependencyGraph::printGraph()`: Either implement this, or remove it, a 'not yet implemented message' shouldn't reach master (replacing it with a `//@todo` + git issue would also suffice).

* `LayerDescription.cpp` I fixed some copy-pasted error messages the other day, I think your changes are trying to revert those, see `addHostFunction()` error message states in `addSubModel()`

* In your tests, I would probably check that `ModelDescription::getLayersCount()` returns `0` before the call to `_m.generateLayers()` and `>0` after, at current it's not clear what you're expecting to happen in tests such as `DependencyGraphTest::AddSubmodelAsDependency`, did you mean to wrap `_m.generateLayers()` with `EXPECT_NO_THROW()`?

include/flamegpu/model/LayerDescription.h
```diff
@@ -319,8 +314,16 @@ void LayerDescription::addAgentFunction(AgentFunction /*af*/) {
         "in LayerDescription::addAgentFunction().");
 }
 
-// Not included in ifdef SWIG intentionally - needs to be available for DependencyGraph to call
+// Can't be moved to cpp file because of inline hint
```
Robadob, line 317: So remove the inline hint??

src/flamegpu/model/DependencyGraph.cpp
```diff
@@ -265,9 +265,13 @@ void DependencyGraph::checkForUnattachedFunctions() {
...
     };
 
+    for (auto& root : roots) {
+        captureFunctions(root);
+    }
+
     // Compare sets
     if (modelFunctions != graphFunctions) {
-        std::cout << "WARNING: Not all agent functions are used in the dependency graph - have you forgotten to add one?";
+        std::cout << "WARNING: Not all agent functions are used in the dependency graph - have you forgotten to add one?" << std::flush;
```
Robadob, line 274: Should this be `cerr`? (Honestly not sure what team preference would be).

src/flamegpu/model/LayerDescription.cpp
```diff
@@ -39,6 +39,9 @@ void LayerDescription::addAgentFunction(const std::string &agentName, const std:
         THROW InvalidLayerMember("A layer containing host functions, may not also contain agent functions, "
             "in LayerDescription::addAgentFunction()\n");
     }
+    if (!layer->host_functions.empty() || !layer->host_functions_callbacks.empty()) {
+        THROW InvalidLayerMember("A layer containing a host function may not also contain an agent function");
```
Robadob, line 43: "in LayerDescription::addAgentFunction()\n");

include/flamegpu/model/LayerDescription.h
MILeach: Assumed whoever wrote it put it there for good reason but I can remove it haha

src/flamegpu/model/DependencyGraph.cpp
ptheywood: Warnigns / erros shoudl go to stderr IMO. We've tended to use `fprintf` rather than `std::cout/std::cerr` although that's mostly just habit (although there's generally a lot of dislike in c++ for output streams for some reason?)

Robadob: Yeah, i don't think use of std::cout vs printf matter, I was mostly concerned that warnings probably don't want to stderr rather than stdout.

In general, console IO has been minimised in the codebase, preferring to throw exceptions. But this warning seems reasonable, otherwise users would need to specify they don't want to use an agent function, or comment it out which seems a bit arkward.

MILeach: I have updated to use cerr in light of this.

Robadob:
Hmm, your CI failure is complaining that.

```
flamegpuPYTHON_wrap.obj : error LNK2019: unresolved external symbol "public: void __cdecl LayerDescription::addHostFunctionCallback(class HostFunctionCallback *)" (?addHostFunctionCallback@LayerDescription@@QEAAXPEAVHostFunctionCallback@@@Z) referenced in function _wrap_LayerDescription_addHostFunctionCallback [D:\a\FLAMEGPU2\FLAMEGPU2\build\swig\python\pyflamegpu.vcxproj]
```

It appears MSVC encodes the visibility of a member function into it's C++ name mangled name (Given gcc doesn't care). I guess the simple hack, is to instead define a second inline host function `_addHostFunctionCallback()` (wrapped with `#ifdef SWIG`, which forwards to the private one.

Robadob (approved):
As I said before, I think tests which are expecting no throw, should wrap the call with an `EXPECT_NO_THROW`.

a) Makes the test syntax clearer (e.g. why are you calling these functions at the end if you're not checking their impact)
b) Improves the reporting by googletest


These C tests all suffer from it, less bothered about pytest as that doesn't even have a proper way of reporting exceptions.

```
AddSubmodelAsDependency
AddSubmodelAsDependent
AddHostFunctionAsDependency
AddHostFunctionAsDependent
ConstructLayersSingleChain
```


Otherwise it's looking good, and I think you've addressed all my other points.

PR #462, opened 2021-03-05
Visualiser: Agent/State colour customisation

Robadob:
Adds the ability for a modeller to change the colours of agents/agent-states.

Adds new classes `Color`, `Palette` (with some stock provided instances) and several helpers for configuring how an agent decides it's Color.

Agent's are now coloured according to the Dark2 palette by default. Agent-state's all share the same colour, until `AgentVis::State()` is called for each state.

Example usages:

```c++
        AgentVis &agent = ...
        agent.setColor(Stock::Colors::WHITE);
        agent.setColor(Color{"#aabbcc"});
        agent.setColor(Color{1.0f, 1.0f, 1.0f});
        agent.setColor(DiscreteColor("x", Stock::Palette::DARK2, Stock::Colors::WHITE));
        agent.setColor(DiscreteColor("x", Stock::Palette::Viridis(10), Stock::Colors::WHITE));
        agent.setColor(HSVInterpolation::REDGREEN("x"));
        agent.setColor(HSVInterpolation("x", 0.0f, 360.0f));
        agent.setColor(ViridisInterpolation("x", 0.0f, 360.0f));
```

Testing this branch requires the [branch of the same name in the visualiser repo](https://github.com/FLAMEGPU/FLAMEGPU2_visualiser/pull/33). Easiest way to manage this is to clone/checkout that repo/branch and set the CMake variable `VISUALISATION_ROOT` to the root of that repo.


Robadob:
Not a fan of the current Palette UI. Ideally it would be an iterable/array accessible enum. But that's not how C++ works.

So latest idea looks like

```c
struct Test: :: public Palette {
    enum Tag {BLUE, RED,};
    StaticColor operator[](int);
    StaticColor operator[](Tag);
    unsigned int size();
    iterator begin();
    iterator end();
    static const StaticColor [] colors {};
}
````

MILeach (approved):

PR #463, opened 2021-03-07
Reduce RTC memcpys

Robadob:
Updates RTC so a single memcpy is performed per agent function/agent function condition. Can't make it leaner than this, as all agent functions are not linked, so require separate memcpys (unless we want another layer of indirection, so that every rtc func holds a ptr to common storage, but with most models 1 func per layer seems harmless).

Rough changelist:
* `CurveRTCHost` now has a fat buffer, which it stores all env data and curve variable pointers in.
* Data is ordered in the cache; Env data, agent ptrs, msg out ptrs, msg in ptrs, new agent ptrs. (This relies on their order remaining consistent, e.g. no new agent vars added to header after it's initial creation.
* `CUDAAgent` now holds onto `CurveRTCHost` instances when it creates header.
* The various map runtime variables methods now instead call e.g. `CurveRTCHost::getAgentVariableCachePtr()`, to retrieve the pointer to store that variable's data within the RTC header's host cache.
* After the function/condition is full init, `CurveRTCHost::updateDevice()` is called to perform the memcpy.
* Also disabled RTC functions from calling the non-rtc Curve/Env updateDevice methods. (Though I haven't removed the now redundant RTC internals from env).
* Also added a branch to the map runtime variables methods, so RTC doesn't bother updating Curve.

### DIFF
--- a/include/flamegpu/gpu/CUDAAgent.h
+++ b/include/flamegpu/gpu/CUDAAgent.h
@@ -13,6 +13,7 @@
 #include "flamegpu/gpu/CUDAAgentStateList.h"
 #include "flamegpu/model/AgentFunctionData.h"
 #include "flamegpu/model/SubAgentData.h"
+#include "flamegpu/runtime/cuRVE/curve_rtc.h"
 #include "flamegpu/sim/AgentInterface.h"
 
 class CUDAScatter;
@@ -30,7 +31,8 @@ class CUDAAgent : public AgentInterface {
     /**
      *  map of agent function name to RTC function instance
      */
-    typedef std::map<const std::string, std::unique_ptr<jitify::experimental::KernelInstantiation>> CUDARTCFuncMap;
+     typedef std::map<const std::string, std::unique_ptr<jitify::experimental::KernelInstantiation>> CUDARTCFuncMap;
+     typedef std::map<const std::string, std::unique_ptr<CurveRTCHost>> CUDARTCHeaderMap;
     /**
      * Element type of CUDARTCFuncMap
      */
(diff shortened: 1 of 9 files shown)


Robadob:
Fixed alot of small bugs in this change.
One larger issue was that the EnvMgr rtc cache was never being pushed to device. I falsely believed  it would be updating the header cache (despite that pointer never being shared).

Robadob:
Finally think I've fixed it. First 5 or so tests have passed. Need to remove some prints. Commit, rerun full suite.

Robadob:
Full C test suite passes on Windows/Debug/SeatbeltsOn

ptheywood:
Some quick profiling of `boids_rtc_spatial3d` under linux shows good improvements to the time spent performing memcpy. 

For a single iteration of boids, `39` h2d memcpy are replaced by `2` memcpy, corresponding to `~330us` vs  `~11us` per iteration.

For the concurrency benchmarking this will improve both the serial and concurrent cases, but by reducing the time of serial portions the relative portion of each step which is concurrent will increase. 

## Timelines showing the copy region. 

Scales are not exaclty the same because that's hard to do in nsys.

Current master

image


This branch 

image

Robadob:
image
6 fails in the full SWIG suite. 5 seem related to outputting a different agent type.

Investigating.

Robadob:
Should be fixed now, It was a tiny error in mapping new agents (using the newagent CUDAAgent, rather than agent fn CUDAAgent.

The Brute force msg failure is more unusual, was an assert fail with getting a wrong result (some small negative value, should have been 48?). After applying this fix, brute force tests all passed 10x in a row though, so I will ignore it until it resurfaces.

ptheywood (approved):
```
============================= 442 passed, 3 skipped in 894.11s (0:14:54) ==============================
```

All rtc tests pass under linux

PR #474, opened 2021-03-19
Fix Windows CI

Robadob:
An update to the CI server between Sunday/Monday has caused windows builds of the lib to fail due to __floorf() not being defined in device code.

Appears the latest update to visual studio is to blame: https://devtalk.blender.org/t/cuda-compile-error-windows-10/17886/5


Robadob:
Worth testing this with RTC, as it affects device headers. I'm building the boids_rtc_spatial3D example currently, but I might need to dash off before it's finished.

Robadob:
Have confirmed that `boids_rtc_spatial3D` still works after this update to spatial messaging device headers.

PR #475, opened 2021-03-19
Fix some visualisation related issues for linux

ptheywood:
Also switches references to the vis repo to the new url of FLAMEGPU/FLAMEGPU2-visualiser

`-DVISUALISATION_ROOT` can now also be a relative path


Robadob:
You've made some mistakes inside `flamegpu2-visualiser.cmake`, some things are treating the visualiser target as `flamegpu2_visualser`, some as `flamegpu2-visualiser`.

I had to make these changes to build it locally, but all the other targets use `_` still so might be worth fixing  `flamegpu2-visualiser.cmake` instead.

image

Robadob:
As discussed in Slack, the problem stems from you changing the FetchContent target to `flamegpu2-visualiser`, and assuming that causes the variables `flamegpu2_visualiser_BINARY_DIR`, `flamegpu2_visualiser_SOURCE_DIR` to change too, but it appears they are based off the actual vis projects cmake target instead.

Robadob (changes requested):
See comments, doesn't currently build.

Robadob (approved):
Haven't tested, but I think you have and it looks right.

ptheywood:
Tested on windows and linux and is currently working (for me)

PR #477, opened 2021-03-22
Add custom lint exclusion filters to calls to add_flamegpu_executable

ptheywood:
This allowes individual examples to filter out any third party code, using regular expressions

e.g.     `LINT_EXCLUDE_FILTERS .*/thirdparty/.*`

will exclude any files with thirdparty as a directory in the path

### DIFF
--- a/cmake/common.cmake
+++ b/cmake/common.cmake
@@ -270,8 +270,18 @@ endif()
 find_file(CPPLINT NAMES cpplint cpplint.exe)
 if(CPPLINT)
     function(new_linter_target NAME SRC)
+        cmake_parse_arguments(
+            NEW_LINTER_TARGET
+            ""
+            ""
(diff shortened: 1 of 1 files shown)


cmake/common.cmake
```diff
@@ -270,8 +270,21 @@ endif()
 find_file(CPPLINT NAMES cpplint cpplint.exe)
 if(CPPLINT)
     function(new_linter_target NAME SRC)
+        set(options)
```
Robadob, line 273: Are these 3 lines not redundant? Left over copy-paste from the docs for `cmake_parse_arguments()`?

Robadob (approved):
Assuming you have checked this doesn't break when the variable isn't passed. Should be ok.

Might want to remove the 3 lines I highlighted they appear redundant.

PR #479, opened 2021-03-23
Embed an icon for the executable.

Robadob:
Only works on Windows, Linux requires a .desktop file in a system directory (so we can't do it without using CMake install).

Also displays in vis, but will override this at some point.


Robadob:
image
image
image
image

ptheywood (approved):

PR #484, opened 2021-03-26
Add workaround for incremental build on MSVC issue

### LINKED ISSUE #483: Windows incremental build failure after only flamegpu2 project modified
We've had this issue for a long time. Rebuild an example, resave any `.cu` file in the main flamegpu2 library and then incrementally build the same example. It fails with a link error as `-dlink` has not been called for the new flamegpu2 library changes.

Pre CUDA 11.2, the only solution is to resave a file in the example and incrementally build. This tricks `-dlink` into occuring.

CUDA 11.2 It becomes possible to set the `CUDA Linker->Additional Dependencies` to point to the path of `flamegpu2.lib`. However it ii not currently possible to target this setting with CMake. ([CMake issue](https://gitlab.kitware.com/cmake/cmake/-/issues/21992)).

I've discussed this issue through with an NV QA team member, and it appears to be an issue with NVCC, as the `CUDA Linker->Use Host Library Dependencies` option is enabled, but host dependencies are not being detected for changes. I'm yet to receive confirmation that will be/has been fixed.

At this time it's not something we can automatically/permanently fix, but we have workaround to save time, so I will keep this updated as more progress is made.

ptheywood:
+ Specifies the absolute path to the static library as an additional link target, which bypasses the nvcc/msvc bug
+ Warnings are emitted by cmake in msvc/nvcc combinations which may/will have issues.
+ Adds some notes to the readme.

closes #483

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,10 @@ Only documentation can be built without the required dependencies (however Doxyg
 + Windows/MSVC and CUDA < 11.0
   + There may be intermittent compilation errors due to an NVCC+MSVC bug exposed by Thrust/CUB. 
   + Re-running the build appears to work in most cases, Otherwise consider upgrading to CUDA 11.0+ if possible.
++ Windows/MSVC and CUDA <= 11.0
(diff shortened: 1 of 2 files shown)


Robadob (approved):
Haven't tested but this looks great.

PR #489, opened 2021-04-01
Proposed fix for #488

### LINKED ISSUE #488: Messagelist size insufficient when two agents write to same list.
**Describe the bug**
Message list is not resized big enough when two agents output to same message list (in the same layer?).
[This](https://github.com/FLAMEGPU/FLAMEGPU2/blob/034d2ac7b5906202612f01eb0f76cf341def5f6e/src/flamegpu/gpu/CUDAMessage.cu#L288) assert is triggered. message_count = 512, newMsgCount = 512, max_list_size = 711.

**To Reproduce**
Steps to reproduce the behaviour:
1. Define a model where two agent functions output to the same (Spatial3D) message list in the same layer
2. Run model

**Expected behaviour**
a) Throw a proper exception at model definition time saying two agents can't output to same message list in same layer
~b) Model work as intended as message list is resized correctly.~

Robadob:
Haven't tested yet, need to reset CMake cache, after i upgraded visual studio the other day.
Should probably also add some additional test cases for the newly added exceptions (I improved a few related ones too).

Closes #488

### DIFF
--- a/include/flamegpu/model/LayerDescription.h
+++ b/include/flamegpu/model/LayerDescription.h
@@ -64,6 +64,7 @@ class LayerDescription {
      * @throw InvalidAgentFunc If the agent function does not exist within the model hierarchy
      * @throw InvalidAgentFunc If the agent function has already been added to the layer
      * @throw InvalidLayerMember If the layer already contains a SubModel
+     * @throw InvalidLayerMember If the agent function outputs to a message list output to by an existing agent function of the layer
      * @note The agent function must first be added to an Agent
      * @see AgentDescription::newFunction(const std::string &, AgentFunction)
      */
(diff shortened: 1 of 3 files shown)


Robadob:
Have tested that existing C and new tests all pass (Windows, Debug), didn't feel it worth checking py tests.

@MILeach In reviewing this, can you acknowledge that you will prevent the same bug from occurring via dependency layer generation.

Thanks

Robadob:
Need to do more work on this, @ptheywood has pointed out shouldn't allow message out in same layer as message in of same type.

Robadob:
Pete has listed some other cases which I don't think are enforced in #488

ptheywood (approved):

PR #494, opened 2021-04-07
Define NOMINMAX to stop windows.h defining min and max macros

ptheywood:
This should improve intellisense, and reduce pollution of the global namespace

Needs testing under windows, in case any dependencies rely on these being defined on windows (hopefully not).
This has no impact under linux.


Robadob:
(Context: The `HostAgentAPI::min`, `HostAgentAPI:;max` reductions were marked as not existing until this define is passed, similarly the `windows.h` defined min/max macros have been a general nuisance requiring calls to min/max to be prepended with `std::` or `glm::`.(

PR #496, opened 2021-04-07
Improve fgpu2_inc_dir detection to work for standalone models

### LINKED ISSUE #467: TemplateExample: Can't find default RTC include path.
[Template example repo](https://github.com/FLAMEGPU/FLAMEGPU2_example_template) was setup before RTC was properly integrated, so doesn't auto detect RTC required include path in the example repo folder layout.

This existing function needs to be updated: https://github.com/FLAMEGPU/FLAMEGPU2_dev/blob/d23833392c0262e11253cf48f59aec1df2145276/src/flamegpu/util/JitifyCache.cu#L93

A second check needs to be added for something like `_deps/flamegpu2-src/include/flamegpu/version.h`. This assumes that the working directory is the build directory (as by default in visual studio), and checking up 5? levels should find it if an executable is called from the binary directory.

Linux users may have other requirements?
Python users may have other requirements?

Worth noting any limitations in the readme of the repo too.

@MILeach / @ptheywood first found this last week.
@jcbpyle also had the same problem today, unclear if he's working on a fix.

Robadob:
Closes #467

Attempted to lint it, but did this in a standalone examples `_deps` folder, so got alot of garbage hiding any real errors.

### DIFF
--- a/src/flamegpu/util/JitifyCache.cu
+++ b/src/flamegpu/util/JitifyCache.cu
@@ -80,7 +80,6 @@ std::string loadFile(const path &filepath) {
 };  // namespace
 
 std::mutex JitifyCache::instance_mutex;
-
 std::unique_ptr<KernelInstantiation> JitifyCache::compileKernel(const std::string &func_name, const std::vector<std::string> &template_args, const std::string &kernel_src, const std::string &dynamic_header) {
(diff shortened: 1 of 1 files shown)


ptheywood (approved):
Fixed for linux, by changing the initial dir to `"."` from `""` and added try-catch blocks to eat any file permission related excpetions.

PR #502, opened 2021-04-08
CI: Add Ubuntu 2004 and remove 1604

ptheywood:
Ubuntu 1604 is EOL soon, and will be removed from GHA at some point.
Ubuntu 2004 is no longer in beta
    
Ubuntu CI compilation times have increased as we've expanded it to include pyflamegpu and more tests.
Reducing the number of cuda architectures being built from 2 (35,75) to 1, being the minimum non-deprecated arch for the cuda version.
    
CUDA 11.0 + deprecates SM < 52

### DIFF
--- a/.github/workflows/Docs.yml
+++ b/.github/workflows/Docs.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   doxygen:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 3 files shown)


ptheywood:
Swithcin to a single cuda architecture has improved total CI time by ~ 50% (45 ~> 30) and brings it in line with windows ci  (although linux also builds the test suite)

PR #506, opened 2021-04-09
Game of Life: Replace console printing output with visualisation.

### LINKED ISSUE #505: Game of Life Example Visualisation
Similar to that demonstrated inside #322 for the discrete SugarScape example, the existing Game of Life example should be updated to provide a proper visualisation using coloured cubes. It was originally tested with a command line logger.

Robadob:
Short video of it here with 1000x1000 grid (not all on screen).
https://www.youtube.com/watch?v=innIvVna3ZE

There are some vertical tear lines in the video, these are not an encoding error, I saw them in the live visualisation. This points to a visualisation issue, but buffer updates should be blocked by mutex, so it's not clear how they would be occurring. Visible that the vis FPS was running below max framerate though.

Closes #505

### DIFF
--- a/examples/game_of_life/src/main.cu
+++ b/examples/game_of_life/src/main.cu
@@ -1,10 +1,8 @@
 #include "flamegpu/flame_api.h"
 #include "flamegpu/util/nvtx.h"
 
-void printPopulation(AgentVector &pop);
-
 FLAMEGPU_AGENT_FUNCTION(output, MsgNone, MsgArray2D) {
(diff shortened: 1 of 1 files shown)


ptheywood (approved):
No vertical tearing for me under linux with a 1070. 

Runs at 2.5sps currently which is a known bug. 
Framrate is at 120fps with limited sps (my vsync). At unlimited sps it drops to 100fps on my system at the defualt resolution, and down to 60fps at 3440x1440 so perf is fine.

PR #507, opened 2021-04-09
Visualise Scale

Robadob:
* Correct issue with previous vis change, that caused model step rate to be wrong.
* Add better checks for correctness to all `AgentVis` texture buffer binds.
* Add support for specifying an agent's unique scale, either as a uniform value, or per axis of the original model.
* Also fixed a bug in `DirectionFunction` that I noticed by chance. Specifying up vector would have generated a shader with bad syntax.

Happy for these commits to be merged without a squash, they should all be distinct.

Paired with vis PR: https://github.com/FLAMEGPU/FLAMEGPU2-visualiser/pull/57


Robadob:
Couple of test images showing uniform/non uniform scaling.

image
image

PR #508, opened 2021-04-10
Reduce header includes

Robadob:
Work in progress, requires testing that I haven't broken RTC.

Probably leave this as a draft for a long while, it's mostly an experiment.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,9 @@
 *.json
 *.bin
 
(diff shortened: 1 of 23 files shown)


Robadob:
Next step is to try and break out the nested class `EnvironmentDescription::PropData`, as this causes `EnvironmentDescription.h` to be included in alot of places.

PR #518, opened 2021-04-16
Update CI to Cuda 11.3

ptheywood:
+ CUDA 11.3 on CI. 
+ Scripts moved to `.github/scripts` rather than `scripts/actions`, as it is the only place these scripts are intended to be ran
+ Moves github actions setting of environment variables into the scripts, which are only emmitted if the `GITHUB_ACTIONS` env var is set.
+ Windows CI includes `CUDA_ARCH` in the `matrix` so it can be set to an appropriate version for the installed CUDA verison.
+ Also removes a cmake dev-warning on MSVC about possible future compiler issues that would trip CI for each new CUDA  version


Robadob (approved):
This should probably be squashed.

PR #536, opened 2021-05-17
Fix Doxygen Warnings

Robadob:
All fixed, now the question is whether we can error on warning (and increase the warning level?, e.g. it don't think it currently warns about missing params if you haven't specified any params, or outright missing doc comments).

## Todo

There appear to be config options for warnings, all begin `WARN_`: https://www.doxygen.nl/manual/config.html
`QUIET` may remove all non warnings output from stdout.
`WARN_NO_PARAMDOC` is disabled by default
`FAIL_ON_WARNINGS` will also cause a CI failure on warning.


ptheywood:
> All fixed, now the question is whether we can error on warning (and increase the warning level?).

[`WARN_AS_ERROR`](https://www.doxygen.nl/manual/config.html#cfg_warn_as_error) exists as a doxygen setting, so if our relevant CMake variable `WARNINGS_AS_ERRORS` is set to `ON` then  we can add that to the doxygen config. 

It's not obvioulsy supported by CMake's doxygen integration that I think we use, although it seems to suggest that if it starts with `DOXYGEN_` it will  make it into the config file. 

> Any variable with a name of the form DOXYGEN_<tag> will have its value substituted for the corresponding <tag> configuration option in the Doxyfile. See the Doxygen documentation for the full list of supported configuration options.

[source](https://cmake.org/cmake/help/latest/module/FindDoxygen.html#command:doxygen_add_docs)

Robadob:
CMake doxygen integration supports all options, you just declare the options as variables, which it consumes.

`FAIL_ON_WARNINGS` is better than `WARN_AS_ERROR`, the latter fails on first warning rather than printing all warnings first.

Robadob:
As expected `QUIET` doesn't suppress warnings, so have enabled that too.

Haven't actually turned on WERROR, to check that it does fail. But should handle CI right assuming the action enables the WERROR setting.

Robadob:
Well that didn't work -_-

Robadob:
There was an issue, but I fixed that. So now unsure why it's not failing. I'll play with it more tomorrow.

Robadob:
Found it, forgot to append `DOXYGEN_` to most the doxy variables.

ptheywood:
```
make docs 
[100%] Generate API documentation for docs
/home/ptheywood/code/flamegpu/FLAMEGPU2/include/flamegpu/io/xmlReader.h:4: warning: the name 'xmlreader.h' supplied as the argument in the \file statement is not an input file
/home/ptheywood/code/flamegpu/FLAMEGPU2/include/flamegpu/io/xmlWriter.h:4: warning: the name 'xmlwriter.h' supplied as the argument in the \file statement is not an input file
/home/ptheywood/code/flamegpu/FLAMEGPU2/include/flamegpu/runtime/messaging/BruteForce/BruteForceDevice.h:182: warning: Found unknown command '\padsram'
/home/ptheywood/code/flamegpu/FLAMEGPU2/include/flamegpu/runtime/messaging/BruteForce/BruteForceDevice.h:189: warning: The following parameter of MsgBruteForce::Out::setVariable(const char(&) variable_name[N], T value) const is not documented:
  parameter 'value'
[100%] Built target docs
[18:22:47 ptheywood@porter] ~/code/flamegpu/FLAMEGPU2/build doxygen_warnings → 
echo $?
0
```
It's still returning a 0 error code currently, so CI doesn't get angry. 

Looks like Cmake generates `build/FLAMEGPU2/Doxyfile.docs`, which for me locally has the following as the `WARN_` content:

```
WARNINGS               = YES
WARN_IF_UNDOCUMENTED   = YES
WARN_IF_DOC_ERROR      = YES
WARN_NO_PARAMDOC       = NO
WARN_AS_ERROR          = NO
WARN_FORMAT            = "$file:$line: $text"
WARN_LOGFILE           = 
```

Looks like you've just missed the `DOXYGEN_` prefix from the cmake variables. (Will check push a fix if it works)

Robadob:
Also now noticed that it's not `FAIL_ON_WARN`, you have to set `WARN_AS_ERROR` to that rather than `ON` or `OFF`

Robadob:
Have managed to get it to fail locally.

ptheywood:
It's also that `FAIL_ON_WARNINGS` is not a doxygen config argument, it is a value for `WARN_AS_ERROR` (other than `NO` or `YES`). 

> Possible values are: NO, YES and FAIL_ON_WARNINGS.

Howver, CMake does not like it being used that way emitting the following: 

```
warning: argument 'FAIL_ON_WARNINGS' for option WARN_AS_ERROR is not a valid boolean value
```

`set(DOXYGEN_WARN_AS_ERROR YES)` however does atleast fail the build, although it is immediate rather than containing multiple errors.

Robadob:
Despite the CMake warning, it works. So i guess need to specify as string

Robadob:
How the fuck has that passed, it failed locally

Robadob:
The file CMake output on config has the right value
image

ptheywood:
`WARN_AS_ERROR` was added in doxygen 1.9.0, ubuntu 20.04 ships with 1.18.17 (because 1.19.0 was released 27-12-2020)

https://www.doxygen.nl/manual/changelog.html

Robadob:
RIP

Robadob:
May explain why I don't get that CMake warning either. Assuming CMake sets all those variables to their defaults as strings. So parses setVar as string.

Robadob:
I''l try making CI build doxygen from src tomorrow, cba tonight.

Robadob:
Or that, GJ.

I'll squash and cleanup the commits tomorrow (I want to keep the doxygen changes, separate to the doc comment changes).

ptheywood:
CI now builds doxygen 1.9.1 so we get the better error behaviour. I'll let you sort the history how you want it / fix the errors / warnings.

Robadob:
I think this commit separation is good, final two could potentially be merged.

Robadob:
Guess those last 2 warnings were filename case sensitive on Linux only.

ptheywood (approved):

PR #539, opened 2021-05-21
Py doxygen

Robadob:
Experimental work trying to make python API docs build without needing to compile the lib.

Cautiously appears to be working. But will obviously need further testing and cleanup.


Robadob:
Appears somehow I've broken regular pyflamegpu target, so will need to investigate that too.

Robadob:
Ok no it's not working.
The file output by swig is empty, so doxygen is outputting nothing either. Just both of these are build successes.

Robadob:
Ok, from checking this.

Failing to provide the `-doxygen` to swig, stops all doc comments being transferred to python in any form (had assumed lack of translation would simply mean comments were copied verbatim).

Earlier reading of SWIG docs had suggested that `%feature("doxygen:notranslate")` would copy the doxygen, without translating it to pydocs format. But on testing this, it was breaking compilation of the swig code and the above discovery suggests it will similarly just block comments from being copied.

This might mean the experiment is a failure, and we need to try 'pydocs'.

Robadob:
Issue with the SWIG docs, is that many functions have their args switched to `* args`, which appears to by Python variadic args format.

Example here:
image
image


Similar example of it not being a problem here:

image


Assumption is that it's related to overloading.

Robadob:
James Knight, UK RSE slack, re:swig docu for overloaded fns

PR #542, opened 2021-05-27
Add AgentDescription::newRTCFunctionFile()

Robadob:
Loads RTC function string from file.

Have checked the new tests pass.

### DIFF
--- a/include/flamegpu/model/AgentDescription.h
+++ b/include/flamegpu/model/AgentDescription.h
@@ -150,7 +150,7 @@ class AgentDescription {
     template<typename AgentFunction>
     AgentFunctionDescription &newFunction(const std::string &function_name, AgentFunction a = AgentFunction());
     /**
-     * Adds a new runtime (device) function to the agent
+     * Adds a new runtime (device) function to the agent from a string containing the source code
      * @param function_name Name of the functions
      * @param func_src representation of an agent function
      * Should be declared using FLAMEGPU_AGENT_FUNCTION notation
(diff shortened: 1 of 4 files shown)


ptheywood (approved):
Looks fine to me, but not tested it myself.  

I wonder if this has any potential impact on rtc debugging (i.e. breakpoints are attached to functions by name? Though lines will be 2 out compared to what is atually compiled anyway)

Robadob:
> Looks fine to me, but not tested it myself.
> 
> I wonder if this has any potential impact on rtc debugging (i.e. breakpoints are attached to functions by name? Though lines will be 2 out compared to what is atually compiled anyway)

RTC debugging depends on whatever filename is told to jitify. So this won't work for that.

Lines should only be 1 out in this format (the header include). They're normally an extra out, because the opening quote is on a line by itself.

ptheywood:
Isn't the other line the `program` name that gets inserted rather than just a blank new line at the start of the string? 
Eitherway, this should make the development experience nicer (although users may need to add the `#include "flamegpu/runtime/DeviceAPI.h"` themselves for autocompletions depending on editor?).

Robadob:
> Isn't the other line the program name that gets inserted rather than just a blank new line at the start of the string?

Mm possibly, though i'm not aware of the input string being trimmed.


> (although users may need to add the #include "flamegpu/runtime/DeviceAPI.h" themselves for autocompletions depending on editor?).

Yes

PR #544, opened 2021-05-31
ViridisInterpolation was missing from Python interface.

Robadob:
2nd commit is a bugfix of something noticed by Carlos.
3rd commit is another bugfix of something noticed by Carlos.

So this PR shouldn't be squashed on merge.

### DIFF
--- a/examples/game_of_life/src/main.cu
+++ b/examples/game_of_life/src/main.cu
@@ -11,8 +11,8 @@ FLAMEGPU_AGENT_FUNCTION(update, MsgArray2D, MsgNone) {
     const unsigned int my_y = FLAMEGPU->getVariable<unsigned int, 2>("pos", 1);
 
     unsigned int living_neighbours = 0;
-    // Iterate 3x3 grid
-    for (auto &msg : FLAMEGPU->message_in(my_x, my_y, 2)) {
+    // Iterate 3x3 Moore neighbourhood (this does no include the central cell)
+    for (auto &msg : FLAMEGPU->message_in(my_x, my_y)) {
         living_neighbours += msg.getVariable<char>("is_alive") ? 1 : 0;
     }
     // Using count, decide and output new value for is_alive
--- a/src/flamegpu/visualiser/color/HSVInterpolation.cpp
+++ b/src/flamegpu/visualiser/color/HSVInterpolation.cpp
@@ -92,9 +92,9 @@ vec3 hsv2rgb(vec3 hsv) {
     // Fetch the modifier from texture cache
     ss << "    float modifier = texelFetch(color_arg, gl_InstanceID).x;" << "\n";
     // Clamp the modifier to bounds
-    ss << "    modifier = clamp(modifier, " << min_bound << ", " << max_bound << ");" << "\n";
+    ss << "    modifier = clamp(modifier, float(" << min_bound << "), float(" << max_bound << "));" << "\n";
     // Scale modifier to range [0.0, 1.0]
-    ss << "    modifier = (modifier - " << min_bound << ") / " << (max_bound - min_bound) << ";" << "\n";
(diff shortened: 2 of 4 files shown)


Robadob:
Appears CI has updated the version of visual studio and now we're getting syntax errors inside system headers.

This issue might be related to CUDA, random other project that uses CUDA, user reporting similar issue after updating visual studio: https://github.com/lightvector/KataGo/issues/490

ptheywood:
> Pete has suggested it might be related to C++14 (which we are using), so updating to C++17 could potentially fix it (although provide incompat to old CUDA versions?)

The following was the vs dev community post i found suggesting it only impacted things when the c++14 flag was passed.

https://developercommunity.visualstudio.com/t/vs-16100-isnt-compatible-with-cuda-110/1433342

I made some quick changes to see if C++ 17 would resolve it, however that fails to build due to some c++ 17 removals (`std::unary_function/binary_function`) and new warnings which are promoted to errors, although it did get past the file which errored first in the c++14 build (`FGPUDeviceException.cu`).
There are still som errors in standard library headers, but at a glance they seem to be due to the removed features.

Might also be worth trying older cuda version(s) just in case, but I'm doubtful.

Robadob:
>  some c++ 17 removals (std::unary_function/binary_function)

These have apparently been deprecated/removed as they are unnecessary. So simply removing any inheritance of them should (apparently) work fine. Only believe we used them to mimic how Thrust was operating custom transform/reduce ops.

ptheywood:
Removing the std:unary/binary fixed a bunch of the warnings (not done very cleanly so far to see if it's viable or not). Looks like there's just a std::iteratior change too that i'd not spotted when skimming the previous error logs. 

GCC (at least the version i'm using) doesn't correctly mark these as being deprecated either (or removing things like `std:binary_function`, although that is an active decision and still technically spec conforming for it to still be available) which is why i'd missed these issues previously. 

> warning STL4015: The std::iterator class template (used as a base class to provide typedefs) is deprecated in C++17. (The <iterator> header is NOT deprecated.) The C++ Standard has never required user-defined iterators to derive from std::iterator. To fix this warning, stop deriving from std::iterator and start providing publicly accessible typedefs named iterator_category, value_type, difference_type, pointer, and reference. Note that value_type is required to be non-const, even for constant iterators. You can define _SILENCE_CXX17_ITERATOR_BASE_CLASS_DEPRECATION_WARNING or _SILENCE_ALL_CXX17_DEPRECATION_WARNINGS to acknowledge that you have received this warning.

Robadob:
Yeah that iterator thing seems like a similar case. I never properly understood all the typedefs (and they weren't documented greatly either). So no harm in removing those too, although would want to check all tests are happy.

ptheywood:
Each of the template parameters from the base std::iteartor class map nicely onto each typedef, so it seems like its been a simple fix. 

Not done it very tidily, nor checked its not broken anything but it still builds flamegpu2 on linux and set CI going again. If it does work on CI then i'll put some time in to make sure things still work and make it selectively build as c++14 or c++17 depending on support. 

I guess we could build as 17 if available (nvcc > 11.x and host compiler support), otherwise build as 14, and we will just need to make sure that we have atleast once CI build on each (current linux CUDA 10.x, CUDA 11.x and windows CUDA 11.y should provide this. Though we can't have a windows c++14 build as that is the root cause which is requiring these fixes)

Robadob:
That all sounds good.

ptheywood:
flamegpu2 target builds, however pyflamegpu fails which is likely to put a complete stop to this plan. Swig only supports upto c++14 ... 

And it looks like the cmake swig integration might set the cxx standard to c++14 itself (`--std=c++14`  is being passed to nvcc and it doesn't like we're setting that anywhere in the swig/python cmake files ourselves)

Robadob:
Mmm, that's a problem then. Guess we might need to ignore CI failures and wait for a patch :|

ptheywood (approved):

PR #549, opened 2021-06-03
Array Message Improvements

### LINKED ISSUE #547: MessageArray3D not working as intended
**Describe the bug**
I'm running a simulation with NxNxN agents placed in a regular grid. It works fine except when N = 3

**Expected behaviour**
Every agent should check its 26 neighbours, and it does, provided the grid is (apparently) larger than 3x3x3

**Screenshots**
Output when N = 3 
imagen

Output when N > 3
imagen

**Configuration(please complete the following information):**
 - OS: [Windows 10]
 - CUDA version: [11.3]
 - GPU [GeForce GTX 1650 with Max-Q Design]
 - GPU Driver Version [27.21.14.6627]

### LINKED ISSUE #548: Array Message Wrapping Behaviour
While investigating the issue with 3x3x3 `Array3D` messages (#547) and adding additional tests to catch more edge cases some additional things have been highlighted.

+ A `radius` of `0` is blocked 
    + this will prevent models like `PedestrianNavigation`
+ A `radius` of `1` does not read the self message 
    + this is consistent with F1, but may confuse users comapared to other messaging schemes (spatial)
+ If one or more dimensions are insufficient for the radius, the same message will be read multiple times
    + I.e. for a `2x2x2`, with a `radius >= 1`, or for a `5x5x4` with `radius >= 2`
    + Additionally, in this case the original agents message *will*  be read (several times, but one time less than other messages)

I assume this applies to `Array` and `Array2D` messages as well but have not yet checked.

We should decide:
1. Should `radius == 0` be supported, returning just the message for the specific location.
    + **Answer**: No - `.at()` provides direct access to a single message.
2. Should the self message be returned in normal conditions for consistency with other messaging types (spatial)?
3. Shopuld configuraitons which lead to double reading due to wrapping be allowed?
    + If so, should messages be read multiple times? 
    + If so, and depending on the answer to Q2, should the self message be returned when wrapped?
    + How should `1x1x1` be dealt with. Blocked? 
         + This may make flexible code more difficult (as method prototypes change if message type is changed).

PR #549 will eventually address these issues once the solution is decided.

### LINKED ISSUE #552: [FeatureReq] MsgArray/2D/3D Moore iteration sans wrapping
**Is your feature request related to a problem? Please describe.**
Moore iteration of array messages wraps the array, so accesses at boundaries get messages from far boundaries.

**Describe the solution you'd like**
An alternative iteration scheme which does not wrap.
We should also carefully consider which should be the default behaviour, if not wrapping, then that change should be made sooner rather than later.

ptheywood:
Array messages are currently broken for combinations of dimensions and communication radii.

Closes #547, closes #548 and closes #552.

+ [x] Add new tests to detect errors
+ [x] Adjust old tests that `+1` to the number of read messages to expect the correct value
+ [x] Where the bounds are exactly correct for the communication radius (#547)
+ [x] Where the bounds are less than required for the communication radius, leading to reading of more messages than intended (#548)
+ [x] Adds wrapped array messaging (#552)

### DIFF
--- a/examples/game_of_life/src/main.cu
+++ b/examples/game_of_life/src/main.cu
@@ -11,7 +11,7 @@ FLAMEGPU_AGENT_FUNCTION(update, flamegpu::MsgArray2D, flamegpu::MsgNone) {
 
     unsigned int living_neighbours = 0;
     // Iterate 3x3 Moore neighbourhood (this does no include the central cell)
-    for (auto &msg : FLAMEGPU->message_in(my_x, my_y)) {
+    for (auto &msg : FLAMEGPU->message_in.wrap(my_x, my_y)) {
         living_neighbours += msg.getVariable<char>("is_alive") ? 1 : 0;
     }
     // Using count, decide and output new value for is_alive
--- a/examples/sugarscape/src/main.cu
+++ b/examples/sugarscape/src/main.cu
@@ -90,7 +90,7 @@ FLAMEGPU_AGENT_FUNCTION(movement_request, flamegpu::MsgArray2D, flamegpu::MsgArr
 
     // if occupied then look for empty cells
     if (status == AGENT_STATUS_MOVEMENT_UNRESOLVED) {
-        for (auto current_message : FLAMEGPU->message_in(agent_x, agent_y)) {
+        for (auto current_message : FLAMEGPU->message_in.wrap(agent_x, agent_y)) {
             // if location is unoccupied then check for empty locations
             if (current_message.getVariable<int>("status") == AGENT_STATUS_UNOCCUPIED) {
                 // if the sugar level at current location is better than currently stored then update
@@ -131,7 +131,7 @@ FLAMEGPU_AGENT_FUNCTION(movement_response, flamegpu::MsgArray2D, flamegpu::MsgAr
     const unsigned int agent_x = FLAMEGPU->getVariable<unsigned int, 2>("pos", 0);
     const unsigned int agent_y = FLAMEGPU->getVariable<unsigned int, 2>("pos", 1);
 
-    for (auto current_message : FLAMEGPU->message_in(agent_x, agent_y)) {
+    for (auto current_message : FLAMEGPU->message_in.wrap(agent_x, agent_y)) {
         // if the location is unoccupied then check for agents requesting to move here
         if (status == AGENT_STATUS_UNOCCUPIED) {
             // check if request is to move to this location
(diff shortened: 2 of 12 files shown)


Robadob:
Have investigated `TestMessage_Array2D.MooreX3Y3R2`.

The issue comes down to how iteration works.

```
abs == (src + rltv + dims)%dims
```

Begin Iterator
```
(1, 1) == ((0, 0,) + (-2, -2) + (3, 3)) % (3, 3)
```
End Iterator
```
(0, 1) == ((0, 0,) + (3, -2) + (3, 3)) % (3, 3)
```
One thread's iteration
```
(1, 2) == ((0, 0,) + (-2, -1) + (3, 3)) % (3, 3)
(1, 0) == ((0, 0,) + (-2, 0) + (3, 3)) % (3, 3)
(1, 1) == ((0, 0,) + (-2, 1) + (3, 3)) % (3, 3)
(1, 2) == ((0, 0,) + (-2, 2) + (3, 3)) % (3, 3)
(2, 1) == ((0, 0,) + (-1, -2) + (3, 3)) % (3, 3)
(2, 2) == ((0, 0,) + (-1, -1) + (3, 3)) % (3, 3)
(2, 0) == ((0, 0,) + (-1, 0) + (3, 3)) % (3, 3)
(2, 1) == ((0, 0,) + (-1, 1) + (3, 3)) % (3, 3)
(2, 2) == ((0, 0,) + (-1, 2) + (3, 3)) % (3, 3)
(0, 1) == ((0, 0,) + (0, -2) + (3, 3)) % (3, 3)
```

The iteration isn't designed for double message accesses, so it compares the absolute message pointed to, rather than the relative position of the iterator

```C++
class MsgArray2D::In {
    // ...
    class Filter {
       // ...
        class iterator {
            // ...
            /**
             * Equality operator
             * Compares message
             */
            __device__ bool operator==(const iterator& rhs) const { return  _message == rhs._message; }
            /**
             * Inequality operator
             * Compares message
             */
            __device__ bool operator!=(const iterator& rhs) const { return  _message != rhs._message; }
        };
    };
};
```

Presumed fix would be to change these to instead compare relative position.

Robadob:
Thoughts:

Add alternative not-wrapped iterator (will need to be careful of how end iterator behaves at final corner), which has no restriction on size.
Update existing wrapped iterator, to not be the default iterator, and to throw an error for dim/rad combinations that would lead to double message reads.
All this should be well documented in docu comments and userguide.

ptheywood:
This should be ready to go once CI has ran IMO. 

The array message tests now all pass (C++ and python)  and are in a decent state now. Will open an issue for partially wrapped array messages as a futre feature (if required).

Squashed merge will be best, just keeping the commits separate for ease of review.

Robadob (approved):
Looks good.

This should positively affect Carlos (as it was his original bug that raised it), will need to let him know.

Issue #550, opened 2021-06-03
Class / Namespaces to Rename / Move


ptheywood:
Making an issue to keep track of class / namespaces / files I come across during namespacing that I feel should be renamed. 

This list will evolve over time, and at some point I'll bring things up for dicussion.

+ [x] `FGPU_SA` / `exception/FGPUStaticAssert.h` -> `flamegpu::util::StaticAssert`
    + `static_assert` is a reserved word, so might still want renaming?
+ Move additional things into detail:
    + [x] `util::StaticAssert` - needed in headers for template reasons, but not intended to be user-facing.
    + [x] `curve/cuRVE` - and/or probably also needs its own namespace, so `flamegpu::curve::Curve` and `flamegpu::curve::detail`, or `flamegpu::detail::curve`, although it also has some namespaced internal stuff, though it should be safe to place that in flamgepu::detail::curve` in general.
     + [ ] `model/*Data` - this is not inded for users to use, but they kind of have to? Unsure.
     + [x] `util` A significant amount of util is not intended to be user-facing, could move to `flamegpu::detail::util` / `flamegpu::util::detail`. NVTX is semi-user facing however.
+ [ ] Additional sub namespaces
    + [x] `flamegpu::io` - the IO directory makes sense as a group of things. It will grow over time as more formats are supported, so  using an io namespace may make sense (and match disk)
+ [x] `FGPUException`
    + Rename to `FLAMEGPUEException`? or just `flamegpu::Exception`? Not sure how this will translate to python, or if using the generic name `Exception` is a good idea, even inside a namespace. 
+ [x] `flamegpu_internal` -> replace with `flamegpu::detail`
+ [x]  `flamegpu/runtime/flamegpu_api.h` only includes 3 other header filles. It is superflous. Users should include `flamegpu/flamegpu.h`, as should any tests.
+ [ ] Should `include/flamegpu/runtime/messaging(_device).h` actually be in `include/flamegpu/runtime/messaging/`?
+ [x] `io/*` - CapitolCase for class/filenames?
+ [x] `curve_internal` -> `curve::detail`
+ [ ] `ModelVis:model` is actually a reference to a `CUDASimulation`. `ModelVis::simulatiion` or `SimulationVis`? 
+ [ ] `Simualtion`/`CUDASimulation`. Simulation or Simulator? Doesnt' really matter grand scheme.
[...]

Issue #78, opened 2018-11-02
Name headers .cuh and .h as appropriate


ptheywood:
Cmake uses the appropriate compiler based on the extension of teh file. i.e `.cpp` files will use `g++` while `.cu` files will use `nvcc`. 
Visual studio with cuda enabled will pass all files through `nvcc`. 

This means that including cuda related headers in a `.cpp` file will work under windows but not linux. 

I.e. 

```
In file included from /home/ptheywood/code/flamegpu/flamegpu2_dev/include/flamegpu/runtime/flame_functions_api.h:14:0,
                 from /home/ptheywood/code/flamegpu/flamegpu2_dev/include/flamegpu/model/AgentFunctionDescription.h:22,
                 from /home/ptheywood/code/flamegpu/flamegpu2_dev/include/flamegpu/model/AgentDescription.h:22,
                 from /home/ptheywood/code/flamegpu/flamegpu2_dev/include/flamegpu/model/ModelDescription.h:22,
                 from /home/ptheywood/code/flamegpu/flamegpu2_dev/include/flamegpu/io/statereader.h:15,
                 from /home/ptheywood/code/flamegpu/flamegpu2_dev/include/flamegpu/io/xmlReader.h:14,
                 from /home/ptheywood/code/flamegpu/flamegpu2_dev/src/flamegpu/io/xmlReader.cpp:16:
/home/ptheywood/code/flamegpu/flamegpu2_dev/include/flamegpu/gpu/CUDAErrorChecking.h:16:26: fatal error: cuda_runtime.h: No such file or directory
```

The file `xmlReader.cpp` is compiled by g++, but tthrough the inclusion chain it attempts to include `cuda_runtime.h` which is not available. 

Using the appropriate naming convention should help us to not include cuda files in cpp files.

ptheywood:
Renaming `flame_functions_api.h` to `flame_functions_api.cuh` would let cmake correctly handle it as a cuda file and use `nvcc` rather than a host c compiler, with appropriate flags to actually find `cuda_runtime.h`

PR #551, opened 2021-06-03
Namespaces

### LINKED ISSUE #531: Separate config specific directory for python builds
The swig/python output files are not config aware, i.e. a Release and Debug build in the same build directory will overwrite one another.

It should probably go into `lib/<os>/<config>/python/`. 

`<os>` is not  strictly required, as we do not track the build directory and do not support cross-compilation, it is mostly just a  FLAME GPU 1 legacy thing. We could remove it but it also isn't causing  any harm. 

Currently, the `build/lib` directory on linux after a Release build and a reconfigure to a debug build is looks as follows (directories, 4 layers deep):

```
tree -d -L 4 lib/
lib/
└── linux-x64
    ├── Debug
    ├── python
    │   ├── build
    │   │   ├── bdist.linux-x86_64
    │   │   └── lib
    │   ├── dist
    │   ├── pyflamegpu
    │   ├── pyflamegpu.egg-info
    │   └── venv
    │       ├── bin
    │       ├── include
    │       ├── lib
    │       ├── lib64 -> lib
    │       └── share
    └── Release
```

This can probably just be addressed when the swig cmake is adapted to create several swig/python modules to support namespacing. 

It might be worth considering moving the swig/python bindings out of `lib`, as they don't really belong there (i.e. it's not a `.so`/`.lib` file). Perhaps `build/python/` or `build/pyflamegpu`? (although need to be careful with directories that swig will already create due to target names / subdirectories that have been added, i.e. `build/swig/python` already exists.

### LINKED ISSUE #568: Missing threadsync in the agent function wrapper
If `SEATBELTS` is enabled thread 0 writes to shared memory, but there is no sync to ensure that this write has completed prior to other threads (potentially) writing to it.

As this is only ever used in exceptional casses, it is probably unlikely to ever be triggered as a race.

https://github.com/FLAMEGPU/FLAMEGPU2/blob/eb03a165c93c78fd1c5309b69181edb8c7296efe/include/flamegpu/runtime/AgentFunction.h#L75-L81

### LINKED ISSUE #578: Unused Exception Types
Several Exception types are currently defined but never thrown. 

These should either be removed, or made use of?

+ `InvalidHashList`
+ `InvalidMapEntry`
+ `InvalidMessageSize`
+ `InvalidPopulationData`

ptheywood:
Introduce C++ namespaces to the library, to avoid polluting the global namespace.

This is not as namespacey as I would like, however due to time / complexity it will do for now. 

+ [x] Everything goes into `flamegpu` namespace
+ [x] Adds several sub-namespaces 
    + `exception` - mostly just to clean up the docs
    + `io` - nicely self-contained.
    + `util` - quite self contained
    + `visualiser - distinct from the main repo, although there is some mixing.
    + `detail` namespaces - parts of headers that users ideally shouldn't know exists
        + Ideallt more would be in detail, but due to mutual friendship + how protected etc interact with namespaces this is more of a problem than desired.
    + Did not split out the host/device components of the API, too interlinked.
        + ~Flatten the file hierarchy for `gpu`, `pop`, `runtime`, `sim`? - 70+ headers currently...~
            + A number of these should be in detail, but moving them to detail so far has generally broken things. 
+ [x] Full rewrite of the swig .i file to achieve working namespaces. 
    + Now more well-structured based on the order in which different swig features need using.
    + Care needs to be taken when wrapping new classes, to ensure that swig is aware of the type prior to it being used. Forward declarations in included header files may need to be manually defined in some cases.
+ [x] Renames many things, part of #550
    + Not complete but decent amount done.
+ [x] Rename headers which contain CUDA symbols `.cuh`.
    + Doesn't follow the chain of .cuh includes in headers fully, so only part of #78
+ [x] Move some files around to more appropraite locations
+ [x] Expand util with additional typedefs used as keys in maps (`std::pair<X,Y>` mostly)
+ New tests
    + [x] Check that RTC Message type checkign correctly matches when the rtc string includes `namespace flamegpu {`, `using namespace flamegpu`, or `using flamegpu::MsgNone`  etc)
    + [x] Add tests to check that regular agent functions build using the various approaches to namespaces (explicit, using directive (maybe, cpplint will hate it), using declaration, alias, alias + explcicit)
+ [x]  Adds utiltiy method to get the unqualified class name from a (potentially) qualified class name, for error human friendly errors.
+ [x] Remove `os-platform` from `lib`/`bin` directory paths. It was unneccesary.
+ [x] Config specific python lib directories (closes #531)
+ [x] Disables some rogue printf's in DependencyGraph.
+ [x] Misc CMake improvements
    + More to be done.
+ [x] Block some namespaces from appearing in doxygen. 
+ [x] CMake option `BUILD_SWIG_PYTHON_VIRTUALENV` -> `BUILD_SWIG_PYTHON_VENV`
+ [x] Add missing `__threadsync()` to `AgentFunctionWrapper` when `SEATBELTS` is on. Closes #568.
+ [x] Remove unused exceptions. Closes #578.
+ [x] Adds python test checking ID_t works correctly (in test_cuda_simulation.py) 
+ [x] RTC Compile Time improvment(s), part of #402.
    + Only include the required messaging types for each agent function. C++ test suite time 500s -> 355s. 

Todo pre-merge:
+ [x] Merge vis repo PR once re-tested

Todo pre/post-merge:
+ [x] Open issues:
   + [x] Extra namespacing / detail namespacing
   + [x] Developer docs CMake option / target which includes things which are not part of the public API.
   + [x] Message class names, i.e. `MsgNone` does not match `messaging/None.h` / `messaging/None/NoneHost.h`. Couldn't move this to `messaging` namespace because of mutual friendship (probably fixable, with effort / time, might cause issues with swig too, unsure). 
    + [x] Update docs.

+ [ ] Update each standalone repo to use the namespaces


ptheywood:
With namespaces, the generated `swig_types[]` deifnes have the full flattened namespace in the name. If they have been encountered before that type is wrapped, it isn't prefixed by the appropraite namespaces, leading to something that builds but the generated python is incorrect.

ptheywood:
Right now, a single pytest is failing in `tests/swig/python/runtime/messaging/test_brute_force.py`. Unclear why it is only one, need to investigate. 

Building with vis enabled is currently generating some wonky types. Looks like either the vis swigging needs to occur earlier, or if vis is enabled some additional forward declarations might be required (possibly just the one of ModelVis, which is available via CUDASimulation?)

After that, python should work and actual work on additional namespacing / renaming things / sorting header file types / moving files around can resume.

ptheywood:
Oddly all tests are currently passing for VIS=OFF builds, unsure why the brute force test failiure is no longer occuring..

ptheywood:
Once CI passes this should be in a mergable state. History is made up of meaningful isolated commits so i'd prefer a rebase / ff merge than a squash. 
It needs to be done at the same time as the matching PR on the vis repo, and then I'll update all the standalone repos. 

I've opened new issues to track the other outstanding changes which didn't make it in.

PR #566, opened 2021-06-21
Concurrency/cuda stream Improvements

ptheywood:
Additional use of concurrency / improvments implemented during the cineca hackathon

+ [x] Async/streamed id generation related functions (for birth)
+ [ ]  Message output post processing
+ [ ] Agent birth post processing
+ [ ] Agent death post processing
+ [ ] Batched memcpy
+ [ ] Pinned memory.
+ [ ] Stream ownership / access.


ptheywood:
A decent chunk of the todo's in this have been covered by @Robadob since this was made, however there are still uses of non-async default stream `cudaMemcpy` in `getDeviceNextID`, so I won't close this yet (but won't resume working on this as a priority either)

PR #567, opened 2021-06-21
BugFix: Not having CUDA_PATH set when building RTC agent functions wo…

### LINKED ISSUE #565: Gracefully handle CUDA_PATH environment variable not being set
Currently if the users's environment does not contain the `CUDA_PATH` environment variable then a segfault will occur. 

This should be handeled gracefully in some way.

https://github.com/FLAMEGPU/FLAMEGPU2/blob/ebf73de9d81d6f42ef9b5229580d06bab655f5de/src/flamegpu/util/JitifyCache.cu#L88

Robadob:
…uld cause a crash.

There was already a check for empty string below.

Closes #565

(Haven't tested this but it's pretty trivial)


ptheywood (approved):
Now produces a sensible exception if `CUDA_PATH` is unset.

PR #588, opened 2021-07-09
Fix use of id_t in boids examples

ptheywood:


Robadob:
Is there even a reason why Boids isn't using proper ID? rather than a manual variable?

ptheywood:
It is, but messages do not have an automatic ID?

Robadob:
Ah, that's my misunderstanding. I only skim read the changelog. Should be fine then.

Robadob (approved):
Python example was already fixed (just looked).

PR #596, opened 2021-07-16
Embedded Version Number (2.0.0-alpha)

### LINKED ISSUE #163: Project Rename / Version numbering
@mondus suggested thinking of a new name rather than `FLAME GPU 2`, which will lead to potentially confusing version numbers (at least until the first full release)

+ `FLAMEGPU`
    + Just call the project `FLAME GPU` as a a major release in the FLAME GPU system - i.e `FLAME GPU 2.0.0`
    + Massive breaking changes / entirely different model to FLAME GPU 1
    + Overwriting the FLAMEGPU/FLAMEGPU repo might lead to issues with people using it in submodules and not targetting specific tags. 
 + `FLAMEGPU2`
    + The 2 is in the name, rather than a version number
    + Pre-releases could be `FLAMEGPU2 0.x.y`, with `FLAMEGPU2 1.0.0` for the first release with a stable interface.
    + Separate repo `FLAMEGPU/FLAMEGPU2` doesn't break other projects
    + Keeps history / clear successor to FLAMEGPU1
    + Would want to update the flamegpu1 readme to point to fgpu2 for all future dev. 
+ Other renames
    + Competely change the name to clearly show its different from FLAME GPU 1. 
    + Might lose the clear history / weight from FLAME GPU
    + Do we want to keep/lose the association with `FLAME` as the parent project? 
    + Semver starting `0.1.0` and moving to `1.0.0` once api is stable would make sense. 
    + Suggestions (not all serious)
        + Pheonix (via mondus)
        + FLAML GPU - it's a **L**ibrary not an **e**nvironment

### LINKED ISSUE #591: Embeded Version Number
We should embed the version number in the shared library and python lib for header checking purposes.

+ [x] Embed the version number as a macro in a header file
    + `FLAMEGPU_VERSION` - a 7+ digit integer endcoding major, minor and patch, if multiple API versions are intended to be supported
+ [x] `static constexpr` for `MAJOR`, `MINOR`, `PATCH`, `PRERELEASE` and `BUILDMETADATA`
    + `BUILDMETADATA` is the old git commit hash. 
+ [x] Embed the namespaced members into the python wheel
+ [x] Factor in the release version number in RTC comparisons as well / isntead of git commit hash?
+ [x] Document how the version numbers work (Semver, -alpha[.N] for very early pre-releases, -rc[.x] for almost stable releases candidates

Ideally this would be defined in a header, and extracted by CMake, but for now encode in the CMake and include in the generated header.

Need to decide how we will mark this during development between releases. I.e. Do we bump the version number for the first commit after a release, or only mark it on the commit which should become the new tag?

ptheywood:
Embeds the FLAME GPU version number in the c++ headers and python module.

+ [x] Define the Major, Minor, Patch and Pre-release version components in CMake 
    + Initially `2.0.0-alpha`, to be followed by `2.0.0-alpha.1`, `2.0.0-beta[.N]`, `2.0.0-pr.N` and then `2.0.0`
    + For now this will be defined in CMake rather than the header file itself, which is currently dynamically generaetd
+ [x] Embed the version number as a macro in a header file
    + `FLAMEGPU_VERSION` - a 7+ digit integer endcoding major, minor and patch, if multiple API versions are intended to be supported
+ [x] `static constexpr` for `MAJOR`, `MINOR`, `PATCH`, `PRERELEASE` and `BUILDMETADATA` in the `flamegpu` namespace
    + `BUILDMETADATA` is the old git commit hash. 
+ [x] Embed the namespaced members into the python wheel
+ [x] Factor in the release version number in RTC comparisons, including the metadata.
+ [x] Document how the version numbers work (SemVer, -alpha[.N] for very early pre-releases, -rc[.x] for almost stable releases candidates
+ [x] Rename `flamegpu2` CMake target to `flamegpu`, as it will/may eventually become `3.x.y` 

This will close #591 and close #163. 

In the future, the location of the version definiation, and checking of the include directory metadata may change (see #600)


cmake/version.cmake
```diff
@@ -0,0 +1,114 @@
+# Manually define parts of the version hash. 
+# @todo - extract these from the header instead, if an alternative can be found for RTC header checking.
+# Due to python limitations, restrict Prerelease component to: alpha, alpha.N, beta, beta.N or rcN (where N is a non-negative integer) or the empty string for final versions
```
mondus, line 3: rc.N

PR #597, opened 2021-07-16
Rtc glm

Robadob:
This PR can be merged as-is, the 3 commits are all be distinct and clearly separated. There is a matching Vis PR [here](https://github.com/FLAMEGPU/FLAMEGPU2-visualiser/pull/70), which is also required for the first of these commits.

# Tests
* Windows/Release/CUDA 11.2/USE_GLM=OFF/SEATBELTS=ON: 880 passes, 25 disabled (Laptop, 10m45s runtime)
* Windows/Release/CUDA 11.4/USE_GLM=ON/SEATBELTS=ON: 902 passes, 5 disabled (Office, 81m8s runtime)
* Windows/Release/CUDA 11.4/USE_GLM=ON/SEATBELTS=ON: 902 passes, 5 disabled (Office, 25m22s runtime, after I added a patch to remove GLM headers if not required)

# To Do
- [x] Rebase vs master
- [x] Improve how GLM is linked in, currently it depends on visualisation
- [x] Make USE_GLM support set/get variables inside RTC curve
- [x] Update vis to support vector variables for agent location (and rotation?)
- [x] Cleanup commits, I think 2918630 needs to be reverted.
- [x] Agent Variable
- [x] New Agent Variable tests
- [x] Environment property test
- [x] ?Add array variable support to messages
- [x] Message Variable tests
- [x] Squash commits, but keep vis changes + message array variable + curve changes separate 
* ~~Update boids examples to use vec?~~ (Not worth doing whilst option is considered experimental, could add a 2nd one just for testing purposes that doesn't get merged).

# Issues
## `__ldg()`
The CUDA intrinsic `__ldg()` cannot be used to load non-primitive types (e.g. all glm types). As such, the brute force variable path, which uses `__ldg()` leads to a compile error. For this reason I have simply disabled `__ldg()` where `USE_GLM` is enabled.

I did try a `constexpr if` to make the `__ldg()` path disabled for glm types, however that did not work (presumably because the path is optimised out after it has been parsed and thrown that error).

I found [this](https://github.com/bryancatanzaro/generics) small header library which overrides `__ldg()`, to load any generic type as multiple loads of a smaller primitive type. We could perhaps use this or something similar.

## RTC build times.

Adding the glm header to the include hierarchy makes RTC build time increase by ~6.4x. I've added a patch that will only enable `USE_GLM` during RTC compilation if a file contains the string `glm` (either them using a glm type or using the namespace). This works as intended.

I looked into this further, and found that NVRTC (`createProgram`, `compileProgram`, ..., `destroyProgram`) has a fairly stable runtime of 0.6s. Whereas the entire call to `jitify::Program` constructor, Takes 9s without glm headers, 63s with glm headers.
So Jitify's mini pre-processor that loads the headers, is to blame not NVRTC. 

---------

Pairs with: https://github.com/FLAMEGPU/FLAMEGPU2-visualiser/pull/70

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,6 +71,7 @@ option(BUILD_FLAMEGPU "Enable building FLAMEGPU library" ON)
 
 # Option to enable/disable building the static library
 option(VISUALISATION "Enable visualisation support" OFF)
+
 if(NOT NO_EXAMPLES)
     # Option to enable building all examples
     option(BUILD_ALL_EXAMPLES "Enable building examples" ON)
--- a/cmake/common.cmake
+++ b/cmake/common.cmake
@@ -18,6 +18,9 @@ include(${CMAKE_CURRENT_LIST_DIR}/Thrust.cmake)
 include(${CMAKE_CURRENT_LIST_DIR}/Jitify.cmake)
 include(${CMAKE_CURRENT_LIST_DIR}/Tinyxml2.cmake)
 include(${CMAKE_CURRENT_LIST_DIR}/rapidjson.cmake)
+if(USE_GLM)
+include(${CMAKE_CURRENT_LIST_DIR}/glm.cmake)
+endif()
 
 # Common rules for other cmake files
 # Don't create installation scripts (and hide CMAKE_INSTALL_PREFIX from cmake-gui)
@@ -332,6 +335,16 @@ function(add_flamegpu_executable NAME SRC FLAMEGPU_ROOT PROJECT_ROOT IS_EXAMPLE)
                     $<TARGET_FILE_DIR:${NAME}>)
         endif()
         add_compile_definitions(VISUALISATION)
+    
+        # Make GLM accessible via include
+        if (USE_GLM)
+            if(glm_FOUND)
+                target_include_directories(${NAME} PUBLIC "${glm_INCLUDE_DIRS}")
+            else()
+                message(WARNING "USE_GLM enabled, but glm_FOUND is False.")
+            endif()
+            add_compile_definitions(USE_GLM)
+        endif()
     endif()
     
     # Pass the SEATBELTS macro, which when set to off/0 (for non debug builds) removes expensive operations.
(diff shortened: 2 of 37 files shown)


Robadob:
Decision to update RTC, so that behind `USE_GLM` it matches non-rtc behaviour. Could do the inverse for non-rtc if it's not too much hassle.

Robadob:
I have now tested the follow combinations for applying rotation. All run without shader compile errors or dissapearing agents due to bad maths.

Would need a more thorough vis example to check for correctness of the math, but that shouldn't have been too badly affected by the PR.

```c++
        // 0
        //circ_agt.setForwardXVariable("fx");
        //circ_agt.setForwardZVariable("fz");
        // 1
        //circ_agt.setForwardXVariable("fx");
        //circ_agt.setForwardYVariable("fy");
        //circ_agt.setForwardZVariable("fz");
        // 2
        //circ_agt.setForwardXVariable("fx");
        //circ_agt.setForwardYVariable("fy");
        //circ_agt.setForwardZVariable("fz");
        //circ_agt.setUpXVariable("x");
        //circ_agt.setUpYVariable("y");
        //circ_agt.setUpZVariable("z");
        // 3
        //circ_agt.setForwardXZVariable("fxz");
        //circ_agt.setForwardYVariable("fy");
        //circ_agt.setUpXVariable("x");
        //circ_agt.setUpYVariable("y");
        //circ_agt.setUpZVariable("z");
        // 4
        //circ_agt.setForwardXZVariable("xz");
        //circ_agt.setForwardYVariable("fy");
        //circ_agt.setUpXYZVariable("xyz");
        // 5
        //circ_agt.setForwardXZVariable("fxz");
        // 6
        //circ_agt.setDirectionYPVariable("xz");
        // 7
        //circ_agt.setDirectionYPRVariable("xyz");
        // 8
        //circ_agt.setPitchVariable("x");
        //circ_agt.setYawVariable("y");
        //circ_agt.setRollVariable("z");
        // 9
        //circ_agt.setPitchVariable("x");
        //circ_agt.setRollVariable("z");
        // 10
        //circ_agt.setPitchVariable("x");
        //circ_agt.setYawVariable("y");
        // 11
        //circ_agt.setYawVariable("y");
        //circ_agt.setRollVariable("z");
        // 12
        //circ_agt.setPitchVariable("x");
        // 13
        //circ_agt.setYawVariable("y");
        // 14
        //circ_agt.setRollVariable("z");
        // 15
        //circ_agt.setDirectionYPVariable("xz");
        //circ_agt.setRollVariable("z");
```

Robadob:
Slight thought, I think i've tested glm from python. But I almost certainly haven't updated python wheel to include/extract glm headers, what are the license implications of this?

Not the end of the world if this isn't in before merge as we're considering it experimental, but something to be aware of (new issue?).

ptheywood:
GLM is MIT, so we just need to make sure the licence ends up in the wheel too.

ptheywood (approved):
Looks fine at a glance, once the jitify compile time issues are sorted we can potentially move this out from behind the flag.

Issue #368, opened 2020-07-30
CLI output to disk


ptheywood:
For F1 parity need to be able to write out state lists to disk every N iterations, when `simulate` is used.

via the config / cli:
+ specify format
+ specify location (Possibly optional, using working dir /next to input if specified?)
+ specify frequency (0 means never, 1 = every iteration, 100 is every 100 iterations etc.)

Should default to being off (much better performance)

Issue #257, opened 2020-05-08
Improved CLI - User provided arguments


ptheywood:
Current CLI could be improved/extended to support user-provided cli args. ~

I.e. if a users wishes to provide environmetn vars from the command line in place of an initial states file, or if they wish to provide a number of agents to generate or similar (some of these cases could be made default optional arguments)

Doing this in a similar way to python's `argparse` module would likely be a nice way to do this, either by exposing an existing cli library to the user, or masking it behind methods on the `Simulator`/`Simulation` object.

It might be useful to have an extra class for the cli as a member of the simulation/simulator, or just do this directly.

This should support:
+ Positional arguments? i.e. `simulation.exe path/to/file`
+ Optional arguments i.e. `simulation.exe --file path/to/file`
+ Flags i.e. `simulation.exe --verbose`
+ Short forms as well as long forms `-v/--verbose`
+ Multiple short arguments with a common `-` - i.e. `-am` maps to `--add --message`

The following alternative/possible methods are to add the following flags/args.

+ `-v`/`--verbose` - A `bool` which is set to `true` if passed, otherwise `false`. 
+ `-n/--number` - An integer number of something such as agents. defaults to 0. 
+ `seed` - A positional <long>, defaults to -1  to imply un-set.

`-h/--help` would be generated by default.

There are more complicated cases of cli which I've not attempted to cover here.

Removal of default arguments might also be useful? I.e. if for some reason the user does not wish to support loading from disk (Which should also be made configurable once we have more than one format supported). 

Flags with multiple arguments would also likely be useful - i.e. `--seeds 1 2 3` or `--seed-max-min 0 1000` or similar, but I've not considered that.

### Option 1: methods on the simulation object

```
// Get the simulation object
simulation = CUDASimulation(); // CUDASimulator()

// -v/--verbose as a boolean flag. False by default, !default if passed.
simulation.add_cli_flag<bool>("-v", "--verbose", false, "Sets verbose output");
simulation.add_cli_argument<int>("-n", "--number", 0, "Number of agents to generate");
simulation.add_cli_positional<long>("seed", -1, "Seed for the simulation/simulator");

// Process the arguments (maybe just use current `initalise()`)
simulation.process_cl(argc, argv);
```

### Option 2: Use a CLI object 

```
// Get the simulation object
simulation = CUDASimulation(); // CUDASimulator()

cli = simualtion.cli();

[...]

PR #609, opened 2021-07-29
CLI Tweaks

ptheywood:
+ Output Full version number in help and if output is verbose / not silent
+ Replace CUDAEnsemble `-s/--silent` with `-q/--quiet` to avoid confusion with Simulation -s/--steps
+ `-h/--help` as expected arguments not unexpected for Ensembles and Simulations
+ remove short options `-oe` `-ol` `-os`. Short options should be single char

### DIFF
--- a/include/flamegpu/gpu/CUDAEnsemble.h
+++ b/include/flamegpu/gpu/CUDAEnsemble.h
@@ -47,10 +47,10 @@ class CUDAEnsemble {
         /**
          * If true progress logging to stdout will be suppressed
          */
-        bool silent = false;
+        bool quiet = false;
         /**
          * If true, the total runtime for the ensemble will be printed to stdout at completion
-         * This is independent of the EnsembleConfig::silent
+         * This is independent of the EnsembleConfig::quiet
          */
         bool timing = false;
     };
--- a/src/flamegpu/gpu/CUDAEnsemble.cu
+++ b/src/flamegpu/gpu/CUDAEnsemble.cu
@@ -183,6 +183,11 @@ int CUDAEnsemble::checkArgs(int argc, const char** argv) {
         // Get arg as lowercase
         std::string arg(argv[i]);
         std::transform(arg.begin(), arg.end(), arg.begin(), [](unsigned char c) { return std::use_facet< std::ctype<char>>(std::locale()).tolower(c); });
+        // -h/--help. Print the help output and exit.
+        if (arg.compare("--help") == 0 || arg.compare("-h") == 0) {
+            printHelp(argv[0]);
+            return false;
+        }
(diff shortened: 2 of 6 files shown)


Robadob:
> * Replace CUDAEnsemble -s/--silent with -q/--quiet to avoid confusion with Simulation -s/--steps
> * -h/--help as expected arguments not unexpected for Ensembles and Simulations

These seem like good shouts

> * remove short options -oe -ol -os. Short options should be single char

I disagree with convention here. If you really want change it to `-o [e/l/s/exit/l?/step] <path>` so the o is a single char.

ptheywood:
The convention for single char only short options is so that you can provide multiple at once (although our implementation doesn't currently support it, and but it will probably come for free once #257 is addressed). 

I.e. `tar -zxvf <file>` is synonymous with `tar -z -x -v -f <file>`.

>  If you really want change it to -o [e/l/s/exit/l?/step] <path> so the o is a single char.

This is more than a quick fix that I was making while other stuff built, and I'm not currently clear on what those files actually produce. The CLI for output is something that needs further work too (#368) for FLAMEGPU 1 parity / checkpointing, so I'd rather remove the short options now (pre alpha release) and then a more correct implementation be provided (soon?) after.

Robadob:
> The convention for single char only short options is so that you can provide multiple at once 

Oh.

> for FLAMEGPU 1 parity

We shouldn't hamper our CLI trying to enforce compatibility with the arguably bad FGPU1. If people want to pass FGPU args it should be after a unique arg `--legacy` or similar.

ptheywood:
> We shouldn't hamper our CLI trying to enforce compatibility with the arguably bad FGPU1. If people want to pass FGPU args it should be after a unique arg --legacy or similar.

I'm not suggesting the CLI has to be the same, but the peiriodic output is useful (i.e. checkpointing) so we should have a similar feature. CLI will be different anyway as we're not going to start using positional arguments for it. That can be part of #368 in the future though, rahter than in this quick fix before alpha release PR.

Robadob:
> but the peiriodic output is useful (i.e. checkpointing)

Having a `printf("\r%d/%u", stepCount, config.steps)` or similar is fine. Could even make it default behaviour. Just want to avoid the spam that FGPU1 creates by doing 1 line per step by default.

ptheywood:
By checkpointing i mean periodic output of agent state (and ideally rng state), so long runnign simulations may be resumed from a fixed point (i.e. to support pre-emptable HPC queues, or to run very long duration simulations on machines with too short maximum runtimes).

Robadob:
Oh, that's not even something I've thought about. Makes sense, but doesn't seem like an urgent priority. Shouldn't be too difficult to support in future though.

PR #612, opened 2021-08-03
Adjust RTC environmental variables.

ptheywood:
+ Checks for `CUDA_PATH` and then `CUDA_HOME` to find the cuda include directory.
+ Checks for `FLAMEGPU_INC_DIR` and then `FLAMEGPU2_INC_DIR` for the flamegpu include direcotry

Part of #253

### DIFF
--- a/src/flamegpu/util/detail/JitifyCache.cu
+++ b/src/flamegpu/util/detail/JitifyCache.cu
@@ -2,6 +2,7 @@
 
 #include <cassert>
 #include <regex>
+#include <array>
 
 #include "flamegpu/version.h"
 #include "flamegpu/exception/FLAMEGPUException.h"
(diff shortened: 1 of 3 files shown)


src/flamegpu/util/detail/JitifyCache.cu
```diff
@@ -81,74 +82,138 @@ std::string loadFile(const path &filepath) {
     ifs.close();
     return rtn;
 }
-}  // namespace
 
-std::mutex JitifyCache::instance_mutex;
```
Robadob, line 86: Doc comments go in the header pls

src/flamegpu/util/detail/JitifyCache.cu
```diff
@@ -81,74 +82,138 @@ std::string loadFile(const path &filepath) {
...
-    static std::string env_cuda_path = std::getenv("CUDA_PATH") ? std::getenv("CUDA_PATH") : "";
-    if (env_inc_fgp2.empty()) {
+/**
+ * Find the cuda include directory.
+ * Throws exceptions if it can not be found.
+ * @return the path to the CUDA include directory.
+ */
+std::string getCUDAIncludeDir() {
+    // Define an array of environment variables to check in order
+    std::array<const std::string, 2> ENV_VARS { "CUDA_PATH", "CUDA_HOME" };
```
Robadob, line 93: This feels like something vs gives warnings about (don't think it likes const std::string in collections for some reason).

Robadob (approved):
Looks good.

src/flamegpu/util/detail/JitifyCache.cu
ptheywood: CI didn't seem to trigger so it looks like this isnt' quite what MSVC gets angry about, but I do seem to recall something along these lines being problematic in the past.

Issue #605, opened 2021-07-27
Python Binary Distribution with multiple configurations


ptheywood:
The initial python binary distributioin(s) will be / are Release mode, seatbelts ON only. 

This is the default, sensible choice for python users to be able to develop models that run reasonably well (i.e. not terribly slow debug builds) while also getting sane error messages if there are issues with agent functions. If python users require faster builds, they will have to build from source initially.

Longer term, it would be nice to provide pre-built binaries for seatbelts=OFF, or to support 2 major cuda versions, or other options that result in a larger build matrix. Doing this in a wheel compliant way is non-trivial. 

Python namespace packages are a potential solution, or multiple packages which users can import with alternate names (i.e. `pyflamegpu-seatbeltsoff`, which users can then import as pyflamegpu for a 1 line change)

ptheywood:
Using cibuildwheel might be a reasonable way to build pypi compliant wheels? Would need investigation.

https://pypi.org/project/cibuildwheel/

ptheywood:
Rather than renaming wheels `pyflamegpu` to `pyflamegpu-console` etc manually, it appears that this might actually be possible via setuptools, while the python package itself is still called `pyflamegpu`. 

I.e. 
```python
setuptools.setup(name="pyflamegpu-console", packages=["pyflamegu"])
```

In this case, the generated wheel would be `pyflamegpu-console-<stuff>.whl`.

If we were uploading these to pip, it woudl be separate package, available as `pyflamegpu-console` while the main `pyflamegpu` package would have vis on?. The wheels would still be out-of-spec however due to the libcuda.so dependency preventing them from being truly conforming manylinux wheels.  

Package extras may be another option. e.g. `pyflamegpu` would be the console build, and `pyflamegpu[visusliation]` would also install the visusaltiion components. This would need changing to how the `VISUALISATION` macro works in C++, and might also need swig splitting into modules, I'm not sure.  The general use of extras seems to be that the main package is minimal, with extras being opt in. This would work for vis if we resturcture it, but I don't think it'll work for seatbetls on/off. 
I am leaning towards console only being default, and vis be an opt-in extra, as vis is not always required and doesn't exist for ensembles.

So I if we do go down the multiple packages route `pyflamegpu` is console mode, and `pyflamegpu-visusalition` is with vis. 
It might be possible to combine these approaches (but I need to read more / try things). 

[...]

Issue #604, opened 2021-07-27
Static Library Distribution


ptheywood:
To save compilation time, it may be useful to make the static library / include directory installable from a pre-compiled version (via CI).

The best way to do this, is probably via a cmake install phase which copies the static library somewhere useful (and configurable), along with the include directory, and creates cmake config files which provides it as an imported target (via find_package). 

This would then make downloading from a release viable. Standalone examples could then be updated to find a required flamegpu version, and if it cannot be found to download it, or download and build from source instead. 

There may be some complexity relating to the CUDA version used, and the compute capabilities built for  that will need handling/considering.

This will not be in the initial alpha release(s).

Issue #603, opened 2021-07-27
Example Suite Repository/Repositories


ptheywood:
The plan is to separate out the examples into thier own repository(s), allowing them to be versioned separately to the core library.

Potentially with some cmake to allow a "fetch me the examples and build them all please" option or somethign to that effect. This may cause a version / dependency loop though, so it might be better to have a separate repo with all examples that also pulls in FLAMEGPU2's core repository. 

This will then allow generation of executables at a different release candence to the main library (and mean that any fixes to say `boids_spatial3D` does not implicitly lead to a new library release (if not neccesary).

Robadob:
Fine with me.
Minimum request being we add examples/* to git ignore, so I've atleast got a hacky way of keeping some minimal example around for debugging.

ptheywood:
The alternative would be that we adjust the examples cmake to make it easy to specify FLAMEGPU_ROOT or similar pointing at a separate local directory, but yeah no harm in keeping examples in .gitignore either.

Robadob:
Yeah, I suppose the inverse might work better. A hidden `EXAMPLES_ROOT` option to enable a non-rtc and rtc example for testing in main repo.

ptheywood:
An alternative (or additional point) may be to split the current `examples` directory into subfolders to distinguish between C++ and python examples. Probably better not to use `+` in a directory path though, so `cxx`?

ptheywood:
During meeting splitting the examples out or not prior to rc1 was discussed, concluding that: 

+ We don't want any heavy-weight examples in the main repo, to keep the lib itself lighter (i.e. no massive vis objects or input state files)
+ The chicken-and-egg version dependency issues of pulling in versioned examples to the main repo via a flag makes it not worth splitting out but allowing to pull in from the main repo.
  + Having standalone examples which pull in a speicfic version is OK/Good though
+ Folders are a good idea, to split the examples into:
  + Python (`python`)
  + c++ interface (`CUDA`? `cxx`? `CUDACXX`?)
  + c++ RTC interface (`RTC`? `(CUDA/cxx/CUDAcxx)RTC`
  + c++ not actual models but rather demonstrations of features
    + I.e. `ensemble` so we have an ensemble binary to build and test the CLI.
    + Alternatively we could just make it an actual ensemble of a given model? 
+ The host functions example is not super useful, as the tests + docs cover this use case too. If another example covers the host angle though, this could safely be removed.

ptheywood:
[...]

PR #613, opened 2021-08-03
Add CI Workflow to create draft releases from pushed tags

### LINKED ISSUE #296: Tag/Release Strategy using GitHub Actions
Releases are tagged commits with extra information (prerelease?, description, extra zip files etc) which can then be used with predictable download locations (for zip download).

For FLAMEGPU 1, when we create a (minor?) release we also manually attach a zip file containing the compiled executables for windows, using the visual studio and CUDA versions within the vcxproj files, so users do not have to compile code to try FLAME GPU.

For FLAME GPU 2 we can continue and automate this tradition (through github actions). We could also attach artifacts containing python related parts, or the library + headers.

GitHub Actions can be triggered for both `push.tag` events and `release` events.

When a new **tag** is pushed to github the `on.push.tags` event is triggered (can be filtered by tag name). 
This includes new tags created by creating new releases from untagged commits through the github releases UI.

**Release** events are assocaited with a given tag, either pre-existing or newly created through the UI. There are 6 types of release events. 
Actions only occur for non-draft releases.

+ `published` - when an event is made non-draft
+ `unpublished` - never acted upon, as unpublished === draft.
+ `created` - when a new release is created and published at the same time (can be created but unpublished, so action does not trigger)
+ `edited` - when the text of the release is modified, but not when the tag is changed (during testing)
+ `deleted` - when a published release is deleted
+ `prereleased` - when a published release is mared as a pre-release (creation or edit)
+ `released` - when a published release is marked as not pre-release (creation or edit)

#### Option 1 - manual releases
+ On `tag` events, trigger more thoroguh CI (all supported compute capabilities on all supported platforms) (if not already triggered for that commit?)
+ On `release` events of the types `published` build artifacts, and attach to the release

New releases *should* **always** be created pre-release, and only marked as release once we are confident in the release (i.e. all tests + maybe manual tests).

Once we have GPU enabled actions (#97) we could automate/enforce the release/pre-release status.

+ on `pre-release` events, run the gpu-enabled CI test suties - if it passes mark as a `release` 
+ on `released` events, if the mode detailed CI has not been ran, mark it as pre-release
    + This might require adding a PAT to result in the workflow-launched-by-workflow.
    + Or alternatively, on released events mark as pre-release then run the tests and if it passes mark it as a release again.

There is an edge case where if the tag which a release points to is changed, it does not trigger a release event - The solution is to **not change release tags**. Alternatively it *may* be possibly to run an action on push.tag events, which checks for an existing release associated with that commt hash via the api. 

We may also need to restrict the author of the tag (not sure how PRs from forks with tags work, but we don't want to do this for external tags without explicit support)

#### Option 2 - automatic releases based on tags.

On `tag` events trigger the more thorough CI / actual tests. 
If it passes, create a new release based on the tag, and attach relevant artefacts.
This would likely need the thorough build to include multiple jobs for the respective OS's being tested:

+ Job 1 - linux, save artefacts
+ Job 2 - windows, save artefacts
+ Job 3 - create and the release and attach. Depends on Jobs 1 and 2 being successeful (for the full build matrices)

A mechanism for getting the changelog for the description would be required.

---

I've made this separate to #290 as it's lower priority / longer-term.

### LINKED ISSUE #514: Release Binary distribution
Need to figure out how to do binary distributions.

c++ should be fine similar to F1, but automated via an action (on push to tags which match a pattern?)

Python will be more difficult. [PyQuest](https://github.com/HQSquantumsimulations/PyQuEST-cffi) is potentially a very useful reference.

Related:
+ #296 
+ #163

ptheywood:
+ Add CI Workflow to create draft releases from pushed tags
    + This is only triggered by tag pushes that match semver version numbers, prefixed with `v`. 
+ Also prevents other workflows running on tag pushes.
+ Adds a CMake variable to specify an exact python version to search for. This is required for windows github actions, but otherwise shouldn't be neccesary

The new workdlow will not be ran by this PR which is intentional. Instead i've pushed a tag to [ptheywood/FLAMEGPU2](https://github.com/ptheywood/FLAMEGPU2) as a clean fork to demo this. 

https://github.com/ptheywood/FLAMEGPU2/actions/runs/1095219616

I'll add a screenshot of the draft release and make it non-draft once it has completed, then once this has been merged and anything else to go in pre `2.0.0-alpha` we can make the release. 

Closes #296 
Closes #514  (subsequent improvments to binary releases will instead be covered by #603, #604 and #605)

### DIFF
--- a/.github/workflows/Docs.yml
+++ b/.github/workflows/Docs.yml
@@ -3,6 +3,8 @@ name: Docs
 
 on:
   push:
+    branches:
+      - '**'
   pull_request:
 
 jobs:
--- a/.github/workflows/Draft-Release.yml
+++ b/.github/workflows/Draft-Release.yml
@@ -1,0 +1,335 @@
+# Thorough build for pushed tags.
+# Builds all supported binary platforms to ensure that release artifacts can be generated.
+name: Draft Release
+
+on:
+  # Trigger on pushes to matching tags.
+  push:
+    tags:
+      - 'v*.*.*'
+  # Or trigger on manual dispatch. This will not produce a release, but will perform the thorough build. 
+  workflow_dispatch:
+
+defaults:
+  run:
+    # Default to using bash regardless of OS unless otherwise specified.
+    shell: bash
+
+jobs:
+  # Perform a thorough build, including test suite on all platforums with a wide range of compute capabilities.
+  # Build artifacts are saved to attach to a draft release if successful 
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      # Large build matrix, with linux vis, linux console, windows vis builds to produce .whl's.
+      # Single CUDA version, building for each supported major architecture, using each python version. Release builds with SEATBELTS=ON for now. GCC 8 for CUDA 11.0 builds of the test suite.
+      # Only one build per python config is set to build the test suite (as it would be the same work otherwise)
+      # Additionally an linux vis build using cuda 10.0 to ensure oldest supported CUDA builds OK, but does not produce binaries.
+      matrix:
+        include:
+          # CUDA 11.0 Linux builds
+          - os: ubuntu-20.04
+            cuda: "11.0"
+            python: "3.9"
+            cuda_arch: "35 52 60 70 80"
+            gcc: 8
+            config: "Release"
+            SEATBELTS: "ON"
+            VISUALISATION: "ON"
+            build_tests: "ON"
+            binary_release: "ON"
+          - os: ubuntu-20.04
+            cuda: "11.0"
+            python: "3.8"
+            cuda_arch: "35 52 60 70 80"
+            gcc: 8
+            config: "Release"
+            SEATBELTS: "ON"
+            VISUALISATION: "ON"
+            build_tests: "OFF"
+            binary_release: "ON"
+          - os: ubuntu-20.04
+            cuda: "11.0"
+            python: "3.7"
+            cuda_arch: "35 52 60 70 80"
+            gcc: 8
+            config: "Release"
+            SEATBELTS: "ON"
+            VISUALISATION: "ON"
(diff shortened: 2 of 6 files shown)


ptheywood:
After the new CI workflow completes, it creates a draft Release  on the releases page, which can only be seen by people with access. I.e. for my fork with a mock FLAME GPU 2.0.0-alpha release:

image

Clicking `Edit` allows the title / description to be written, and the release to be published (made public). 

I've written a very rough idea of what the description might look like for the early alpha release(s) and plublished this release on my fork. 

https://github.com/ptheywood/FLAMEGPU2/releases/tag/v2.0.0-alpha.

Currently, this builds for the 4 currently supported python versions (3.6-3.9) while official 3.6 support will be dropped at the end of the year, so we can adjust which python versions are in the matrix as required. 

Seatbelts on, Release builds, CUDA 11.2, SM 52+. `pyflamegpu` wheels have visusaliastion ON, pyflamegpu-console has visusalisation off for linux builds, but will overwrite the vis=on wheel if both are installed (as the actual python module is just pyflamegpu, we should be able to find a cleaner solution for this with some faff.).

ptheywood:
Need to switch to CUDA 11.0 and SM 35 as the minimum due to wanting to support google colab for python binary releases.

ptheywood:
Set a SM35 CUDA 11.0 build going on my fork to test teh force-pushed commit.

ptheywood:
Set another demo going on ptheywood/FLAMEGPU2, with tag `v2.0.0-alpha.2`. This should be the final state of the repo for the initial alpha release, other than the readme (which won't impact the Release CI workflow). 

Once that's confirmed to be working as intended (I.e. no new warnings have been introduced for the windows test suite CI)  I'll get this PR merged in, and once the new readme PR is sorted and merged I'll make the initial release (or atleast push the tag)

Issue #608, opened 2021-07-28
RTC compilation should provide the correct line numbers in compile errors.


Robadob:
Currently line numbers are offset in compile errors, due to the injected includes.

This can be resolved by finally injecting a `#line` pragma, to reset the line number. I have tested that NVRTC supports this feature.

`#line 1` should suffice, unless we also want to use this to rename the file to match that exported by the relevant CMake flag to solve another issue.

https://gcc.gnu.org/onlinedocs/cpp/Line-Control.html

PR #615, opened 2021-08-04
RTC #Line

Robadob:
~~This is currently a draft, we need to decide how we want these line numbers to appear.~~

e.g.

If a user has not got `EXPORT_RTC_SOURCES` enabled, then their only reference to line, is the string/file they pass to FLAMEGPU. So first line of that should be line 1 (ignoring injected includes etc).

If a user has got `EXPORT_RTC_SOURCES` enabled, then instead it can be assumed they will be referring to the exported files. In which case, the start of that string should be appended `#line 2 filename`. 2, because the exported source will include the `#line` statement on line 1.

I'm not really a fan of this inconsistency, but it makes sense. So if agreed I will tweak the PR to handle those cases.

Addresses #608

### DIFF
--- a/include/flamegpu/model/AgentFunctionDescription.h
+++ b/include/flamegpu/model/AgentFunctionDescription.h
@@ -188,9 +188,20 @@ class AgentFunctionDescription : public DependencyNode {
      * Only agents which return true perform the attached FLAMEGPU_AGENT_FUNCTION
      * and transition from the initial to end state.
      * The string will be compiled at runtime.
-     *
+     * @param func_cond_src Source containing RTC agent function condition source
+     * @see setRTCFunctionConditionFile()
      */
     void setRTCFunctionCondition(std::string func_cond_src);
+    /**
+     * Sets the RTC function condition for the agent function, with an agent function condition loaded from file
+     * This is file containing a definition of an FLAMEGPU_AGENT_FUNCTION_CONDITION which returns a boolean value (true or false)
+     * Only agents which return true perform the attached FLAMEGPU_AGENT_FUNCTION
+     * and transition from the initial to end state.
+     * The string will be compiled at runtime.
+     * @param file_path Location on disk of file containing RTC agent function condition source
+     * @see setRTCFunctionCondition()
+     */
+    void setRTCFunctionConditionFile(const std::string& file_path);
 
     /**
      * @return A mutable reference to the message input of this agent function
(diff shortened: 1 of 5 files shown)


ptheywood:
Sounds like the best choice, although i agree that the inconcistency feels a bit grim, but probably the most useful / practical option.

Robadob:
Another thought, re RTC agent functions defined in code. They open the string on it's own line. So presumably the string begins with a newline, so really want to trim that opening newline if it exists, because no one is considering an empty newline line 1

ptheywood (changes requested):
Have you ran the  full test suite on this? I get errors with RTC function conditions

```
[  FAILED  ] 2 tests, listed below:
[  FAILED  ] DeviceRTCAPITest.AgentFunction_cond_rtc
[  FAILED  ] RTCMultiThreadDeviceTest.SameModelMultiDevice_AgentFunctionCondition
```

```
[ RUN      ] DeviceRTCAPITest.AgentFunction_cond_rtc
--- JIT compile log for #line 1 "rtc_test_func_impl_condition.cu" ---
#line 1 "rtc_test_func_impl_condition.cu"(1): error: this declaration has no storage class or type specifier

limits(7): error: expected a ";"

limits(50): warning: parsing restarts here after previous syntax error

limits(51): error: expected a declaration

limits(108): error: expected a declaration

limits(111): warning: parsing restarts here after previous syntax error

limits(112): error: numeric_limits is not a template

limits(113): error: name followed by "::" must be a class or namespace name

limits(113): error: expected an identifier

limits(113): error: not a class or struct name

limits(113): error: class or struct definition is missing

At end of source: warning: parsing restarts here after previous syntax error

At end of source: error: expected a ";"

10 errors detected in the compilation of "#line 1 "rtc_test_func_impl_condition.cu"".

unknown file: Failure
C++ exception with description "/home/ptheywood/code/flamegpu/FLAMEGPU2/src/flamegpu/util/detail/JitifyCache.cu(331): Error compiling runtime agent function (or function condition) ('rtc_test_func_condition'): function had compilation errors (see std::cout), in JitifyCache::buildProgram()." thrown in the test body.
```

Robadob:
Nope, will get around to that Monday. Good shout, was supposed to be a simple change.

Robadob:
I presume it's the same issue I had on agent functions, I just forgot to remove the breaking bit from conditions too.

Robadob:
It can't be, that's one bit of code so same fix applies.

Looking at this, it might not even like the `#line` statement

> `#line 1 "rtc_test_func_impl_condition.cu"(1): error: this declaration has no storage class or type specifier`

Robadob:
Debug build.
```
[==========] 868 tests from 68 test suites ran. (791154 ms total)
[  PASSED  ] 868 tests.

  YOU HAVE 38 DISABLED TESTS
```

ptheywood (approved):
Linux tests now all pass.

PR #621, opened 2021-08-04
Rename Msg* classes to Message*

### LINKED ISSUE #585: Message Class Names / paths on disk
As of #551, messages classes are `flamegpu::MsgNone`, `flamegpu::MsgBruteForce` etc, however the header files are `flamegpu/runtime/messaging/None.h`, `flamegpu/runtime/messaging/None/NoneHost.h` etc. 
This does not match.

Options:
+  Move these classes into a `messaging` namespace, and remove `Msg` from the class name.
    + Friendship related issues, but only the one. Swig might need some adjusting.
+ Change the file location on disk to match the class name , i.e. `flamegpu/runtime/messaging/MsgNone.h` and `flamegpu/runtime/messaging/MsgNone/MsgNoneHost.h` (or `flamegpu/runtime/messaging/MsgNone/Host.h`?)
+ Rename the classes *and* change the location on disk. `Msg` -> `Message`, there is no need to avoid  verbosity.

ptheywood:
Closes #585.


ptheywood:
This is going to conflict with #597. I'll merge GLM after giving it a quick review on Friday and then rebase this on top of it (then fix the conflicts, and update any extra `Msg*` which have been introduced by #597.

PR #635, opened 2021-08-11
Change the default number of Simulation steps to 1

### LINKED ISSUE #634: Set -s/--steps default to 1
The current default of running forever by default is bad form

ptheywood:
Closes #634

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
++ Created `CHANGELOG.md` ([#630](https://github.com/FLAMEGPU/FLAMEGPU2/pull/630))
+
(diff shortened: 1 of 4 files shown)


Robadob (approved):
Looks good.

Thought you'd said, would leave it 0 for vis, but the consistency like this is probably better.

ptheywood:
Paul suggested that he preferred consistency during the meeting (If I remember correctly). 

Users can still programatically override the default value themselves `#ifdef VISUALISATION` if required. 
If the vis default were different too it would have some interesting behaviour with the vis build but run as console flag, should it be 0 or 1 in that case etc.

Robadob:
Yeah, it would need a fancy workaround that tracks arg parsing etc, to change it if not set and --console passed.

Robadob:
Don't worry, merge it. Will just have to change my muscle memory for running examples.

PR #638, opened 2021-08-11
MessageBucketDevice Sign Comparison Warnings

### LINKED ISSUE #554: BucketDevice.h sign comparison warning (clang)
Buidling with clang as the host compiler highlighted the following warning (not emitted by msvc/ gcc even though sign comparion warnings are enabled afaik). 

```
BucketDevice.h:309:20: warning: comparison of integers of different signs: 'const unsigned int' and 'const flamegpu::IntT' (aka 'const int') [-Wsign-compare]
```

Not super urgent, but worth remembering to resolve.

Also highlighted missing `override` statements which i've enabled in an ongoing PR.

ptheywood:
Use IntT in place of unsigned int where required in MessageBucketDevice

This resolves the clang warnings about sign comparison. 

Closes #554

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Changed
++ Use `IntT` in `MessageBucketDevice` to resolve clang sign comparison warnings ([#554](https://github.com/FLAMEGPU/FLAMEGPU2/issues/554))
 
 ### Deprecated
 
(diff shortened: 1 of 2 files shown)


Robadob:
Where's the sign comparison though? Visual studio should be giving same warning, Unsigned is the appropriate type.

ptheywood:
It was comparisons like lines 322 and 335 I think: https://github.com/FLAMEGPU/FLAMEGPU2/blob/67055a59dae7a16c82519c57cb5b246e015f112a/include/flamegpu/runtime/messaging/MessageBucket/MessageBucketDevice.cuh#L322

Wasn't sure if the `IntT` or `unsigned int` was correct, but went with `IntT` as that felt more intentional.

Robadob:
Yeah I removed my comment. I noticed the docucoment already said `IntT`

PR #640, opened 2021-08-13
Use SteadyClockTimer in CUDASimulation for WDDM devices

### LINKED ISSUE #451: WDDM and cudaEventElapsedTime
`util/CUDAEventTimer.h` uses `cudaEvent_t` created in the default stream and `cudaEventElapsedTime` to get record elapsed time with high precision.

Under linux (and presumably TCC windows) this appears to be accurate regardless of what happens between the events.

Under windows on WDDM devices, the `cudaEventRecord` calls appear to be handled through the WDDM command buffers (not too suprising). This however prevents accurate timing of host code.
Timing of device code between the events looks OK based on profiler output, but host code after the first `cudaEventRecord` is issued and before the WDDM command buffer is flushed may be skipped (i.e. a thread sleep). 

For now, the timing is probably fine as it'll almost always be used on linux or tcc mode devices in our case. 
We should investigate this, and potentially drop back to using `chrono::stead_clock` on windows WDDM, which will be fine for timing full simulations, but may not be good enough for timing individual steps or individual layers.

The precision of `steady_clock` is implementation specific. `std::chrono::high_resolution` clock at a glance looks like a better choice, however it is also implementation specific, either being `std::chrono::steady_clock` or `std::chrono::system_clock` depending on the compiler. We want a monotonic clock for timing purposes, so `system_clock` is not a good choice.

### LINKED ISSUE #564: Do not capture timing if there is no work to time.
Cuda event timers are used to collect timing information for the various sections of a simulation (init functions, step tim, step functions, exit functions).

Although very cheap, these timers do have a cost (`~12us` according to `nsys` to create 2 events, record 2 events, stop 2 events, and syncrhonise them / get the elapsed time).

In cases where there is nothing to do (i.e. if there are no exit functions) then there is no point in timing the exit functions.

For input and exit, this won't help much, but for per-iteratiion recording of step functions if there are none then this might add up to something measurable eventually.

Super low priority as low cost, but would be a very minor efficiency improvement.

ptheywood:
Use SteadyClockTimer in CUDASimulation for WDDM devices
    
+ Adds flamegpu::util::detail::wddm utiltiy methods to check if a device is WDDM or not.
  + This may need revising once WSL is available, where currently only WDDM devices are available, but this implementation assumes they only exist if MSC_VER is defined (which won't be the  case for WSL?)
    + In part this is due to the cudaDevAttrTccDriver property being only being set to 1 on windows if using tcc.
+ Creates a based flamegpu::util::detail::Timer class which CUDAEventTimer and SteadyClockTimer derive from, for a common implementation
+ CUDASimulation selects the timer to use via an anon namespace static / method (to avoid excessivly querying the WDDM properties of the device)
+ Timing resolution is removed from docstrings where SteadyClock timer may be used as it is unknown
+ SteadyClockTimers are use all the time for RTC compile time timing, Init and Exit functions as these are not device work, so timers may not be accurate.

Closes #451 
Closes #564 

## Todo:

+ [x] Test this on Windows, with a WDDM device (and maybe TCC too?)
   + Correctly Identifies WDDM devices on windows. I've not currently got access to a TCC device booted into windows to check, but I'm confident it will work because the windows code branch works using the property which will be set to `1` on tcc windows devices.
+ [x] Decide if we should just use SteadyClockTimers in place of CUDAEventTimers for anything that includes a significant amount of non-device code in general?
+ [x] Decide if the timers being used should be output or not. 
    + It is unclear what the exact precision of `std::chrono::stead_clock` is, as it is implementation specific, and they available values programatically just return `std::ratio` such as `std::nano` when built with gcc under linux which is the minimum unit representable by the type used, not the minimum recordable time. There doesn't seem to be a good way to access this information. Grand scheme, is it even relevant?

### DIFF
--- a/include/flamegpu/exception/FLAMEGPUException.h
+++ b/include/flamegpu/exception/FLAMEGPUException.h
@@ -399,9 +399,9 @@ DERIVED_FLAMEGPUException(VersionMismatch, "Versions do not match");
  */
 DERIVED_FLAMEGPUException(InvalidFilePath, "File does not exist.");
 /**
- * Defines an error indicating that a CUDAEventTimer was queried without being synced.
+ * Defines an exception indicating that the flamegpu::util::detail::Timer has been used incorrectly.
  */
-DERIVED_FLAMEGPUException(UnsycnedCUDAEventTimer, "Elapsed time requested for Un-synced CUDAEventTimer");
+DERIVED_FLAMEGPUException(TimerException, "Invalid use of Timer");
 /**
  * Defines an error reported by AgentFunctionDependencyGraph if the graph is invalid
  */
--- a/include/flamegpu/gpu/CUDASimulation.h
+++ b/include/flamegpu/gpu/CUDASimulation.h
@@ -218,35 +218,30 @@ class CUDASimulation : public Simulation {
 
    /**
      * Get the duration of the last time RTC was iniitliased 
-     * With a resolution of around 0.5 microseconds (cudaEventElapsedtime)
      * @return elapsed time of last simulation call in milliseconds.
      */
     float getElapsedTimeRTCInitialisation() const;
 
     /**
      * Get the duration of the last call to simulate() in milliseconds. 
-     * With a resolution of around 0.5 microseconds (cudaEventElapsedtime)
      * @return elapsed time of last simulation call in milliseconds.
      */
     float getElapsedTimeSimulation() const;
 
     /**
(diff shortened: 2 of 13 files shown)


ptheywood:
Meeting verdict:
+ Use Steadyclock timer for total simualtiontime, CUDA event for more precise timers (if possible)
+ Not oututting which ised or precision

PR #642, opened 2021-08-13
Install more specific CUDA packages on ubuntu CI

### LINKED ISSUE #641: Install more precise CUDA packages install_cuda_ubuntu
The linux CI CUDA install script installs the general `cuda-libraries-dev-X-Y` package. Some of these dev libraries are quire larger, resulting in over a GB being downloaded. 

The Windows install script  is more precies, installing just the dev subpackages for curand and nvrtc. 

We can apply the same to the cuda packages, but need to remove some of the automation (the prefixing of the user-provided package with `cuda-` as the individual library packages required are `cuda-nvrtc-dev-11-4` and `libcurand-dev-11-4`. 

This will only be a minor improvement to the CI, with the install step only taking ~2 minutes as is, but faster CI is better. 

I may end up doing this while writing the centos/manylinux install script.

ptheywood:
Closes #641 

Might be more temperamental with new CUDA releases if packages move around rather than using the much larger, less volatile meta packages, but saving ~2mins per build is probably worthwhile in terms of productivity?

For CUDA 11.4, the install CUDA phase now takes 25s compared to 2m37s from the last build of master, downloading 132MB instead of 1309MB
For CUDA 10.0, the install CUDA phase now takes 25s compared to 1m30s from the last build of master, downloading 124MB instead of 667MB

Most of the time was spent unpacking the packages rather than downloading directly, but the download sizes give an indication of scale.

### DIFF
--- a/.github/scripts/install_cuda_ubuntu.sh
+++ b/.github/scripts/install_cuda_ubuntu.sh
@@ -11,10 +11,13 @@
 # @todo - pass this in from outside the script? 
 # @todo - check the specified subpackages exist via apt pre-install?  apt-rdepends cuda-9-0 | grep "^cuda-"?
 
-# Ideally choose from the list of meta-packages to minimise variance between cuda versions (although it does change too)
+# Ideally choose from the list of meta-packages to minimise variance between cuda versions (although it does change too). Some of these packages may not be availble pre cuda 10.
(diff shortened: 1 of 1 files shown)


Robadob (approved):
If you think it's better, then probably worth while.

Small change regardless, so can revert if it turns out problematic.

Issue #308, opened 2020-06-12
Atomically Mutable Environment Variables


ptheywood:
FLAME GPU 1 only has environmental variables, which are mutable from the host (as they are `__constant__`.

FLAME GPU 2 so far has environmental constants and environmental variables, where constants cannot be changed one initialised (i.e. `PI`) and variables which can be modified from the host.

It would also be useful to have environmental variables which are atomically mutable from within device code, i.e. for debug counters. 

These would need to be accessed and modified through specific methods which wrap the appropriate CUDA atomic operations, such as `atomicInc`, `atomicDec`, `atomicAdd`, `atomicSub` and `atomicExch` (maybe others). 

Care would be required when reading these variables within a layer which also contains mutations to those variables, as this would very easily lead to unintended/unexpected behaviour / care would be required with them in general.
This could possibly controlled/prevented by dependency analysis, although this would require extra information on the agent function description objects.

Robadob:
I've been thinking about these in the context of `CUDAEnsemble` logging and the validation counters I use heavily when ensuring primage model correctness (#397).

For my requirements, I'd personally require a separate data type e.g. 'counters', that can be atomically incremented (or we could stretch as far as manipulated for more general applications). These would then be read by accessible in host code, and `CUDASimulation`/`CUDAEnsemble` logging.

Functionally, they would also load addresses via curve, but these would then be operated on atomically.

Personally, I mostly use them for validation which in part can lead to extra runtime cost. So it would be nice if a further CMake flag could silently disable them, similar to `NO_SEATBELTS`. I'm not however sure how this would fit alongside making them available for regular model operation (Primage may require them for the CA grid thing, but we did discuss alternate strategies).

Robadob:
Closed by #643

Issue #162, opened 2019-11-29
Large Environment Properties


Robadob:
Striuvad has the requirement of storing ~32,000 host managed agents. These should be represented as environment properties (for technical reasons), but clearly won't fit in constant cache so need to be allocated in global memory (accessed via ldg?).

cc @ptheywood

ptheywood:
The upper limit on the population is more like `2^12`, although in practice that will never be achieved.

MILeach:
Came up with this which allows pretty uniform creation of properties with any dimensions [https://gist.github.com/MILeach/1ff2da85091705bdaec97b05504df5c4](link). I think this solves what you wanted with same class name with different template args @Robadob? Internally uses std::array at the moment to store the dimensions but could use any representation. Also currently uses c++17 for the fold expression but can workaround that if necessary

I'm in favour of allowing multiple dimensions as it gives a better abstraction of domain problems and reduces scope for error with type safety & bounds checking. It also is less work for a modeller if we do it, although even for multi-dimensional properties, some conversion will probably be required from location to index.

Robadob:
Few problems with your suggested code @MILeach :
* Using `new` in device code is VERY expensive.
* RTC is unable to compile `std::array` currently (the system header is missing, it would probably work if a replacement was added to Jitify).

MILeach:
Yeah the new and std::array were really just there as examples/placeholders, memory allocation would obviously need to be handled differently. 

Yeah that would work too, I think it depends where we want to draw the line on flexibility. My implementation would allow 4D ones too which could be used to track different layers across a 3D environment etc in a single macro property.

Robadob:
Yeah, I agree yours is far more dynamic, which is always a good thing. I'm just not really 100% up on variadic template soup.

4D makes sense to me in a context of a 3D array of Array variables, so 3D, 1D, rather than outright 4D. E.g. if  a user needed a 3D array of vec3.

MILeach:
Yes exactly re: 4D, alternatively an array of field variables, e.g. if you had 100 different chemical concentrations or something they could be indexed by the first value, then x, y, z.

ptheywood:
> Going to be working on what we've coined 'macro environment' over the next few weeks. 
 Which will combine this and #308.

Is this going to be the only way to have atomically mutable environmental properties? In which case I'm not convinved by the "MacroProperty" name (but I don't have a better suggestion yet).
[...]

PR #643, opened 2021-08-16
Macro Environment

Robadob:
Work in progress support for large environment properties with atomic access from agent functions.

**Progress**
- [x] `EnvironmentDescription` changes
- [x] New `CUDAMacroEnvironment` class
- [x] `CUDASimulation` integration
- [x] `DeviceAPI` access
- [x] Seatbelts support for preventing atomic/read conflicts
- [x] `HostAPI` access
- [x] Submodel mapping support
- [x] Tests
- [x] RTC support
- [x] SWIG support
- [x] SWIG Tests
- [x] docs.fgpu (https://github.com/FLAMEGPU/FLAMEGPU2-docs/pull/60)
    - [x] Document clearly that py doesn't support overiding operator=, so have to use `[0]` or `set()`
- [x] ReadOnly API support (agent function conditions currently have read/write access to macro env).

**Still Requiring Thought**
* Allowing users to default init these properties
* Seatbelts is guaranteed to protect against double use across multiple agentfn in the same layer? 
    * *Yes*
* Does seatbelts detect a host fn, trying to access a property during the same layer?
    * *Host functions are not currently allowed in the same layer as agent functions (or other host functions), although before I realised this I added a checking mechanism anyway.*

**Misc**
* Renamed, `AgentEnvironment` tests to `DeviceEnvironment`, this is more consistent with everything else.

Addresses #162, #308


Robadob:
Code changes should be finished now. Just docu comments (and separate user guide PR) remain.

All python tests pass (Debug + Seatbelts)
image

Robadob:
In writing docs, have noticed a few safety considerations I need to review as I can't recall if they were implemented as fully as possible.

Robadob:
Discussed name change in meeting. Agreed to keep the same. This will mean removing the inheritance between `DeviceAPI` and `ReadOnlyDeviceAPI` to avoid wasting registers by hiding `ReadOnlyDeviceEnvironment` inside regular `DeviceAPI`.

Robadob:
Well I've now split DeviceAPI into mutable/immutable locally (not commit yet), but 5-6 tests are failing in release+seatbelts mode which I need to investigate. (6th being concurrency test)

```
[ RUN      ] DeviceMacroPropertyTest.max2
c:\users\robadob\fgpu2\tests\test_cases\runtime\test_device_macro_property.cu(457): error: Expected: cudaSimulation.simulate() doesn't throw an exception.
  Actual: it throws class flamegpu::exception::DeviceError with description "Device function 'init' reported 1 errors.
First error:
C:/Users/Robadob/fgpu2/include\flamegpu/runtime/utility/DeviceMacroProperty.cuh(191)[0,0,0][0,0,0]:
DeviceMacroProperty read and atomic write operations cannot be mixed in the same layer as this may cause race conditions.
".
[  FAILED  ] DeviceMacroPropertyTest.max2 (3 ms)
[ RUN      ] DeviceMacroPropertyTest.exchange_int32
c:\users\robadob\fgpu2\tests\test_cases\runtime\test_device_macro_property.cu(608): error: Expected: cudaSimulation.simulate() doesn't throw an exception.
  Actual: it throws class flamegpu::exception::DeviceError with description "Device function 'init' reported 1 errors.
First error:
C:/Users/Robadob/fgpu2/include\flamegpu/runtime/utility/DeviceMacroProperty.cuh(191)[0,0,0][0,0,0]:
DeviceMacroProperty read and atomic write operations cannot be mixed in the same layer as this may cause race conditions.
".
[  FAILED  ] DeviceMacroPropertyTest.exchange_int32 (2 ms)
[ RUN      ] DeviceMacroPropertyTest.exchange_uint32
c:\users\robadob\fgpu2\tests\test_cases\runtime\test_device_macro_property.cu(637): error: Expected: cudaSimulation.simulate() doesn't throw an exception.
  Actual: it throws class flamegpu::exception::DeviceError with description "Device function 'init' reported 1 errors.
First error:
C:/Users/Robadob/fgpu2/include\flamegpu/runtime/utility/DeviceMacroProperty.cuh(191)[0,0,0][0,0,0]:
DeviceMacroProperty read and atomic write operations cannot be mixed in the same layer as this may cause race conditions.
".
[  FAILED  ] DeviceMacroPropertyTest.exchange_uint32 (2 ms)
[ RUN      ] HostMacroPropertyTest.ReadTest
c:\users\robadob\fgpu2\tests\test_cases\runtime\test_host_macro_property.cu(76): error: Expected: cudaSimulation.simulate() doesn't throw an exception.
  Actual: it throws class flamegpu::exception::DeviceError with description "Device function 'agentwrite' reported 120 errors.
First error:
C:/Users/Robadob/fgpu2/include\flamegpu/runtime/utility/DeviceMacroProperty.cuh(191)[0,0,0][0,0,0]:
DeviceMacroProperty read and atomic write operations cannot be mixed in the same layer as this may cause race conditions.
".
[  FAILED  ] HostMacroPropertyTest.ReadTest (1 ms)
[ RUN      ] HostMacroPropertyTest.ZeroTest
c:\users\robadob\fgpu2\tests\test_cases\runtime\test_host_macro_property.cu(208): error: Expected: cudaSimulation.simulate() doesn't throw an exception.
  Actual: it throws class flamegpu::exception::DeviceError with description "Device function 'agentwrite' reported 120 errors.
First error:
C:/Users/Robadob/fgpu2/include\flamegpu/runtime/utility/DeviceMacroProperty.cuh(191)[0,0,0][32,0,0]:
DeviceMacroProperty read and atomic write operations cannot be mixed in the same layer as this may cause race conditions.
".
[  FAILED  ] HostMacroPropertyTest.ZeroTest (2 ms)
```

Robadob:
A different mix, are failing with the same message under Debug.

```
[  FAILED  ] DeviceMacroPropertyTest.RTC_add
[  FAILED  ] DeviceMacroPropertyTest.sub
[  FAILED  ] DeviceMacroPropertyTest.sub2
[  FAILED  ] DeviceMacroPropertyTest.postincrement
[  FAILED  ] DeviceMacroPropertyTest.preincrement
[  FAILED  ] DeviceMacroPropertyTest.min
[  FAILED  ] DeviceMacroPropertyTest.min2
[  FAILED  ] DeviceMacroPropertyTest.max
[  FAILED  ] DeviceMacroPropertyTest.max2
```
Unclear where this has come from, after a month on leave, but will investigate.

Robadob:
Appear to have fixed it (Debug all now pass), I wasn't explicitly memsetting device memory allocated for macro properties. So the read/write flag would sometimes begin non-zero.

Robadob:
> * Does seatbelts detect a host fn, trying to access a property during the same layer?
>     * No, there is currently an explicit synchronisation before host functions, but documentation discourages people from writing models that modify in same layer.

I think I will add this exception before I merged the PR, it can always be softened in future if desired.

Robadob:
* I've now added a check for device write ops before host fn, requires a test adding
* I still need to add a check for device read ops before a host write, this is a bit more fiddly
* It appears that host fn writing to macro property before device fn reads back has not been tested, as I can't find where the upload is triggered to breakpoint it to add the read ops check (it will be delayed but this shouldn't really matter).

Robadob:
Well I did all that work, then discovered that `LayerDescription` won't let a host function be in a layer with anything else, which renders it redundant. I'll leave the code in and the tests commented out. But this should be ready to merge after a squash and all tests have been checked.

Robadob:
```
Win, C++, Debug Seatbelts=ON: 945 pass, 0 fail, 38 disabled
Win, C++, Release Seatbelts=ON: 957 pass, 1 fail, 25 disabled (TestCUDASimulationConcurrency.RTCLayerConcurrency failed, poss bc im running py tests in parallel)
Win, C++, Release Seatbelts=OFF: ----
Win, Py, Debug Seatbelts=ON: 543 pass, 0 fail, 4 skipped
Win, Py, Release Seatbelts=ON:  ----
Win, Py, Release Seatbelts=OFF:  ----
```

include/flamegpu/runtime/DeviceAPI.cuh
```diff
@@ -88,6 +88,14 @@ class ReadOnlyDeviceAPI {
         return getVariable<id_t>("_id");
     }
 
+    /**
+     * Access the current stepCount
+     * @return the current step count, 0 indexed unsigned.
+     */
+    __forceinline__ __device__ unsigned int getStepCounter() const {
```
ptheywood, line 95: ~would `getCurrentStep` be a better name? (just a suggestion while viewing the diff)~

Just spotted this was a move not an addition, so probably better to avoid the breaking change.

include/flamegpu/runtime/DeviceAPI.cuh
```diff
@@ -343,6 +431,17 @@ __device__ T ReadOnlyDeviceAPI::getVariable(const char(&variable_name)[M], const
     return value;
 }
 
+template<typename MessageIn, typename MessageOut>
+template<typename T, unsigned int N>
+__device__ void DeviceAPI<MessageIn, MessageOut>::setVariable(const char(&variable_name)[N], T value) {
+    if (variable_name[0] == '_') {
```
ptheywood, line 437: Should this DTHROW if setabelts is on instead?

include/flamegpu/runtime/utility/DeviceMacroProperty.cuh
```diff
@@ -0,0 +1,452 @@
...
+     */
+    __device__ __forceinline__ DeviceMacroProperty<T, J, K, W, 1> operator[](unsigned int i) const;
+    /**
+     * atomic add
+     * @param val The 2nd operand
+     * @return a reference to this
+     * Note, taking value of the returned object will fail, due to the risk of atomic conflicts
+     * @note Only suitable where T is type int32_t, uint32_t, uint64_t, float, double
+     */
+    __device__ __forceinline__ DeviceMacroProperty<T, I, J, K, W>& operator +=(const T& val);
```
ptheywood, line 92: I'm not super keen on these overloads. It hides that atomic ops are being used.

include/flamegpu/runtime/utility/DeviceMacroProperty.cuh
```diff
@@ -0,0 +1,452 @@
...
+     * @return the value before increment operation is performed
+     * @note Only suitable where T is type uint32_t
+     */
+    __device__ __forceinline__ T operator--(int);
+    /**
+     * atomic min
+     * @return min(this, val)
+     * @note Only suitable where T is type int32_t, uint32_t, uint64_t
+     */
+    __device__ __forceinline__ T min(T val);
```
ptheywood, line 146: Again i'm not keen on how it is not immediately clear that this is atomic without reading the docs, but that might just be an artifact of using CUDA too much.

src/flamegpu/model/SubEnvironmentDescription.cpp
```diff
@@ -63,28 +63,90 @@ void SubEnvironmentDescription::mapProperty(const std::string &sub_property_name
...
+    // Sub macro property exists
+    auto subEnv = data->subEnvironment.lock();
+    if (!subEnv) {
+        THROW exception::InvalidParent("SubEnvironment pointer has expired, "
+            "in SubEnvironmentDescription::mapMacroProperty()\n");
+    }
+    const auto subProp = subEnv->macro_properties.find(sub_property_name);
+    if (subProp == subEnv->macro_properties.end()) {
+        const auto parent = data->parent.lock();
+        THROW exception::InvalidEnvProperty("SubModel '%s's Environment does not contain property '%s', "
```
ptheywood, line 76: Should this mention macro early in the exception message? (as in exceptions below?)

swig/python/flamegpu.i
```diff
@@ -12,6 +12,7 @@
 #pragma SWIG nowarn=362
 // operator++ ignored (not supported by python)
 #pragma SWIG nowarn=383
+#pragma SWIG nowarn=384
```
ptheywood, line 15: Comment would be good. is this `operator--`?

swig/python/flamegpu.i
```diff
@@ -524,7 +526,14 @@ class ModelVis;
 
 // Must wrap these prior to HostAPI where they are used to avoid issues with no default constructors etc.
 %include "flamegpu/runtime/utility/HostRandom.cuh"
+
+//%ignore flamegpu::HostEnvironment::getMacroProperty();
```
ptheywood, line 530: If these commented out ignore/renames are not required could they be removed? Or atleast explained why they are still here?

ptheywood (changes requested):
I'm still not convinced by macro property as a name, but still don't have a better solution. 

Otherwise just some fairly trivial changes to exception messages would be good. Might be worth considering seatbelts vs non-seatbelts performance at some point too, as I expect seatbelts to suffer here. 

CUDA C++ tests all pass under linux with CUDA 11.2 and GCC 9. Python tests currently running but looking promising so far (80%).

include/flamegpu/gpu/CUDAMacroEnvironment.h
```diff
@@ -0,0 +1,262 @@
...
+    HostMacroProperty_swig<T> getProperty_swig(const std::string& name);
+#endif
+};
+
+template<typename T, unsigned int I, unsigned int J, unsigned int K, unsigned int W>
+HostMacroProperty<T, I, J, K, W> CUDAMacroEnvironment::getProperty(const std::string& name) {
+    // Validation
+    auto prop = properties.find(name);
+    if (prop == properties.end()) {
+        THROW flamegpu::exception::InvalidEnvProperty("Environment property with name '%s' not found, "
```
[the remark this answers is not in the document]
Robadob: probably

include/flamegpu/gpu/CUDAMacroEnvironment.h
```diff
@@ -0,0 +1,262 @@
...
+// forward declare classes from other modules
+
+namespace flamegpu {
+namespace detail {
+namespace curve {
+class CurveRTCHost;
+}  // namespace curve
+}  // namespace detail
+
+    struct SubEnvironmentData;
```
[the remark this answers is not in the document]
Robadob: Yeah vs has a thing where it introduces phantom spaces, so i'm never sure if the spaces are real or not.

include/flamegpu/runtime/DeviceAPI.cuh
Robadob: Yes, but that's a separate issue affecting whole of DeviceAPI. Will create a separate issue/pr.

include/flamegpu/runtime/utility/DeviceMacroProperty.cuh
Robadob: The understanding should be that anything except reading the value is atomic in an agent fn with macro env prop.

swig/python/flamegpu.i
Robadob: That rings a bell, but I can't confirm at this point.

swig/python/flamegpu.i
Robadob: I'll remove, was probably dev junk from trying to get it to work

PR #644, opened 2021-08-18
Glibc 2.17 wheels

ptheywood:
Enables the creation of python wheels which require glibc >= 2.17. 

This will make wheels which are supported on older linux operating systems such as google collab, and HPC sites. 

+ Uses manylinux2014 images for building glibc 2.17 wheels
  + Wheels are *NOT* manylinux compliant, due to linkage to external shared objects such as `libcuda.so`. The resolution to this is to `dlopen` cuda shared libraries rather than link against them controlled via a CMake flag. This will be handeled in a future PR / Issues.
  + Adds manylinux2014 ci which runs on branch push events and pull requests to ensure support
  + Updates all CI workflows for standardisation and to make distributable wheels
    + Not yet suitable for a package index, but we can potentially provide our own. 
+ If CMake >= 3.18 is used, then CMake will only require `Python3 Development.Module` rather than `Python3 Development`. 
  + This is required for building in Manylinux images
  + CMake 3.18 will soon become our minimum supported version, so the alternate branch can be dropped in the future.
+  Embeds the cuda version used in the python local version number, if a CMake option is set to ON (default). 
  + i.e. `v2.0.0-alpha.1+cuda112` was build with CUDA 11.2. This allows us to differentiate wheels, and provide broader cuda support. 
+ Changes the python distribution name if `VISUALISATION=ON`, to `pyflamegpu-vis`. This allows wheels to be differeniated. 

## Todo

+ [ ] Test the wheels produced on a range of systems. A CI workflow to produce wheels has been initiated on my fork:  [action](https://github.com/ptheywood/FLAMEGPU2/actions/runs/1143389071), [release (when published)](https://github.com/ptheywood/FLAMEGPU2/releases/tag/v0.0.1-alpha)
    + [ ] Windows
    + [ ] Ubuntu 20.04
    + [ ] Ubuntu 18.04
    + [ ] Google Collab
    + [ ] Centos 7 (TuOS HPC)
+ [ ] Update / close relevant issues, Open new specialised issues
    + [ ] Manylinux2014 compliance via `dlopen` for cuda libs, packaging or dlopen for visualisation. 
        + `dlopen` for visualisation would allow a visualisation build to be used on headless machines. Use of the VISUSALISATION macro may need adjusting. 
    + [ ] Hosted html file providing links to Release wheels, as an alternative to pip
    + [ ] Pypi hosting todolist 
    + [ ] Potential Conda packaging
    + [ ] Release publication CI workflow


ptheywood:
Linux console wheels are happy everywhere (if appropriate `libcuda.so` and `libnvrtc.so` are available, `cudart` is currently statically). 
This includes collab and ShARC. 

Linux Vis wheels on Ubuntu 20.04 are unhappy, due to a libGLEW mismatch between the old expected version from CentOS and the newer ubuntu 20.04 version. This will require packaging the `.so` or static linking.

I'll test windows when i'm next booted into windows.

ptheywood:
I've set another demo of the draft release CI going on my fork, which should produce visusalition wheels that use static GLEW to be more portable, but not pack .so files which are driver-dependent. 

Once I've verified that these work as intended, I'm keen for this PR to be merged in as it is a big improvement comapred to what current master produces, even though it is not ready for pypi based distribution.

+ [action](https://github.com/ptheywood/FLAMEGPU2/actions/runs/1163478769)
+ [release when published](https://github.com/ptheywood/FLAMEGPU2/releases/tag/v0.0.2-alpha.1)

The build matrix is still rather large, but that's just how many builds we are intentionally producing. Making portable visualisation wheels which work on headless nodes is a long way off still, if even viable.

ptheywood:
I can confirm that the release linked above generates console and vis wheels which can be ran on Ubuntu 20.04. 

Wheels will run on any linux using glibc 2.7 or newer (i.e. centos 7 for HPC systems).

Visualisation wheels may or may not work depending on the opengl installation, but as these wheels are only going to be useful on proprietary nvidia drivers, things will probably be OK.  This is an improvement compared to the current state, and all I have time to sort in the very immediate future.

All that remains to decide before this is merged is how to reduce the number of checks which run for each commit on pull request branches.

ptheywood:
With removed `pull_request` workflows, and adjusted build matrices this now runs 11 checks:

+ docs
+ lint
+ manylinux wheels, primary wheel CUDA
    + `CUDA 11.2`, `VIS=ON`
    + `CUDA 11.2`, `VIS=OFF`
+ Ubuntu, multiple cudas, one vis build
    + `CUDA 11.4`, `VIS=ON`
    + `CUDA 11.4`, `VIS=OFF`
    + `CUDA 11.0`, `VIS=OFF`
    + `CUDA 10.0`, `VIS=OFF` 
+ Windows, multiple cudas, single vis build
   + `CUDA 11.4`, `VIS =ON`
   + `CUDA 11.4`, `VIS=OFF`
   + `CUDA 11.0`, `VIS=OFF`

When we drop CUDA 10.0, this will be reduced to 10 runs in total. CUDA 12's release could result in more builds, or the same number subject to decisions at the time (min cuda, max cuda 11, max cuda 12?). But that's not going to be an issue for a while (11.5 will be the next release based on public info).

The draft CI workflow will still produce 32 wheels (2 OS * 2 CUDAs * 2 VIS * 4 py), and build the test suite on windows and linux, with old and new cuda (11.0 and 11.4 currently)

I will merge once CI passes.

PR #651, opened 2021-08-18
Avoid CUDA 11.0 + GCC 9 + --std=c++17 compilation segfault

### LINKED ISSUE #650: CUDA 11.0 + GCC 9.x test suite segmentation fault
During compilation of the `tests` target on ubuntu 20.04 systems, using gcc-9 and CUDA 11.0 building the test suite results in a segmentation fault on github actions. 

| Ubuntu | CUDA | GCC | STATUS |
|--------|------|-----|--------|
| 20.04  | 11.0 | 9   | **BAD**|
| 20.04  | 11.0 | 8   | GOOD   |
| 20.04  | 11.0 | 7   | GOOD   |
| 20.04  | 11.1 | 9   | GOOD   |
| 20.04  | 11.1 | 8   | GOOD   |
| 20.04  | 11.1 | 7   | GOOD   |
| 20.04  | 11.4 | 9   | GOOD   |

This may just be a compiler bug / update to GCC 9 on ubuntu 20.04 compared to the officailly supported and tested version. 

GCC 9.3.0 is listed as the supported GCC for CUDA 11.0 in the [release notes](https://docs.nvidia.com/cuda/archive/11.0/cuda-installation-guide-linux/index.html#system-requirements), and `gcc --version` does report `9.3.0` still, so it's not an unsupported configuration.

+ [x] Ensure that CI is using gcc-8 with CUDA 11.0 (#644)  
+ [x] Try to further identify the cause, out of interest and to see if there is a simple resolution.
    + CUDA 11.0 + GCC-9 + `--std=c++17` does not approve of `std::vector<std::tuple<>>::push_back()`
+ [x] Add workarounds to `tests` and `sugarscape` to avoid triggering this issue (#651)
+ [x] Emit a CMake developer warning when configuring with CUDA 11.0 and GCC 9

I've narrowed it down to being an issue when building `test_agent_random.cu`. Disabling the body of the file allows compilation to proceed. 

```bash
[ 51%] Building CUDA object tests/CMakeFiles/tests.dir/test_cases/runtime/test_agent_random.cu.o
Segmentation fault (core dumped)
```

+ The segmentation fault only occurs when `AgentRandomTest.AgentRandomCheck` is being compiled.
  + It occurs within the Test Model X sections, regardless  of which are removed. 
+ Other tests in the file are OK. 

Further narrowed down to:
```
            results2.push_back(std::make_tuple(
                instance.getVariable<float>("a"),
                instance.getVariable<float>("b"),
                instance.getVariable<float>("c")));
```

By splitting this line, it has been further narrowed down to the `push_back()` when providing the tuple.

```
            float a = instance.getVariable<float>("a");
            float b = instance.getVariable<float>("b");
            float c = instance.getVariable<float>("c");
            auto t = std::make_tuple(a, b, c);
            results2.push_back(t); // < -- this line
```

I've further narrowed it down to only occur in CUDA 11.0, with GCC 9 when `--std=c++17`. Using lower standards or a different gcc does not show this issue.

ptheywood:
`std::vector<std::tuple<>>::push_back()` leads to compiler segmentation faults when using CUDA 11.0 with gcc-9 and the `--std=c++17` flag.

This is fixed by using a newer CUDA version, older GCC, or lower `--std`, so not worth a CUDA bug report.

+ [x] Uses `std::vector<std::array<>>` as a hacky workaround
+ [x]  Add a CMake dev warning about this compiler combination, but do not prevent it's use. 

Closes #650

Separately, new CI has been adjusted to not enable this bad compiler combination.

### DIFF
--- a/cmake/cxxstd.cmake
+++ b/cmake/cxxstd.cmake
@@ -60,6 +60,20 @@ if(NOT FLAMEGPU_CXX_STD)
     #     # Forward to the parent scope so it persists between calls.
     #     set(FLAMEGPU_CXX_STD ${FLAMEGPU_CXX_STD} PARENT_SCOPE)
     # endif()
+
+    # Emit a developer warning if using CUDA 11.0 and GCC 9 in c++17 mode re std::vector<std::tuple<...>>::push_back
+    if( CMAKE_CUDA_COMPILER_VERSION VERSION_GREATER_EQUAL "11.0"
+        AND CMAKE_CUDA_COMPILER_VERSION VERSION_LESS "11.1"
+        AND CMAKE_CXX_COMPILER_ID STREQUAL "GNU"
+        AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL "9"
+        AND CMAKE_CXX_COMILER_VERSION VERSION_LESS "10"
+        AND (FLAMEGPU_CXX_STD EQUAL 17 OR CMAKE_CUDA_STANDARD EQUAL 17))
+        # https://github.com/FLAMEGPU/FLAMEGPU2/issues/650
+        message(AUTHOR_WARNING 
+            "CUDA 11.0 with g++ 9 in c++17 mode may encounter compiler segmentation faults with 'std::vector<std::tuple<...>>::push_back'.\n"
+            "Consider using CUDA 11.1+ or gcc 8 to avoid potential issues.\n"
+            "See https://github.com/FLAMEGPU/FLAMEGPU2/issues/650 for more information.")
+    endif()
 endif()
 
 # @future - set this on a per target basis using set_target_properties?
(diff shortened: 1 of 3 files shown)


Robadob:
lol ok

ptheywood:
Added a CMake `AUTHOR_WARNING` which looks like this:
```
cmake .. 
-- -----Configuring Project: flamegpu-----
-- Generating Compute Capabilities: 35;50;60;70;80
-- RapidJSON found. Headers: /home/ptheywood/code/flamegpu/FLAMEGPU2/build/_deps/rapidjson-src/include
CMake Warning (dev) at cmake/cxxstd.cmake:72 (message):
  CUDA 11.0 with g++ 9 in c++17 mode may encounter compiler segmentation
  faults with 'std::vector<std::tuple<...>>::push_back'.

  Consider using CUDA 11.1+ or gcc 8 to avoid potential issues.

  See https://github.com/FLAMEGPU/FLAMEGPU2/issues/650 for more information.
Call Stack (most recent call first):
  cmake/common.cmake:37 (include)
  src/CMakeLists.txt:45 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

Issue #601, opened 2021-07-22
libgtest.a and libtinyxm2.a not in lib/<config>


ptheywood:
CMake is not correctly setting the output location for the static libraries created for googletest and tinyxml2 correctly. 

`libgtest.a` and  `libtinyxml2.a` are being output into `lib/` rather than `lib/Release` for instance (although there is also a copy of `libtinyxml2.a`  in `lib/Release` alongside the correctly positioned `libflamegpu2.a`.

I believe the solution is to adjust how we are setting `LIBRARY_OUTPUT_DIRECTORY`, which is currently being defined in many places afaik (with logic based on which is tha parent). 

Once adjusted this needs checking with:

+ main repo `build`
+ individual example in main repo
+ Standalone example

PR #653, opened 2021-08-18
Change how the bin/lib paths are set throughout CMakeLists

ptheywood:
Change how the bin/lib paths are set throughout CMakeLists

This should resovle #601, but standalone repos and the visualisation may also benefit from being updated.

## Todo
+ [ ] Testing
+ [ ] Standalone repos (after this is merged?)


ptheywood:
On Windows this has led `_pyflamegpu.pyd` to be placed in `/bin/Release/`, as it's essentailly a `.dll`. This is probably not  the correct location? 

Each example is producing a `.lib` and `.exp` file which are placed in `lib/Release` I don't think these are something we are intentionally producing? I was only epxecting to see `tinyxml2.lib`, `gtest.lib`, `pyflamegpu.lib` (and maybe `.exp`) and `python/ in there. 

From the output of `cmake --build` it  does state that it's creating a library and an object:

```
 main.obj
     Creating library C:/Users/ptheywood/code/flamegpu/FLAMEGPU2/build/lib/Release/ensemble.lib and object C:/Users/ptheywood/code/flamegpu/FLAMEGPU2/build
  /lib/Release/ensemble.exp
  ensemble.vcxproj -> C:\Users\ptheywood\code\flamegpu\FLAMEGPU2\build\bin\Release\ensemble.exe
  tmpxft_000014ac_00000000-7_main.cudafe1.cpp
```

Current master produces the .lib still, but places it deeply in the build directory, so this is probably an unwanted change?
```
     Creating library C:/Users/ptheywood/code/flamegpu/FLAMEGPU2/build/examples/game_of_life/Release/game_of_life.lib and object C:/Users/ptheywood/code/fl
  amegpu/FLAMEGPU2/build/examples/game_of_life/Release/game_of_life.exp
```

On windows previsusly the values of `CMAKE_ARCHIVE_OUTPUT_DIRECTORY` and `LIBRARY` were unset, while binary was being manually set. Per target properties are inheritted from this manual setting. 

Maybe we should be setting per target ourselves, depending on the type of target?

PR #662, opened 2021-08-24
Fix out of source builds

### LINKED ISSUE #600: Build Metadata in `src` not `include`
Currently, the version build metadata (i.e. commit hash) is stored in `include/flamegpu/version.h`, which is modified at every cmake configuration event. 

This breaks out of tree builds, but is a current neccesary evil to minimise the use of the incorrect include directory for RTC compilation (potentially only for non-python RTC?)

Instead, it would be preferable if `flamegpu::VERSION_BUILDMETADATA` was declared in `version.h` and defined in a dynamic file in the build directory (i.e. `build/flamegpu/version.cpp`), so that the include directory is not modified by cmake. 
This will prevent the use of it while checking the include directory version, instead the major, minor and prerelease version would need to be sufficient. This would be OK for users of stable releases, but may be insufficient during development.

One possible workaround for this would be to add a new CMake option, or CLI/environment varible option to ignore the jitify cache, to avoid incorrect cache hits, however it might still cause some issues during development. 
Another solution would be to dynamically create `version.h` in the build directory, pulling in a copy of the headers and using that as the `FLAMEGPU_INC_DIR`?

It's unclear what is the best option.

+ [ ] Stop `include/flamegpu/version.h` from being generated by CMake
    + [ ] Encode the version information in the header manually, and extract into CMake. Inverting the relationship.
    + [ ] Declare but do not define `static const char VERSION_BUILDMETADATA[];`
    + [ ] Adjust RTC include directory checking to not consider build metadata (subject to any alternatives found). Only major/minor/patch/prerelease should  be sufficient?
+ [ ] Generate `build/flamegpu/version.cpp` via CMake at configure time, including the  git commit hash as build meta data, embedding the build metadata in the static library.

ptheywood:
Embded version build metadata in version.cpp rather than version.h

+ Semver release version is manually embedded within include/flamegpu/version.h (Major, minor, patch, prerelease)
+ CMake generates <build>/FLAMEGPU/src/version.cpp, including the buildmetadata (commit hash) and version strings which may include the commit hash
+ When Cmake is reconfigured, version.cpp is updated with the up to date build metadata, and flamegpu will rebuild
    + This will only rebuild the version.cpp.o, rather than anything which includes version.h
+ The RTC header version check now compares the macro FLAMEGPU_VERSION, and the prerelease version from flamegpu::VERSION_PRERRLEASE
    + Metadata is no longer encoded in the include directory so this cannot be checked for a match.
    + Metadata is still factored into the jitify cache, to catch metadata changes.

Closes #600 

## Todo

+ [ ] Run the test suite on windows and linux
+ [ ] Decide if we want an opt-in via CMake option to enforce include version metadata?
    + I.e. `-DEMBED_VERSION_METADATA_INCLUDE` or something to that effect, defaults to `OFF`
    + Output build meta data to a non source file in the include directory, containing the build metadata, configured by CMake when git has been modified since the last time the file was configured
    + Add a preprocessor definition which instructs the library to read this information (or always  check if the file exists?) 
    + In `JitifyCache::confirmFLAMEGPUHeaderVersion`, read the metadata from this file (if it exists) and compare. Throw an exception if it does not. 
    + This will *only* be useful during RTC header development within a given release, to catch if the include directory is a mismatch. This should not be default, and should not be built into any binary release artifacts.


ptheywood:
For now, merge once tested. 

If not comparing hashes becomes an issue during dev, add the flag and extra check.

PR #690, opened 2021-09-08
Ignore Simulation::getSimulationConfig from python/swig wrapping

### LINKED ISSUE #689: Swig Ignore Simulation::getSimulationConfig
I missed this as part of #665, as I was expecteding `getConfig` not `getSimulationConfig`. 

This should not be acecssible from python as it's marked const, but in python it's still mutable, so changes made are not applied to the underlying c++.

`swig/python/flamegpu.i:362`

```c++
// %ignore flamegpu::Simulation::getConfig; // This doesn't currently exist`
```

Needs uncommenting and adjusting.

Ideally add a test to make sure it's disabled to the appropraite test suite (CUDASimulation as Simulation is virtual so cannot be tested without an implementation)

ptheywood:
Closes #689


ptheywood:
Full python test suite passes under linux

`510 passed, 4 skipped in 479.88s (0:07:59)`

PR #693, opened 2021-09-22
NVHPC Support

ptheywood:
+ [x] Patch RapidJSON if using nvhpc
+ [x] Change warning levels / suppressions if using nvhpc. It does not support isystem unfortunately.
+ [ ] Fix `std::experimental::filesystem` linker errors.
    + This may require C++ 17 support only / may be a blocking issue.
+ [ ] NVHPC CI?
    + This might not be viable, as an NVHPC install is > 8GB, and unsure on the licencing. Using a container may be an option, but this might want to be a less-regular CI job?

### DIFF
--- a/cmake/dependencies/patches/rapidjson-nvhpc.patch
+++ b/cmake/dependencies/patches/rapidjson-nvhpc.patch
@@ -1,0 +1,26 @@
+diff --git a/include/rapidjson/internal/biginteger.h b/include/rapidjson/internal/biginteger.h
+index 8eb87c7c..5db5a2eb 100644
+--- a/include/rapidjson/internal/biginteger.h
++++ b/include/rapidjson/internal/biginteger.h
+@@ -252,7 +252,7 @@ private:
+         if (low < k)
+             (*outHigh)++;
+         return low;
+-#elif (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 6)) && defined(__x86_64__)
++#elif (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 6)) && defined(__x86_64__) && !defined(__NVCOMPILER)
+         __extension__ typedef unsigned __int128 uint128;
+         uint128 p = static_cast<uint128>(a) * static_cast<uint128>(b);
+         p += k;
+diff --git a/include/rapidjson/internal/diyfp.h b/include/rapidjson/internal/diyfp.h
+index 8f7d853a..68039c23 100644
+--- a/include/rapidjson/internal/diyfp.h
++++ b/include/rapidjson/internal/diyfp.h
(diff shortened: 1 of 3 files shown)


ptheywood:
Not actualyl blocked, by `std::experimental::filesystem` - it just uses gcc's stdlib so can pass the appropriate linker args.

Mostly seems to work, other than release RTC test suite failures. Debug is fine which makes tracing the fault more interesting. Release RTC examples also work, so it's test suite specific in some way?

Vis works, but the vis repo needs CMake changes to address warnings (the same as the main repo + some extras).

ptheywood:
If using GCC 8's stdlib rather than GCC 9'this builds ok in Release mode (nvc++ 21.7-0, ubuntu 21.04).

ptheywood:
Currently working on reproducing this with a simpler use case. Currently leaning towards or more of the following:

+ Compiler error (as this only appears to be effected by nvhpc + gcc 9)
+ UB related to the use of `enable_shared_from_this` which the docs for list several opportunities for UB.
+ Unintialised member variables which are potentially present for many of tha *Data classes. 

Will conitnue to work on the MWE a little, but if it doesn't reproduce soon it'll just get dumped into a gist for future reference. Running gcc and nvhpc builds through valgrind (with an appropriate cuda suppressions list) would be good and generally worthwhile on the whole.

ptheywood:
NVHPC repackages the location of curand compared to standalone nvcc. Up to some point between nvhpc 21.11 and 22.3 this is not correctly reflected by the include path during compilation via cmake.

We may be able to resoilve this by requiring curand as a dependency in cmake, otherwise we might need to expliclty add an edge case to cmake to ensure this include path is set.

ptheywood:
after some horrible cmake additions to explicitly add the non-symlinkg math_libs include directory to include path(s) if required, curand is now found when using nvhpc installed nvcc, and nvhpc as the host compiler. 

However, this then exposes an issue with include path ordering and the finding / use of cub and thrust. 

The cub/thrust version mismatch check is identifying that they do not agree. locally using a cuda 11.8 nvhpc 22.11 which ships with cub/thrust 1.15, this is conflicting with the explicitly added cub/thrust 1.17 we fetch. 

This will be due to include directories and precedent. It might not be the case for all cmake/nvhpc combos, so i will force CI to investigate for me.

some commits are wip, as cmake 3.18 needs to use a differnet method for symlink resolution compared to 3.19, which is not tested.

ptheywood:
some NVHPC builds via containers which fail to configure CMake are erroring due to:

```
2024-01-11T18:00:25.4234501Z     #error -- unsupported pgc++ configuration! Only pgc++ 18, 19, 20 and 21 are supported! The nvcc flag '-allow-unsupported-compiler' can be used to override this version check; however, using an unsupported host compiler may cause compilation failure or incorrect run time execution. Use at your own risk.
```

This is with a version of nvcc distributed with the version of nvhpc which is apparently incompatible.

PR #702, opened 2021-10-06
Add README note RE CUDA/VS install order.

Robadob:


ptheywood (commented):
This might be better as a nested bullet below the Visual Studio bullet point in the dependencies section.

PR #714, opened 2021-10-20
Allow glm to be used via HostApi::Environment::getProperty()

Robadob:
Requires the USE_GLM configuration to be enabled.

These changes are required to support the FGPU2 Primage model (it's currently pointed at this branch).

As the model now works and is validated it's probably suitable to be merged.
Unclear if we want explicit tests for GLM support everywhere?

---------

This branch will need squashing before merge (or rebranch and squash), as the primage model is currently pointed at it and i'd rather minimise needing to keep asking kenneth to `git fetch`, `git reset --hard origin/host_env_glm`.


mondus (commented):
Happy with this so long as tests still pass.

mondus (approved):
Happy with this so long as tests still pass.

PR #723, opened 2021-10-27
Agent Sorting

MILeach:
This PR includes support for sorting agents which use spatial messaging by their position to accelerate message read times.

- [x] Identify where sorting should take place
- [x] Explore options for sorting - manual vs host API
- [x] Automatically detect/create necessary variables
- [x] Warning if required environment vars for sorting aren't set
- [x] Implement sort for default state agents
- [x] Add interface for controlling sorting frequency
- [x] Support all agent states
- [x] Move sort from beginning of step to most optimal point, i.e. before read takes place per agent function
- [x] Add required method to message types to query whether sorting is required
- [x] Visual inspection to confirm behaviour is still correct
- [x] Add unit tests

Initial results show that the ideal sorting frequency depends on the model, both boids and circles have achieved a runtime improvement in the range of 20%-40% depending on sort frequency.

### DIFF
--- a/examples/circles_spatial3D/src/main.cu
+++ b/examples/circles_spatial3D/src/main.cu
@@ -186,6 +186,7 @@ int main(int argc, const char ** argv) {
     /**
      * Execution
      */
+    cudaSimulation.setSortAgentsEveryNSteps(10);
     cudaSimulation.simulate();
 
     /**
--- a/examples/circles_spatial3D/src/main.cu
+++ b/examples/circles_spatial3D/src/main.cu
@@ -186,7 +186,6 @@ int main(int argc, const char ** argv) {
     /**
      * Execution
      */
-    cudaSimulation.setSortAgentsEveryNSteps(10);
     cudaSimulation.simulate();
 
     /**
--- a/examples/circles_spatial3D/src/main.cu
+++ b/examples/circles_spatial3D/src/main.cu
@@ -94,6 +94,11 @@ int main(int argc, const char ** argv) {
     {
         flamegpu::EnvironmentDescription  &env = model.Environment();
         env.newProperty("repulse", 0.05f);
+
+        // Setting these allows us to take advantage of the automatic spatial sorting which considerably increases performance
+        env.newProperty("MIN_POSITION", -static_cast<float>(floor(cbrt(AGENT_COUNT))));
+        env.newProperty("MAX_POSITION", static_cast<float>(floor(cbrt(AGENT_COUNT))));
+        env.newProperty("INTERACTION_RADIUS", RADIUS);
     }
 
     /**
--- a/examples/circles_spatial3D/src/main.cu
+++ b/examples/circles_spatial3D/src/main.cu
@@ -94,11 +94,6 @@ int main(int argc, const char ** argv) {
     {
         flamegpu::EnvironmentDescription  &env = model.Environment();
         env.newProperty("repulse", 0.05f);
-
-        // Setting these allows us to take advantage of the automatic spatial sorting which considerably increases performance
-        env.newProperty("MIN_POSITION", -static_cast<float>(floor(cbrt(AGENT_COUNT))));
-        env.newProperty("MAX_POSITION", static_cast<float>(floor(cbrt(AGENT_COUNT))));
-        env.newProperty("INTERACTION_RADIUS", RADIUS);
     }
 
     /**
--- a/include/flamegpu/gpu/CUDASimulation.h
+++ b/include/flamegpu/gpu/CUDASimulation.h
@@ -439,13 +439,7 @@ class CUDASimulation : public Simulation {
     /**
      * How often to sort spatial agents
      */
-    unsigned int sort_agents_every_n_steps = 10;
-
-    /**
-     * Memory used in sorting - TODO: remove and use pre-existing cub temp
-     */
-    void* d_temp_storage;
-    size_t tempBytes = 0;
+    unsigned int sortAgentsEveryNSteps;
 
     /**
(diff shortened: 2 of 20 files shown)


mondus (changes requested):
Suggest that each of the unit tests gives a descriptive comment to what the test is doing.

src/flamegpu/model/AgentFunctionDescription.cpp
```diff
@@ -519,6 +519,12 @@ AgentFunctionDescription& AgentDescription::newRTCFunction(const std::string& fu
                 std::string code_func_name = match[1];  // not yet clear if this is required
                 std::string in_type_name = match[2];
                 std::string out_type_name = match[3];
+                if (in_type_name == "flamegpu::MessageSpatial3D" || in_type_name == "flamegpu::MessageSpatial2D" || out_type_name == "flamegpu::MessageSpatial3D" || out_type_name == "flamegpu::MessageSpatial2D") {
+                    printf("Adding var\n");
```
Robadob, line 523: remove?

examples/circles_spatial3D/src/main.cu
```diff
@@ -94,6 +94,11 @@ int main(int argc, const char ** argv) {
     {
         flamegpu::EnvironmentDescription  &env = model.Environment();
         env.newProperty("repulse", 0.05f);
+
+        // Setting these allows us to take advantage of the automatic spatial sorting which considerably increases performance
```
Robadob, line 98: Why can't these be pulled directly from the message list? (e.g. lines 77-79)

include/flamegpu/gpu/CUDASimulation.h
```diff
@@ -125,6 +126,11 @@ class CUDASimulation : public Simulation {
      * @return False if an exit condition was triggered
      */
     void exitFunctions() override;
+    /**
+     * The simulation will sort agents periodically.
+     * @param n Agents will be sorted every n steps. 0 indicates no sorting
+     */
+    void setSortAgentsEveryNSteps(const unsigned int n);
```
Robadob, line 133: Is there a reason this isn't inside the `ModelDescription` hierarchy on a per-agent basis?

include/flamegpu/gpu/CUDASimulation.h
```diff
@@ -247,14 +253,18 @@ class CUDASimulation : public Simulation {
...
      * Get the duration of an individual step in seconds.
      * @param step Index of step, must be less than the number of steps executed.
      * @return elapsed time of required step in seconds
      */
     double getElapsedTimeStep(unsigned int step) const;
 
+    /**
+     * Determines which agents require sorting - only used once during initialisation. Must be called manually if not using .simulate()
+     */
+    void determineAgentsToSort();
```
Robadob, line 266: I really think this should be hidden from the user. Not looked at it's implementation yet, but I presume it can be built into the initialisation of `CUDASimulation`, regardless of calls to `simulate()`.

include/flamegpu/model/AgentDescription.h
```diff
@@ -292,6 +292,10 @@ void AgentDescription::newVariable(const std::string &variable_name, const std::
         THROW exception::ReservedName("Agent variables cannot be named 'name' or 'state', these are reserved for backwards compatibility reasons, "
             "in AgentDescription::newVariable().");
     }
+    if (lower_variable_name == "fgpu2_reserved_bin_index") {
```
Robadob, line 295: Reserved variables should begin with `_`

Robadob: @ptheywood also wanted to move to using `constexpr` for storing string literals?

include/flamegpu/gpu/CUDASimulation.h
Robadob: Yeah, having looked at this, i don't see why it can't simply be at the end of the constructor or similar.

src/flamegpu/gpu/CUDASimulation.cu
```diff
@@ -277,6 +278,149 @@ void CUDASimulation::exitFunctions() {
...
+
+    unsigned int sm_size = 0;
+    unsigned int streamIdx = 0;
+#if !defined(SEATBELTS) || SEATBELTS
+    auto *error_buffer = this->singletons->exception.getDevicePtr(streamIdx, this->getStream(streamIdx));
+    sm_size = sizeof(error_buffer);
+#endif
+
+    // Launch kernel
+    calculateSpatialHash<<<gridSize, blockSize, sm_size, this->getStream(streamIdx) >>> (reinterpret_cast<float*>(xPtr),
```
Robadob, line 411: I thought sorting was going to use the pre-existing spatial hash from agents outputting a message?

Robadob (commented):
I might be on the wrong page, but I think some of this needs discussion at the next meeting.

include/flamegpu/model/AgentDescription.h
ptheywood: `constexpr` is better than `const` for string literals, though most of the benefits aren't available until C++17 (or 20+ really, but that's not happening any time soon). 

Also strip out `fgpu2` from the var name please

src/flamegpu/gpu/CUDASimulation.cu
ptheywood: Would the pre-existing hash be available in the append case? (I'm just stabbing in the dark, i've not looked into this at all)

Robadob: Without looking at code, I think it should be in the spare scan/scatter buffers during message output. It may get sorted along with the messages, which would require agents to be sorted with their messages.

Robadob (approved):
I'm happy, Paul had a few comments at the meeting?

mondus (approved):

PR #730, opened 2021-10-29
Bugfix: CUDAEnsemble would always output both step/exit files.

### LINKED ISSUE #731: Tests: Ammend LoggingTests to check for disk output
The current `LoggingTest` explicitly do not check that output to disk has worked as expected.

This allowed a bug where if step but not exit logs are requested, both files were created (see #730)

`test_logging.cu` should include: 

+ Request Step log, only step log file expected (not exit or agent)
+ Request Exit log, only exit log file expected (not step or agent)
+ Request Agent log, only agent log file expected (not step or exit)
+ Request multiple types, expecting multiple types.

Ideally checkign for both ensembles and simulations.

This should just require ammending existing tests to explicitly check, and dupe + modify some of these to cover single-file tests.

It might be best/easiest to wait until C++14 support is dropped (#611), to simplify the implementation via `std::filesystem` rather than having to deal with using `std::experimental::filesystem` and the equivalent MSVC TS version.

Robadob:
Very minor bug. If a user requests step log files from a `CUDAEnsemble`, basically empty exit log files would always be created too. I presume the opposite is also true (or rather specifying an output dir gets you log files).

Closes #731

### DIFF
--- a/include/flamegpu/sim/SimLogger.h
+++ b/include/flamegpu/sim/SimLogger.h
@@ -28,6 +28,8 @@ class SimLogger {
      * @param log_export_queue The queue of logs to exported to disk
      * @param log_export_queue_mutex This mutex must be locked to access log_export_queue
      * @param log_export_queue_cdn The condition is notified every time a log has been added to the queue
+     * @param _export_step If true step logs will be exported
+     * @param _export_exit If true exit logs will be exported
      */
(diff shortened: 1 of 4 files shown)


Robadob:
Have added an extra commit which addresses #731, the new tests all pass locally (on Windows).

PR #734, opened 2021-11-16
Bugfix, logging variable mean for an empty agent pop

Robadob:
would lead to a result of NaN, which JSON exported as empty string (invalid JSON).

Also adds a simple test which checks the result is 0.

### DIFF
--- a/tests/test_cases/io/test_logging.cu
+++ b/tests/test_cases/io/test_logging.cu
@@ -323,6 +323,42 @@ FLAMEGPU_INIT_FUNCTION(logging_ensemble_init) {
         instance.setVariable<unsigned int>("uint_var", static_cast<unsigned int>(i+2));
     }
 }
+TEST(LoggingTest, EmptyMean) {
+    /**
+     * There was a bug, where if an agent population died out
+     * Reductions would log in the form `{"mean":}`, rather than `{"mean":0}`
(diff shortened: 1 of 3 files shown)


tests/test_cases/io/test_logging.cu
```diff
@@ -323,6 +323,42 @@ FLAMEGPU_INIT_FUNCTION(logging_ensemble_init) {
...
+    CUDASimulation sim(m);
+    sim.setStepLog(slcfg);
+    sim.SimulationConfig().steps = 5;
+    // sim.SimulationConfig().step_log_file = "valid.json"; // Enabling this produces the bugged output file
+    sim.simulate();
+    auto& sl = sim.getRunLog().getStepLog();
+    for (auto& step : sl) {  // Check step log doesn't contain NaN
+        EXPECT_EQ(step.getAgent(AGENT_NAME1).getMean("float_var"), 0.0);
+    }
+}
```
Robadob, line 361: Note, I didn't put test at the end of the file, in hopes of avoiding a conflict with my other logging pr.

ptheywood (approved):
Looks good to me, tests pass under linux.

Robadob:
Added an explicit flush to `CUDAEnsemble` progress too, to fix that rather than creating a whole separate commit/PR.

PR #736, opened 2021-11-18
Fix memory allocation logic in CUDAScatter::arrayMessageReorder

### LINKED ISSUE #735: CUDAScatter::scatter arrayMessageReorder error
When using arrary messaging in the sugarscape example, using a `49 x 49` array an error occurs in the second iteration. 

I've narrowed this down to (probably) being due to a an out of bounds device write on a hosttodevice copy  due to some incorrect logic on when to resize or how many bytes to copy. 

https://github.com/FLAMEGPU/FLAMEGPU2/blob/ca7dad17bc49b46edc6470b8f55959bef85eb1c4/src/flamegpu/gpu/CUDAScatter.cu#L554-L581

In sugarscape 49x49, on the 0th iteration CUDAScatter::arrayMessageReorder ends up getting called twice. 

On the first invocation, after CUB is used to find the temp bytes required for a deviceReduceMax is called (which is only actually required if SEATBELTS is enabled):

+ `sd.size()` == `3`
+ `t_data_len` == `1`
+ `streamResources[streamResourceId].data_len` == `0`
+ `streamResources[streamResourceId].data_len * sizeof(ScatterData)` == `0`
+ `sd.size() * sizeof(ScatterData)` == `72`
+ `ScatterData` is a struct containing a `size_t`, and two `char*`, with `sizeof(ScatterData)` equal to `24` bytes.

This means that the `t_data_len > streamResources...` condition passes (`1 > 0`), and then the else condition of the inner loop passes as `sd.size()` is bigger, so `streamResources[streamResourceId]` is resized to be large enough for `3` ScatterData elements (`72` bytes)

At this point, the memcpy is issued requesting a 72 byte copy, which is OK.

On the second call to this method in the 0th iteration, at after the DeviceReduce::Max:

+ `sd.size()` == `4`
+ `t_data_len` == `1`
+ `streamResources[streamResourceId].data_len` == `3`
+ `streamResources[streamResourceId].data_len * sizeof(ScatterData)` == `72`
+ `sd.size() * sizeof(ScatterData)` == `96`

`t_data_len` is less then the data_len * sizeof(ScatterData), so the if condition does not pass and a resize is not attempted, leaving the `streamResources[streamResourceId].data_len` with a value of 3, and the array having a length of 72 bytes.

**But** `sd.size()` is 4, so a copy of 96 bytes is requested into a 72 byte pointer

So the resize logic needs correcting, to consider if sd_size is larger than `streamResources[streamResourceId].data_len` or not, and if SEATBELTS is enabled also needs to consider the type bytes required by CUB.

## Todo

+ [ ] Add a test to the test suite which triggers this.
  + I think the conditions to trigger this are:
    + Small enough grid dimenstions to cause 1 byte required by `cub DeviceReduce::Max`
    + 2 array message lists, with differing numbers of message variabels (i.e. 3 and 4 in sugarscape). 
    + The list with fewer variables needs reordering before the larger list, so 2 or 4 agent fns depending on when reorder happens)
+ [ ] Adjust the resize logic so that it:
  + calculates the number of bytes that are required (`std::max(sd.size() * sizeof(ScatterData), t_data_len)`, also only find t_data_len if SEATBELTS is enabled.)
  + calcualtes the number of bytes allocated (`streamResources[streamResourceId].data_len * sizeof(ScatterData)`)
  + if not enough bytes are allocated, allocate enough bytes. This is done in multiple of `sizeof(ScatterData)`, i.e. `bytes_required / sizeof(ScatterData) + 1`.

ptheywood:
Closes #735 

+ [x] Reimplement / fix memory reallocation logic
+ [x] Write one or more tests that would have triggered this bug
+ [x] Add missing fflush to ensemble progress output

### DIFF
--- a/src/flamegpu/gpu/CUDAScatter.cu
+++ b/src/flamegpu/gpu/CUDAScatter.cu
@@ -564,17 +564,19 @@ void CUDAScatter::arrayMessageReorder(
     }
     assert(d_position);  // Not an array message, lacking ___INDEX var
     size_t t_data_len = 0;
-    {  // Decide curve memory requirements
+    {  // Decide per-stream resource memory requirements based on curve data, and potentially cub temp memory
+#if !defined(SEATBELTS) || SEATBELTS
+        // Query cub to find the number of temporary bytes required.
         gpuErrchk(cub::DeviceReduce::Max(nullptr, t_data_len, d_write_flag, d_position, array_length, stream));
-        if (t_data_len > streamResources[streamResourceId].data_len * sizeof(ScatterData)) {
-            // t_data_len is bigger than current allocation
-            if (t_data_len > sd.size() * sizeof(ScatterData)) {
-                // td_data_len is bigger than sd.size()
-                streamResources[streamResourceId].resize(static_cast<unsigned int>((t_data_len / sizeof(ScatterData)) + 1));
-            } else {
-                // sd.size() is bigger
-                streamResources[streamResourceId].resize(static_cast<unsigned int>(sd.size()));
-            }
+#endif
+        // the num bytes required is maximum of the per message variable ScatterData and the cub temp bytes required
+        const size_t srBytesRequired = std::max(sd.size() * sizeof(ScatterData), t_data_len);
+        // The number of bytes currently allocated
+        const size_t srBytesAllocated = streamResources[streamResourceId].data_len * sizeof(ScatterData);
+        // If not enough bytes are allocated, perform an appropriate resize
+        if (srBytesRequired > srBytesAllocated) {
+            const size_t elementsRequired = ((srBytesRequired - 1) / sizeof(ScatterData)) + 1;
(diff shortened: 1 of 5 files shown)


Robadob (commented):
Ok this makes total sense. Temp buffer space for scatter data wasn't being checked/resized/allocated.

If I understand correctly, the bug was only present with `SEATBELTS` off?

I'm fairly sure the cub::max operation is only required with `SEATBELTS` on, as an extra operation is ran to check that no array element had multiple messages sent to it. So assessing the cub::max requirements all the time is probably unnecessary (but harmless/cheap?).

**Edit:** Yes cub  max is only used with seatbelts, it's a few lines below the snippet.

Robadob (approved):

ptheywood:
Tests under linux Release all pass, + new tests pass for debug.

+ `SEATBELTS=OFF`: `[  PASSED  ] 845 tests.`
+ `SEATBELTS=ON `: `[  PASSED  ] 965 tests.`

PR #739, opened 2021-11-30
Fix avoidance behaviour, add wall avoidance, enforce min speed, sensi…

MILeach:
…ble params

### DIFF
--- a/examples/boids_spatial3D/src/main.cu
+++ b/examples/boids_spatial3D/src/main.cu
@@ -154,11 +154,10 @@ FLAMEGPU_AGENT_FUNCTION(inputdata, flamegpu::MessageSpatial3D, flamegpu::Message
     float global_velocity_y = 0.0f;
     float global_velocity_z = 0.0f;
 
-    // Boids short range avoidance centre
-    float collision_centre_x = 0.0f;
-    float collision_centre_y = 0.0f;
-    float collision_centre_z = 0.0f;
-    int collision_count = 0;
+    // Total change in velocity
+    float velocity_change_x = 0.f;
+    float velocity_change_y = 0.f;
+    float velocity_change_z = 0.f;
 
     const float INTERACTION_RADIUS = FLAMEGPU->environment.getProperty<float>("INTERACTION_RADIUS");
     const float SEPARATION_RADIUS = FLAMEGPU->environment.getProperty<float>("SEPARATION_RADIUS");
@@ -174,7 +173,7 @@ FLAMEGPU_AGENT_FUNCTION(inputdata, flamegpu::MessageSpatial3D, flamegpu::Message
             // Check interaction radius
             float separation = vec3Length(agent_x - message_x, agent_y - message_y, agent_z - message_z);
 
-            if (separation < (INTERACTION_RADIUS)) {
+            if (separation < INTERACTION_RADIUS) {
                 // Update the percieved centre
                 perceived_centre_x += message_x;
                 perceived_centre_y += message_y;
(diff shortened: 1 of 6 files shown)


examples/boids_spatial3D/src/main.cu
```diff
@@ -195,62 +200,55 @@ FLAMEGPU_AGENT_FUNCTION(inputdata, flamegpu::MessageSpatial3D, flamegpu::Message
...
-        const float MATCH_SCALE = FLAMEGPU->environment.getProperty<float>("MATCH_SCALE");
-        match_velocity_x = global_velocity_x * MATCH_SCALE;
-        match_velocity_y = global_velocity_y * MATCH_SCALE;
-        match_velocity_z = global_velocity_z * MATCH_SCALE;
-    }
-    velocity_change_x += match_velocity_x;
-    velocity_change_y += match_velocity_y;
-    velocity_change_z += match_velocity_z;
-
-    // Rule 3) Avoid close range neighbours (Separation)
```
ptheywood, line 241: Rule 3 has just been entirely removed / replaced with just wall avoidance not avoidance of others?

examples/boids_spatial3D/src/main.cu
```diff
@@ -266,6 +264,36 @@ FLAMEGPU_AGENT_FUNCTION(inputdata, flamegpu::MessageSpatial3D, flamegpu::Message
...
+        // Normalise
+        vec3Div(agent_fx, agent_fy, agent_fz, agent_fscale);
+        // Scale to min
+        vec3Mult(agent_fx, agent_fy, agent_fz, minSpeed);
+    }
+   
+    // Steer away from walls - Computed post normalization to ensure good avoidance. Prevents constant term getting swamped
+    float wallInteractionDistance = 0.10f;
+    float wallSteerStrength = 0.05f;
+    if (agent_x - FLAMEGPU->environment.getProperty<float>("MIN_POSITION") < wallInteractionDistance) {
```
ptheywood, line 278: I'd suggest storing `MIN_POSITION` and `MAX_POSITION` in `const float`, both to improve readability and just incase the compiler doesn't realise it can safely remove multiple accesses through curve (It should but who knows what it will do to register usage in 11.3+)

examples/boids_spatial3D/src/main.cu
```diff
@@ -300,25 +328,25 @@ int main(int argc, const char ** argv) {
         env.newProperty("POPULATION_TO_GENERATE", 32768u);
 
         // Environment Bounds
-        env.newProperty("MIN_POSITION", -0.5f);
-        env.newProperty("MAX_POSITION", +0.5f);
+        env.newProperty("MIN_POSITION", -1.5f);
```
ptheywood, line 331: This was previously -.5 to .5 to have a 1 unit cube of volume centred on zero. Might be worth checking with @mondus on his preference (via slack?)

MILeach: Yes this along with the other parameter changes need some thought. This volume provides a sensible density for the number of agents present. The 1 unit cube essentially creates one giant flock immediately as there isn't enough room. The other parameter changes have similar impact.

examples/boids_spatial3D/src/main.cu
MILeach: Rule through has moved into the message read loop as it is now dealing with individual agents rather than an aggregate value. I have added a comment to identify it.

examples/boids_spatial3D/src/main.cu
Robadob: Can't you instead change the interaction radius, rather than the env? Or would this mean other weights need changing?

Robadob: (My personal preference however is to make all units roughly equivalent to metres, normalised is hard to work with)

MILeach: Possibly, the other weights have been changed already anyway. Switching to metres would definitely necessitate increasing the env size though, would have to get very creative to fit 32k birds in a 1m box :D

Robadob: Maybe Reynolds was mistaken and it's a flock of flies.

Robadob:
You rly need to rebase this branch against the force pushed master. Super hard to make sense of the diff (I was trying to check if you'd updated all of the boids examples).

Robadob (approved):
Looks good to me.

I haven't personally built/ran the code.

examples/swig_boids_spatial3D/boids_spatial3D.py
```diff
@@ -165,7 +165,7 @@ def clampPosition(x, y, z, MIN_POSITION, MAX_POSITION):
 // Agent function
 FLAMEGPU_AGENT_FUNCTION(inputdata, flamegpu::MessageSpatial3D, flamegpu::MessageNone) {
     // Agent properties in local register
-    const flamegpu::id_t id = FLAMEGPU->getID();
+    const int id = FLAMEGPU->getID();
```
ptheywood, line 168: Why have you changed this from `id_t`, which the method returns? 

Looks like it might have been to standardize it with the C++ implementations, though it should have gone the other way (in case the id_t typedef changes in the future to be a wider integer type)

ptheywood (changes requested):
`id_t` should be used for the results of `getID()`. 

Otherwise behaviour looks good from visualisations. 

I'm still keen for the dependency_graph example to be folded in, but that can wait.

examples/swig_boids_spatial3D/boids_spatial3D.py
MILeach: Accidental change when copy-pasting. I'll revert this and change the others to id_t as well.

Robadob:
This will need to be squash merged.

ptheywood (approved):

PR #740, opened 2021-11-30
WIP Alpha2 Release changes

Robadob:
- [x] Fix windows Warnings
- [x] Update Changelog
- [x] Update citation cff

Changelog and citation cff, currently assume 2021-12-02 release date.

https://github.com/FLAMEGPU/FLAMEGPU2/blob/master/CONTRIBUTING.md#creating-a-release

PR #741, opened 2021-11-30
Manual-only Windows tests CI 

ptheywood:
+ Adds a new GitHub Actions Workflow for the tests suite on Windows runners, which is only manually invoked.
+ Updates Windows CI to use CUDA 11.5.1, which includes compiler bugfixes. 
   + Due to the addition of Windows 11 support, the download URL pattern / filename appears to have changed with this release.

### DIFF
--- a/.github/scripts/install_cuda_windows.ps1
+++ b/.github/scripts/install_cuda_windows.ps1
@@ -30,6 +30,7 @@ $CUDA_KNOWN_URLS = @{
     "11.4.1" = "https://developer.download.nvidia.com/compute/cuda/11.4.1/network_installers/cuda_11.4.1_win10_network.exe"
     "11.4.2" = "https://developer.download.nvidia.com/compute/cuda/11.4.1/network_installers/cuda_11.4.1_win10_network.exe"
     "11.5.0" = "https://developer.download.nvidia.com/compute/cuda/11.5.0/network_installers/cuda_11.5.0_win10_network.exe"
+    "11.5.1" = "https://developer.download.nvidia.com/compute/cuda/11.5.1/network_installers/cuda_11.5.1_windows_network.exe"
 }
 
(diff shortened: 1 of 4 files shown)


ptheywood:
As the new workflow is only triggered by manual dispatch, it cannot be tested in this PR until it has been merged into master. 

Instead, an additional branch `windows-tests-ci-test-do-not-merge` has been pushed with a minor change to trigger on all push events

See https://github.com/FLAMEGPU/FLAMEGPU2/actions/runs/1521864992 for the output of that run (which will be a failure due to known warnings while building the test target). That is the expected output until #740 is merged.

ptheywood:
The test of this above ran as expected, so this is good to merge IMO.

.github/scripts/install_cuda_windows.ps1
```diff
@@ -108,15 +109,23 @@ echo "$($CUDA_PACKAGES)"
...
 # Select the download link if known, otherwise have a guess.
 $CUDA_REPO_PKG_REMOTE=""
+$CUDA_REPO_PKG_LOCAL=""
 if($CUDA_KNOWN_URLS.containsKey($CUDA_VERSION_FULL)){
     $CUDA_REPO_PKG_REMOTE=$CUDA_KNOWN_URLS[$CUDA_VERSION_FULL]
 } else{
     # Guess what the url is given the most recent pattern (at the time of writing, 10.1)
     Write-Output "note: URL for CUDA ${$CUDA_VERSION_FULL} not known, estimating."
-    $CUDA_REPO_PKG_REMOTE="http://developer.download.nvidia.com/compute/cuda/$($CUDA_MAJOR).$($CUDA_MINOR)/Prod/network_installers/cuda_$($CUDA_VERSION_FULL)_win10_network.exe"
+    if([version]$CUDA_VERSION_FULL -ge [version]"11.5.1"){
```
Robadob, line 118: Well this is an ugly edgecase.

Robadob (approved):

PR #744, opened 2021-12-06
BugFix: Release build DTHROW would handle %s incorrectly.

### LINKED ISSUE #743: Property variable name missing from device exception message under linux release with seatbelts
`DeviceExceptions` thrown for environemnt properties that do not exist do not always print the message variable in the exception string. 

This only appears to affect Release builds under linux with (`SEATBELTS=ON`).

E.g. the Test `DeviceExceptionTest.DeviceEnvironmentGetUnknownProperty` should throw an exception with the following error message, which it does under debug builds on linux:

```
First error:
/home/ptheywood/code/flamegpu/FLAMEGPU2/include/flamegpu/runtime/utility/DeviceEnvironment.cuh(143)[0,0,0][0,0,0]:
Environment property with name: nope was not found.
```

But for Release builds (with SEATBELTS) it throws:

```
/home/ptheywood/code/flamegpu/FLAMEGPU2/include/flamegpu/runtime/utility/DeviceEnvironment.cuh(143)[0,0,0][0,0,0]:
Environment property with name:  was not found.
```

To quickly enable the output in all build modes (but break the test suite) `run` in `test_device_exception.cu` can be modified to the following:

```
    void run(int steps = 2) {
        // CudaModel must be declared here
        // As the initial call to constructor fixes the agent population
        // This means if we haven't called model.newAgent(agent) first
        CUDASimulation cudaSimulation(model);
        cudaSimulation.SimulationConfig().steps = steps;
        // This fails as agentMap is empty
        AgentVector population(agent, 1);
        cudaSimulation.setPopulationData(population);
        try {
            cudaSimulation.simulate();
        } catch (const exception::DeviceError& e) {
            printf("%s\n", e.what());
        } 
    }
```

Then rebuild the test suite and run 

```
./bin/Release/tests --gtest_filter="DeviceExceptionTest.DeviceEnvironmentGetUnknownProperty"
```

Ideally we should add a test which checks the error message string is correct, otherwise this could be re-introduces in the future. 
It is likely this occurs for other types of device exception too.

This was highlighted by Kenneth.

Robadob:
This fixes the bug. Currently investigating whether tests can be tweaked to catch this bug.

Closes #743

### DIFF
--- a/include/flamegpu/exception/FLAMEGPUDeviceException_device.cuh
+++ b/include/flamegpu/exception/FLAMEGPUDeviceException_device.cuh
@@ -13,7 +13,12 @@ namespace exception {
 /**
  * This allows us to write DTHROW("My Error message: %d", 12); or similar to report an error in device code
  */
+#ifdef __CUDACC__
 #define DTHROW flamegpu::exception::DeviceException::create(__FILE__, __LINE__).setMessage
+#else
+// Just trying to make host compiler happy when it sees device code by mistake
+#define DTHROW(...)
+#endif
 
 /**
  * This struct should exist in device memory per stream
@@ -55,6 +60,7 @@ struct DeviceExceptionBuffer {
      */
     unsigned int arg_offset;
 };
+#ifdef __CUDACC__
 /**
  * This class is used on device for reporting errors
  * It should be used indirectly via the DTHROW macro
@@ -98,7 +104,7 @@ class DeviceException {
      * This is required to process var args, it performs the processing of a single arg
      */
     template<typename T>
-    __device__ void subformat(DeviceExceptionBuffer *buff, T t);
+    __device__ inline void subformat(DeviceExceptionBuffer *buff, T t);
     /**
      * Called by setMessage
      * This is required to process var args, it performs recursion to scrape of each arg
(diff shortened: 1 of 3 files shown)


ptheywood (approved):
all tests pass on linux

Robadob:
Windows Release SeatBelts=ON

```
[==========] 997 tests from 80 test suites ran. (1599334 ms total)
[  PASSED  ] 997 tests.

  YOU HAVE 5 DISABLED TESTS
 ```

**Comment by Robadob** [comment]
Windows, Release, seatbelts off
I think the fail is a temperenmetnal perf test

```
[==========] 877 tests from 78 test suites ran. (1487155 ms total)
[  PASSED  ] 876 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] TestCUDASimulationConcurrency.RTCLayerConcurrency

 1 FAILED TEST
  YOU HAVE 125 DISABLED TESTS
  ```

ptheywood:
Yep that's #745, not sure when it started being tempremental though (master before this PR triggered it).

Robadob:
Mine was cuda 11_4

PR #746, opened 2021-12-06
Perform more repetitions (of increment) for RTC concurrency testing

### LINKED ISSUE #745: Intermittant RTCLayerConcurrency Test Failure
The test `TestCUDASimulationConcurrency.RTCLayerConcurrency` is interrmittantly failing on my 3060ti machine, for `-DCMAKE_BUILD_TYPE=Release -DCUDA_ARCH=86`. This appears to be the case with `SEATBELTS` `ON` or `OFF`. 

This is a performance test, so it's not a strict failure and may have been introduced by a driver change or just be an unreliable test.

```
./bin/Release/tests --gtest_filter="TestCUDASimulationConcurrency.RTCLayerConcurrency"
```

This failed ~ 40% of the time after running a numeber of times in quick succession. 

Expected: (speedup) >= (SPEEDUP_THRESHOLD), actual: 0.718858 vs 1.5

This may be a side effect of the same GPU being used to drive my displays, however it is the only test in `TestCUDASimulationConcurrency` which failed. It does only take `25ms` in total, which may be a factor (a longer running kernel might be needed). 

I've not looked into this yet, just making a note for future investigation if needed.  

```
/home/ptheywood/code/flamegpu/FLAMEGPU2/tests/test_cases/gpu/test_cuda_simulation_concurrency.cu:1288: Failure
Expected: (speedup) >= (SPEEDUP_THRESHOLD), actual: 0.698954 vs 1.5
```

ptheywood:
RTC runtime was too fast for reliable timing, so a longer running kernel was required.

Closes #745

### DIFF
--- a/tests/test_cases/gpu/test_cuda_simulation_concurrency.cu
+++ b/tests/test_cases/gpu/test_cuda_simulation_concurrency.cu
@@ -1236,11 +1236,11 @@ RELEASE_ONLY_SEATBELTS_TEST(TestCUDASimulationConcurrency, LayerConcurrencyDevic
 const char* rtc_slowAgentFunction = R"###(
     FLAMEGPU_AGENT_FUNCTION(rtc_slowAgentFunction, flamegpu::MessageNone, flamegpu::MessageNone) {
         // Repeatedly do some pointless maths on the value in register
-        const int INTERNAL_REPETITIONS = 65536;
+        const int INTERNAL_REPETITIONS = 1048576;  // More repetations for RTC to get reliable timing, it's (much) faster than non rtc. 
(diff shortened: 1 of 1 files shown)


Robadob:
I'll try and remember to do it tomorrow,

Robadob:
Just ran the version that failed yesterday again, with less other stuff using GPU, and it passed.

Rebuilt it with this branch and it also passed.

Possible new warning:
`test_cuda_simulation_concurrency.cu(70): warning C4244: 'return': conversion from 'double' to 'float', possible loss of data`

Robadob (approved):

ptheywood:
I'll assume that it's good for now then, although we may need to adjust this further in the future

PR #753, opened 2021-12-14
Add CUDASimulation::simulate(RunPlan&)

### LINKED ISSUE #678: CUDASimulation::simulate(RunPlan plan)
It might be nice to provide a mechanism to run simulations using individual `RunPlan` objects?

There is overlap between a RunPlan and various parts of a simulations's Config object however, which would need to be clear. 

I.e. how many steps would be used below:

```c++
flamegpu::CUDASimulation simulation(model);
simualtion.Config().steps=10;
RunPlan plan1(model);
plan1.setSteps(20);
RunPlan plan2(model);
simulation.simulate(); // This would do 10 steps
simulation.simulate(plan1); // This would do 20 steps.
simulation.simulate(plan2); // this would do 1 step
```

Robadob:
CUDASimulation can be simulated with a `RunPlan`, includes a couple of new tests (which pass on Windows).

In writing this, I considered adding a call to `CUDASimulation::reset()` at the end, or should I just manually revert env properties?

Not sure how I feel about that, ~~as a minimum probably worth tweaking the doc comment to suggest it~~.
*I looked, and `reset()` isn't exposed to the user, so not sure how a user would be able to reset a `CUDASimulation`.*

Closes #678

### DIFF
--- a/include/flamegpu/gpu/CUDASimulation.h
+++ b/include/flamegpu/gpu/CUDASimulation.h
@@ -36,6 +36,7 @@ class CUDAAgent;
 class CUDAMessage;
 class LoggingConfig;
 class StepLoggingConfig;
+class RunPlan;
 
 struct RunLog;
 
@@ -131,6 +132,14 @@ class CUDASimulation : public Simulation {
      * Includes init and exit functions calls.
      */
     void simulate() override;
+    /**
+     * Execute the simulation using the configuration and environment properties from the provided RunPlan
+     * @param plan The RunPlan to use to execute the simulation
+     *
+     * @throw exception::InvalidArgument If the provided RunPlan does not match the CUDASimulation's ModelDescription
+     * @note The config.steps and config.random_seed values will be ignored, in favour of those provided by the RunPlan
+     */
+    void simulate(const RunPlan &plan);
     /**
      * Replaces internal population data for the specified agent
      * @param population The agent type and data to replace agents with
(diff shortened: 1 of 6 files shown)


ptheywood:
I don't think resetting by default is a good idea, it would prevent the use of environment properties as a mechanism for gettiing outputs from a simulation without needing to add globally scoped vars into exit functions or similar. 

Reset also sets things to default values, not the value from before the last call to simulate() which might also not be that useful. 

In general, I think its fairly safe to expect a single cudaSimualtion to only be simulated once, with users having to manually re-set it to the desired state before re-starting it (resetting initial populations + env vars) althought it could be useful to make it easier. 

The `TestCUDASimulation` suite passes under linux.

ptheywood (approved):

Robadob:
I was just thinking about users trying to manually implement a `CUDAEnsemble`. Where they might repeatedly call `CUDASimulation::simulate(RunPlan&)`. Feels like an easy oversight, to not reset env properties etc.

ptheywood:
Providing a mechanism to do that might be nice, but do you reset to the value to default, or to before `simulate()` was called (which would require maintaining a copy of the env at that point). 

Stopping usesrs from being able to access the state of the env at the end of a simulation outside of an exit function seems like a poor choice though (without offering a nicer way to get values in/out of exit functions). 

This addition of `simulate(RunPlan)`'s main improvement is for modles that can either run a simulation or an ensemble though, as you can then use `RunPlan` for both rather than having different logic for initialisation.

PR #757, opened 2022-01-07
Suppress note emitted by GCC >= 10 on ppc64le about changes since GCC 5

ptheywood:
On RHEL 8 on a PPC64LE system, GCC 10 emits notes which were not emitted by GCC 8.4.0 (the previously tested version of GCC on PPC64LE). 

This PR suppresses these additional notes. The full test suite passed with this compiler configuration on PPC64LE with CUDA 11.4, executed on an T4 GPU.

### DIFF
--- a/cmake/warnings.cmake
+++ b/cmake/warnings.cmake
@@ -118,6 +118,11 @@ function(SuppressSomeCompilerWarnings)
     endif()
     # Suppress nodiscard warnings from the cuda frontend
     target_compile_options(${SSCW_TARGET} PRIVATE "$<$<COMPILE_LANGUAGE:CUDA>:SHELL:-Xcudafe --diag_suppress=2809>")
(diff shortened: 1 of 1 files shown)


Robadob (approved):
Change looks fine, though I'm not sure of the significance of the actual note.

ptheywood:
g++ can gnerate code which is not compatible with the vendor neutral C++ ABI, and depending on the specific abi version requested it will warn about different things. 
psABI is the processor specific ABI, so this jsut turns off warnings notes about ABI differences that are POWER9 specific only.

For PPC64LE versions of gcc there are some specific flags which may also be relevant, with the default value leading to the not being emitted. 

As we build everything that we can ourselves on linux, we shouldn't really be crossing any ABI boundaries, so it should be fairly safe to suppress these warnings I believe.

PR #758, opened 2022-01-13
Add CUDA 11.6.0 to CI builds

### LINKED ISSUE #761: MSVC + CUB 1.15.0 + Windows.h compiler issues
Under MSVC, including `jitify.hpp` prior to `cub.cuh` results in compiler errors as of Thrust/CUB 1.15.0. 

This is not an issue with CUB <= 1.14, nor with GCC, nor if `cub.cuh` is included first.

With the complexity of our include hierarchies, makeing sure cub is only ever included before jitify is not going to be overly easy. 

For now, I'll be blocking CUB 1.15 on MSVC via CMake (as part of #758), but longer term this needs to be promoted to a CUB and / or Jitify issue (Ideally when we know which project is the root cause of the problem).

I've created a MWE of this which uses CMake to fetch the two dependencies and demonstrate via CI the issue, which can be used as reference in the promoted bug reports

https://github.com/ptheywood/jitify-cub-1.15-MSVC-demo

My next planned step (for when in windows and I have time) is to try and narrow down which part of the jitify headers is leading to the compiler error in CUB. Otherwise we could just open issues in both Nvidia Repos.

ptheywood:
CUDA 11.6 support

+ Suppress `__device__ cudaDeviceSychrozize()` for CUDA >= 11.6 with MSVC (--isystem suppresses this already on linux/gcc).
+ Updates the FLAMEGPU/FLAMEGPU2-visualiser repository pinned version to include the above change
+ Changes CMake functions related to compiler warnings, so they are not overwritten by the copy in the visualisation repository (in case there are differences)
+ Enabled CUDA 11.6 on CI where relevant.

This has been tested under linux, including an RTC build which may have been impacted by a compiler default change which may have potentially had a negative impact.

Closes #761

### DIFF
--- a/cmake/dependencies/flamegpu2-visualiser.cmake
+++ b/cmake/dependencies/flamegpu2-visualiser.cmake
@@ -8,7 +8,7 @@ cmake_policy(SET CMP0079 NEW)
 
 # Set the visualiser repo and tag to use unless overridden by the user.
 # @todo - If the git version has changed in this file, fetch again? 
-set(DEFAULT_VISUALISATION_GIT_VERSION "flamegpu-2.0.0-alpha.1")
+set(DEFAULT_VISUALISATION_GIT_VERSION "d35ad8f265364ee05bca6c4bec62374d1dbbea94")
 set(DEFAULT_VISUALISATION_REPOSITORY "https://github.com/FLAMEGPU/FLAMEGPU2-visualiser.git")
 
 # If overridden by the user, attempt to use that
--- a/.github/scripts/install_cuda_ubuntu.sh
+++ b/.github/scripts/install_cuda_ubuntu.sh
@@ -14,6 +14,7 @@ CUDA_PACKAGES_IN=(
     "cuda-nvtx"
     "cuda-nvrtc-dev"
     "libcurand-dev" # 11-0+
+    "cuda-cccl" # 11.4+, provides cub and thrust. On 11.3 knwon as cuda-thrust-11-3
 )
 
 ## -------------------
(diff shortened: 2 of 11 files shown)


ptheywood:
Due to use of `cudaDeviceSynchronize` in device code in cub, which is marked as deprecated in CUDA 11.6.0 but no alternative provided yet, Windows CI needs additional tweaks for this warning (promoted to an error) to not be emitted.
Presumably this is not an issue for linux CI builds due to the use of `--isystem`, so no warnings from thrust/cub are emitted.

Otherwise, under linux RTC models run successfully (boids_rtc_spatial3D) as does the test suite. This suggests  that my concern about unused kernel optimistaion was not warranted.

ptheywood:
I gave up on the graceful handling of Thrust/CUB 1.15, instead just blocking CUDA 11.6 on MSVC from finding it's own Thrust/CUB. 

CUB's patch has been merged in upstream, so in the future we can just switch to requiring CUB >= 1.16 and no longer worry about it.

cmake/dependencies/flamegpu2-visualiser.cmake
```diff
@@ -8,7 +8,7 @@ cmake_policy(SET CMP0079 NEW)
 
 # Set the visualiser repo and tag to use unless overridden by the user.
 # @todo - If the git version has changed in this file, fetch again? 
-set(DEFAULT_VISUALISATION_GIT_VERSION "flamegpu-2.0.0-alpha.1")
+set(DEFAULT_VISUALISATION_GIT_VERSION "d35ad8f265364ee05bca6c4bec62374d1dbbea94")
```
Robadob, line 11: Worth updating release guide to include visualisation release tag/cmake bind, if required?

Robadob (approved):
Looks good, added a note re: whether we want to do vis releases in sync with main repo releases.

cmake/dependencies/flamegpu2-visualiser.cmake
ptheywood: > The CMake version pinning of the visualisation repo is correct. Ideally to a tag specific to the main repo release (i.e. flamegpu-2.0.0-alpha)

It's mentioned but I'll clarify that a new tag should be made if required.

Robadob: Fair, I guess I don't recall (and maybe even missed it) when I did last release.

PR #760, opened 2022-01-14
CUDASimulation::setEnvironmentProperty()

Robadob:
Adds `CUDASimulation::setEnvironmentProperty()`, `CUDASimulation::getEnvironmentProperty()`, similar to the existing `setPopulationData()`, `getPopulationData()`. Added one thorough test to C and SWIG suites respectively. 

Other Changes:
*  `EnvironmentManager.setProperty()`/`getProperty()` did not check that length=1
* The test HostEnvironment::ExceptionPropertyReadOnly was calling non-array method with array types.

Todo:
- [x] SWIG template expansion
- [x] Tests

### DIFF
--- a/include/flamegpu/gpu/CUDASimulation.h
+++ b/include/flamegpu/gpu/CUDASimulation.h
@@ -155,6 +155,65 @@ class CUDASimulation : public Simulation {
      */
     void getPopulationData(AgentVector& population, const std::string& state_name = ModelData::DEFAULT_STATE) override;
     /**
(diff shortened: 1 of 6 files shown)


ptheywood (approved):
Diff looks good to me, test coverage looks reasonable too. 

I've built the test suite, but only ran the CUDASimulation sub-suite, which all passes under linux.

Robadob:
I ran the full C suite, should be fine.

Will merge it now, as I want/need it for the primage thing i'm working on.

PR #763, opened 2022-01-18
BugFix: Logging would divide by 0 if calculating standard deviation…

Robadob:
…of variable for empty agent population.

Noticed it whilst investigating adding standard deviation to HostAPI.


ptheywood (approved):
Does this want to be `static_cast<T>(0)`? though it probably doesn't matter because casting 0 between any numeric type should be fine.

Robadob:
Function returns type `util::Any`, I just copied the normal return statement rather than looking to deeply into it.

Logging has to store values of a bunch of types inside a map, hence the use of `Any`. The only possible problem i forsee here would be the Any storing 0 as type int, but it being read back as type double.

Robadob:
Small change, tweaked it to `util::Any(0.0)`, so its a double inside an `Any`.

PR #768, opened 2022-01-21
Update file IO to support all config struct members.

### LINKED ISSUE #767: XML & JSON IO needs updating to support new config options
Various newer config options are missing from both the input/output file formats. e.g. the 3 logging ones.

Robadob:
Writers done, readers todo.

Closes #767


ptheywood (approved):
Looks fine, though would be nice if tests were provided to verify (or promote to an issue for future)

Robadob:
> Looks fine, though would be nice if tests were provided to verify (or promote to an issue for future)

I had done tests, i just forgot to push.

Robadob:
Also noticed that the logging writers only output "steps" and "random_seed" within the config struct. This seems like an intentional decision, do you think it's worth expanding?

ptheywood:
> Also noticed that the logging writers only output "steps" and "random_seed" within the config struct. This seems like an intentional decision, do you think it's worth expanding?

I don't see any harm in expanding it, just a shame that c++ doesn't have reflection so can't do it automatically.

Issue #780, opened 2022-01-31
Diffusion Example


mondus:
Would be very useful for multiscale modelling and woudl be trivial to implement with Array2D messaging. Some good reference to analytic approaches at the following link: https://enccs.github.io/OpenACC-CUDA-beginners/2.02_cuda-heat-equation/

Robadob:
Started working on that example, then noticed it does von-neumann neighbourhood, rather than Moore. We don't have that.

PR #781, opened 2022-01-31
Add getOffsetX/Y/Z() to array messages during iteration.

Robadob:
This will benefit users implementing diffusion.

Relates #780


mondus (approved):
Notify on the discussion board once merged.

PR #789, opened 2022-02-07
Minor fix

Robadob:
This has no programatic impact, however it's inconsistent with implementation so was causing some confusion in api docs generation, making it harder to cross ref from user guide,


Robadob:
Will merge after CI passes.

ptheywood (approved):

PR #794, opened 2022-02-09
MessageSpatial2D::In::wrap()

### LINKED ISSUE #185: Wrapped continuous spatial messaging
Following #30, spatial messaging was added, however unlike FGPU1 new spatial doesn't currently support wrapping over environment bounds.

Wasn't added at the time as it doesn't fit well alongside strips optimisation, and best implementation would require a whole new version, rather than a device side flag/branch. This could be written within an hour or so, but seems overkill unless required.

Robadob:
Python Circles 2D wrapped example [here](https://gist.github.com/Robadob/afabd42efc736366b390e74d9b80d6d1).

Currently `operator++()` only returns messages within the search radius, **this is different to regular spatial messages**.
I'm kind of torn on whether to keep this behaviour which adds a redundant extra load of each location variable and caches the distance, or to revert back to non-wrapped behaviour and provide a distance function which users must pass 2-4 floats to for the result.

Not to sure what else can be tested, just feels like it probably needs more.

Docs might need some minor redrafting, didn't feel too clean.

Haven't added an example to the main repo (there's already too much bloat with multiple versions of circles).

Docs PR: https://github.com/FLAMEGPU/FLAMEGPU2-docs/pull/90

Closes #185

### DIFF
--- a/include/flamegpu/runtime/messaging/MessageSpatial2D.h
+++ b/include/flamegpu/runtime/messaging/MessageSpatial2D.h
@@ -65,7 +65,7 @@ class MessageSpatial2D {
         /**
          * max-lowerBound
          */
-        float environmentWidth[3];
+        float environmentWidth[2];
     };
 };
 
--- a/include/flamegpu/runtime/messaging/MessageSpatial2D/MessageSpatial2DDevice.cuh
+++ b/include/flamegpu/runtime/messaging/MessageSpatial2D/MessageSpatial2DDevice.cuh
@@ -216,7 +216,243 @@ class MessageSpatial2D::In {
         const MetaData *metadata;
     };
     /**
-     * Constructer
+     * This class is created when a search origin is provided to MessageSpatial2D::In::wrap()(float, float)
+     * It provides iterator access to a subset of the full message list, according to the provided search origin
+     *
+     * @see MessageSpatial2D::In::wrap()(float, float)
+     */
+    class WrapFilter {
+     public:
+        /**
+         * Provides access to a specific message
+         * Returned by the iterator
+         * @see In::Filter::iterator
+         */
+        class Message {
+            /**
+             * Paired Filter class which created the iterator
+             */
+            const WrapFilter& _parent;
+            /**
+             * Relative cell within the Moore neighbourhood
+             * relative_cell[0] corresponds to x offset
+             * relative_cell[1] corresponds to y offset
+             */
+            int relative_cell[2] = { -2, 1 };
+            /**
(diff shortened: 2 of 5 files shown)


ptheywood:
Initialising with a (relatively) large environment compared to the comms/interaction radius, with agent clustered at opposite edges might increase the liklihood of getting a circle across the boundary?

Explicit test cases are the way to go anyway.

Robadob:
> Initialising with a (relatively) large environment compared to the comms/interaction radius, with agent clustered at opposite edges might increase the liklihood of getting a circle across the boundary?

See slack, I tried that, and it seemed positive biased, and the 1 circle that did get split would never stabilise.

Robadob:
I did try locking an agent to (0,0) and logging messages it read and those seemed right.
I think my issue is the wrapped vectors aren't right.

Robadob:
Have just updated this to work with current master (post shared mem curve), also updated the gist example as it was using old dependency graph API.

Robadob:
Now working on this again, to summarise current status.

The wrapped iterator's `Message` object has these additional methods distinct from the normal iterator.
* `getDistance()`: Returns the shortest distance to message in wrapped environment
* `getVirtualX(float x1)`: Returns the closest x coordinate in a wrapped environment (so if it wraps over the left edge, this will return a coordinate beyond the left edge)
* `getVirtualY(float y1)`

This is different from the regular iterator, that the wrapped iterator **only** returns messages which are within the radius. This was setup that way when I initially developed the example (presumably in an attempt to minimise users doing stuff wrong), and could easily be tweaked to match the regular iterator's behaviour.

ptheywood:
We should think about how we can control wrapping in some but not all dimensions if required (i.e. wrapping in 3 dimensions creates a shape which can't be represented in 3D space, with FLAME GPU 1 only warpping in X and Y  but never Z to form a torus)

Will try and get this reviewed in the net few days otherwise

Robadob:
As discussed in meeting, I've now removed the wrapped distance function entirely. It works more like regular spatial. With additional `getVirtualX()` and `getVirtualX(float)` methods for each axis, these provide the 'wrapped' coordinate relative to the search origin or a user specified coordinate respectively (the latter is probably not required, but probably harmless.

These methods internally call `getVariable<float>("x")`, it's intended that users call them, rather than loading the variables themselves.

mondus (approved):
I am happy with the changes. Think the interface for this is really nice.

mondus:
Delay merging to have a discussion about iterator for spatial messaging. There is a fairly good argument that this should do distance checking.

PR #804, opened 2022-02-22
Cmake 3.18 fix + CI

ptheywood:
+ Adds a CMake Configuration workflow which tests that the oldest supported and current github action default CMake versions configure correctly
+ Fixes an issue with `USE_NVTX=ON` and CMake `3.18.x` due to the use of the `version` imported target property, which was not supported until `3.19.0`

### DIFF
--- a/.github/workflows/CMake.yml
+++ b/.github/workflows/CMake.yml
@@ -1,0 +1,169 @@
+# Configure CMake for a number of CMake versions, but do not compile.
+name: CMake
+
+on:
+  # Branch pushes that do not only modify other workflow files
+  push:
+    branches:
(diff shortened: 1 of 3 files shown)


ptheywood:
First commit shows that this was an issue, via the new CI. 

Second commit fixes it. 

Probably want to squash these for merging though.

Robadob (approved):
CMake changes seem to make sense, and CI presumably proves the new CI run works.

Yeah squash it first if you want.

Robadob:
Build with CMake 3.18 or 3.19+?

Mav default is 3.18, so testing there would be fast?

PR #805, opened 2022-02-22
Add support for automatically sorting agents with xy and xyz location vars

Robadob:
Also added a bit of extra type checking and cuda error checking, this caught a bug when trying to sort an empty agent pop.

I've additionally extended agent sort detection to submodels. For the time being, this should suffice for Primage, where I'm hoping to sort once per submodel (I just set the sort period unreasonably high). But we may want to consider if a model like primage only wanted to sort every N parent model steps.

*Note, this merges into `logging_timings` branch, because I'm going to test it with Primage, which currently requires that for the timing stuff. Although after all this faff, I've manually applied the subsequent changes in both places.*


mondus (changes requested):
Add a test to expose the previous missing feature

PR #812, opened 2022-02-24
type_decode: GLM types can now mostly be treated like native types

### LINKED ISSUE #809: USE_GLM Test failures
A number of tests fail due to the `USE_GLM` relaxed type checking.

```
[  FAILED  ] 6 tests, listed below:
[  FAILED  ] TestCUDASimulation.setEnvironmentProperty
[  FAILED  ] HostAgentCreationTest.BadVarType
[  FAILED  ] HostAgentCreationTest.GetterBadVarType
[  FAILED  ] HostEnvironmentTest.ExceptionPropertyType_double
[  FAILED  ] HostMiscTest.ArrayVarNotSupported_reduce
[  FAILED  ] HostMiscTest.ArrayVarNotSupported_transformReduce
```

Rather than disable these tests for `USE_GLM`, it would be better for me to finally get around to allowing GLM to do proper type checking by splitting GLM types into a type and length.

Robadob:
By using `type_decode<glm::vec3>` the base type and length can be extracted. The same `type_decode` works on the native types, e.g. `float`, reporting them with a length of 1.

This requires some thorough testing, as it's possible I've missed stuff. I will probably need to add some extra tests.

## Testing: 

- [x] `AgentDescription`
- [x] `MessageDescription`
- [x] `EnvironmentDescription`
    - [ ]  macro property?
- [x] Device Agent *(DeviceAPI can't type check, so has much weaker checks, no code changes required)*
- [x] Device New Agent *(DeviceAPI can't type check, so has much weaker checks, no code changes required)*
- [x] Device Environment *(DeviceAPI can't type check, so has much weaker checks, no code changes required)*
    - [ ]  macro property?
- [x] Host New Agent
- [x] Host Agent reduce *(Only custom transform and reduce are suitable for vec types)*
- [x] `HostEnvironment`
    - [ ]  macro property?
- [x] `AgentVector::Agent`
- [x] `AgentInstance`
- [x] `CUDASimulation` (set/get environment)
- [x] `RunPlan` ?
- [ ] Python stuff?
   
Closes #809

### DIFF
--- a/include/flamegpu/gpu/CUDASimulation.h
+++ b/include/flamegpu/gpu/CUDASimulation.h
@@ -173,6 +173,17 @@ class CUDASimulation : public Simulation {
      */
     template<typename T, unsigned int N>
     void setEnvironmentProperty(const std::string &property_name, const std::array<T, N> &value);
+    /**
+     * Update the value of the specified element of the named environment property array
+     * @param property_name Name of the environment property to be updated
+     * @param index Index of the element within the property array to be updated
+     * @param value New value for the named environment property
+     * @tparam T Type of the elements of the environment property array
+     * @throws exception::InvalidEnvPropertyType If the named environment property does not exist with the specified type
+     * @throws exception::ReadOnlyEnvProperty If the named environment property is marked as read-only
+     */
+    template<typename T>
+    void setEnvironmentProperty(const std::string& property_name, const EnvironmentManager::size_type &index, const T& value);
 #ifdef SWIG
     /**
      * Update the current value of the named environment property array
@@ -188,7 +199,7 @@ class CUDASimulation : public Simulation {
 #endif
     /**
      * Return the current value of the named environment property
-     * @param property_name Name of the environment property to be updated
+     * @param property_name Name of the environment property to be returned
      * @tparam T Type of the environment property
      * @throws exception::InvalidEnvPropertyType If the named environment property does not exist with the specified type
      */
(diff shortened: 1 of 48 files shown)


Robadob:
At current, I appear to have broken RTC dynamic header

Getting this error repeated quite a few times

```
rtc_test_func_impl_curve_rtc_dynamic.h(396): error: list-initialization of an object type "void" is not allowed because the type is incomplete
```

Robadob:
so it appears gcc and windows disagree
gcc wants `typeid(typename type_decode<T>::type_t)`
visual studio wants `typeid(type_decode<T>::type_t)`
Neither seems to accept the other
rtc seems to agree with visual studio

Robadob:
Remaining tasks include `RunPlan`, `RunPlanVec`, testing the python API and documentation.

I implemented, the basic set/get methods of `RunPlan`, `RunPlanVec`, and discovered they have an unusual edge case test which breaks most of my existing changes. `-1` is passed to the `unsigned int` index, this then overflows to a low value when doing the range check. But becomes the large value again when doing the memory access. So I need to go back through all the array element operators i've updated and improve that check.

Similarly, `RunPlanVector` has some fancy methods for random/lerp initialisation. These will require some special handling for glm, need to write that.

Robadob:
SWIG doesn't have GLM support yet (although it could be used inside agent fns). So need to decide whether will do that.

Also need to come up with a clear list of places GLM is not supported, so it can be detailed in the docs.

Robadob:
ToDo:
* I think we need more RTC tests of glm support in device code. Python doesn't have `USE_GLM` switch, so making optional tests in that is harder.
* Need to solve whatever is up with Ubuntu/CUDA 10.0 CI

Things not supported:
* Python Model Desc/Execution/Host Runtime APIs
* Macro properties
* Predefined reduction operators (custom can work)

mondus:
Merge soon and delay swig wrapping to another issue

Robadob:
So back at looking why CUDA 10 is failing to build.

Compilation error is that the linker can't find an implementation of `static constexpr int len_t = 1;` from inside the new `type_decode` template class.

It's `constexpr`, so I don't really think it should need to be linked, it should be treated as a literal.

CUDA did have issues with constexpr, so latest commit is intended to pass `--expt-relaxed-constexpr`, which is intended to allow device scope to access host scope constexpr and vice versa. Not clear whether this needs to be passed to compiler or linker, so might take some fiddling to actually try it properly on CI.

I previously did try changing it to `static const in len_t = 1;`, but this changed the CI failure, didn't fix it. Since rebase that's been lost so I don't know the exact error.

ptheywood:
> Have fixed the test, was due to integer underflow py vs c++ difference (because py int is probably a long or something)

python 3's generic `int` does not have a fixed upper limit in the language, it'll just keep growing until there isn't any host memory left to allocate to represent it with. (see `sys.maxsize` for the maximum number of bytes that can be used on a given system, i.e. the value of `std::numeric_limits<int>::max()`). 

I think numpy is the most commonly used provider of fixed-size integer types, and would assume that casting a larger value into a shorter numpy type would throw, but I've not checked this.

Robadob:
The test was already setup to wrap positive values before overflow, I just updated them all to wrap negative values before underflow too. New tests appear to work.

mondus (approved):

PR #814, opened 2022-03-02
BugFix: Calling RunPlan::setProperty() would fail silently

Robadob:
when called on a property which already had an override stored.

This was caused by `util::Any` being const, so limited `map` operations can be used. As such I probably got stuck with using `emplace` and didn't realise it won't replace existing values.

Improved the related test case to catch it.

Had a quick skim through other uses of emplace via gh search and couldn't spot any similar offenders.


ptheywood (approved):
Looks good w/ thorough test additions.

PR #817, opened 2022-03-03
BugFix: step counter internal property was being mapped to submodels.

### LINKED ISSUE #815: _stepCount internal env prop should not be mapped
It appears that the step counter's internal environment property is being mapped to submodels.

This causes any agent functions (and host functions) after a submodel, before the end of the step to report the incorrect step value when `FLAMEGPU->getStepCounter()`.

Confirmed the problem, by updating the step counter env property after the submodel returns. However, the proper fix would be to block it from being mapped in the first place.

I would presume agent `_id` var wants to always be mapped. But there doesn't appear to be anything explicit in the code to handle this so far, just all existing usages use automap which maps all matching.

Robadob:
Now blocking all internal environment properties/macro properties from being mapped (by users).

Whether the same should be applied to agent variables (e.g. `_id`) remains an open question. (I had a look around but couldn't find any explicit code mapping it)

Also extended tests.

Closes #815


ptheywood (approved):
looks good and new tests would catch the bug. I've not built/ran it myself.

PR #823, opened 2022-03-25
BugFix: RunPlanVector::setPropertyUniformDistribution() 

Robadob:
would round non-integer values.

Have improved the existing test to now check exact values (previously it just checked ordering). Similarly added a floating point cases in a range whereby lerp will produce non integer values.


ptheywood (approved):
:+1: 

Tests pass under linux
```
./bin/Release/tests --gtest_filter="TestRunPlanVector.*"
... 
[  PASSED  ] 15 tests.
```

PR #827, opened 2022-03-31
Add custom GitHub Action Annotations via problem-matchers

### LINKED ISSUE #480: Github Actions Annotations
Annotations are a newer github actions feature that clearly show warnings/errors without having to manually inspect the log. 

Some of the actiosn from the marketplace for running linters or compilers may emit these, but they might also not do things we want them to. 

An alterantive it is to post-process the output using sed/awk/grep and emit the relevant statements which the workflow runner will pick up on. 
Although this will complicate our action a little,  it might be better than executing random third party actions.

https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-a-warning-message

ptheywood:
Closes #480

### DIFF
--- a/.github/problem-matchers.json
+++ b/.github/problem-matchers.json
@@ -1,0 +1,56 @@
+{
+    "problemMatcher": [
+        {
+            "owner": "cpplint",
+            "severity": "error",
+            "pattern": [
+                {
+                    "regexp": "^(.+):(\\d+):\\s+(.+)\\[(.+)\\]\\s+\\[(.+)\\]$",
+                    "file": 1,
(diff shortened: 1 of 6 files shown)


ptheywood:
I've tested this with another branch which introduces a couple of linter / compiler warnings / errors. 

I've not enabled this on the release CI, as with such a large build matrix the number of annotations would be huge (GHA might be working on de-duplication/grouping, there's a long standing issue about it)

### Windows

image

### Manylinux

image

### Linux

image


### Lint

image

Robadob (approved):
Really nice.

PR #829, opened 2022-04-04
Replace cbrt() with sqrtf() in Circles examples.

Robadob:
Use of double function, incorrectly leads to large compute usage.

Also noticed `cbrt()` being used rather than `sqrt()`, a mistake when mapping from glm to raw math, (Presumably it doesn't significantly affect model behaviour as it was using `glm::length()` in fgpu1 models, just awkwardly reduce interaction radius, but haven't tested it).


ptheywood (approved):
I've confirmed via profiling that this removes FP64 pipe utilisation: 

image
(The blue bar is missing from FP64)

This change has improved performance:

+ Partially due to a change in the number of messages processed (fewer pesseages will pass the separation check now on average)
+ Partially due to the removal of fp32/fp64 conversions and the more expensive fp64 ops, especially on consumer hardware

I.e. for 80k agents on a 3060ti using CUDA 11.2:

image

PR #833, opened 2022-04-12
Remove certain mutexes around RTC kernel launches.

### LINKED ISSUE #469: Unnecessary RTC Mutexes
I think holding these two shared locks is redundant each RTC kernel launches. Potentially other mutex's related to curve/environment that can be avoided too. Might explain why we're getting less concurrency than expected for RTC when using ensembles.

https://github.com/FLAMEGPU/FLAMEGPU2_dev/blob/d23833392c0262e11253cf48f59aec1df2145276/src/flamegpu/gpu/CUDASimulation.cu#L361
https://github.com/FLAMEGPU/FLAMEGPU2_dev/blob/d23833392c0262e11253cf48f59aec1df2145276/src/flamegpu/gpu/CUDASimulation.cu#L578

Robadob:
Relates to the ongoing ensemble benchmarks, a cleaner version of what's implemented in branch [`DO-NOT-USE-no-mutex`](https://github.com/FLAMEGPU/FLAMEGPU2/tree/DO-NOT-USE-no-mutex).

I trust this is safe, as it only wraps the kernel launch and we know for certain that RTC kernel launches don't touch the device symbols it's protecting. There are smaller locks around accesses to curve/env for setting up RTC memory.

Ran Windows/Release/SEATBELTS=ON tests:

```
[==========] 1038 tests from 81 test suites ran. (43318 ms total)
[  PASSED  ] 1038 tests.

  YOU HAVE 5 DISABLED TESTS
```

Closes #469

### DIFF
--- a/src/flamegpu/gpu/CUDASimulation.cu
+++ b/src/flamegpu/gpu/CUDASimulation.cu
@@ -675,9 +675,11 @@ void CUDASimulation::stepLayer(const std::shared_ptr<LayerData>& layer, const un
 
     // If any condition kernel needs to be executed, do so, by checking the number of threads from before.
     if (totalThreads > 0) {
-        auto env_shared_lock = this->singletons->environment.getSharedLock();
-        auto env_device_lock = this->singletons->environment.getDeviceSharedLock();
+        std::shared_lock<std::shared_timed_mutex> env_shared_lock;
+        std::shared_lock<std::shared_timed_mutex> env_device_lock;
         if (!has_rtc_func_cond) {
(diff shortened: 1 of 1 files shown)


ptheywood:
Are there any RTC ensmeble tests that would catch any errors here if there were any?

[test_cuda_ensemble.cu](
https://github.com/FLAMEGPU/FLAMEGPU2/blob/master/tests/test_cases/gpu/test_cuda_ensemble.cu) doesn't include any by the looks of it.

Robadob:
You answered your own question, can't reliably test a race condition.

mondus (approved):

PR #836, opened 2022-04-19
Minor CMake Fix

ptheywood:
In practice the old state will have been working, but this just improves consistency within the function and will ensure it works when the project name and binary name are different.


cmake/common.cmake
```diff
@@ -328,7 +328,7 @@ function(add_flamegpu_executable NAME SRC FLAMEGPU_ROOT PROJECT_ROOT IS_EXAMPLE)
                 # mark_as_advanced(FORCE SDL2_DIR)
                 # find_package(SDL2 REQUIRED)
             # endif()
-            add_custom_command(TARGET "${PROJECT_NAME}" POST_BUILD        # Adds a post-build event to MyTest
+            add_custom_command(TARGET "${NAME}" POST_BUILD        # Adds a post-build event to MyTest
```
Robadob, line 331: The comments are nolonger inline

Robadob (approved):
Did fresh configure, build and run of circles brute force with vis.

Works.

PR #838, opened 2022-04-27
WIP Ensemble Benchmark optimisations

Robadob:
Most of this branch is removal of default stream usage.
In a few places, have also started making async versions of methods to later support in-layer concurrency.

Included in the second commit, required for peak ensemble performance:
* Removal of CUDA event timing during ensembles (this implicitly uses default stream blocking stuff to ensure greatest precision).
* Removal of CUDA device reset from CUDASimulation destructor (this exists for cleanliness, however adds a 250ms cost per run in 1x concurrent ensembles, making speed ups 'false')

This is the latest ensemble bench graph, collected using 3090 RTX on waimeu (and the 2 above noted changes not in this branch). Note the axis `E` now refers to the number of concurrent runs configured within the `CUDAEnsemble`. Now we are getting results we can understand, with high concurrent stream utilisation, it would appear we need to increase the population sizes. My hypothesis is that the speedup is more tightly linked to input kernel duration than population size.

As noted in the Slack thread, using `taskset 1,3,5 ./ensemble_benchmark` or similar, we can test how the benchmark performs with less CPU cores (than threads).

image

### DIFF
--- a/include/flamegpu/gpu/CUDAAgent.h
+++ b/include/flamegpu/gpu/CUDAAgent.h
@@ -169,7 +169,7 @@ class CUDAAgent : public AgentInterface {
      * @param stream CUDA stream to be used for async CUDA operations
      * @see HostAgentAPI::sort(const std::string &, HostAgentAPI::Order, int, int)
      */
-    void scatterSort(const std::string &state_name, CUDAScatter &scatter, const unsigned int &streamId, const cudaStream_t &stream);
+    void scatterSort_async(const std::string &state_name, CUDAScatter &scatter, unsigned int streamId, cudaStream_t stream);
     /**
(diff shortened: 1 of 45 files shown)


Robadob:
This should be all the desired changes, with the caveat that the changes in the second commit have not yet been tested.

Robadob:
Mav Release build, SEATBELTS probably off

```
[==========] 995 tests from 79 test suites ran. (83475 ms total)
[  PASSED  ] 995 tests.

  YOU HAVE 40 DISABLED TESTS
```

ptheywood (approved):
Diff looks good to me. I've not compiled / ran anything though.

Robadob:
image
Half the benchmark has executed so far, and it looks much the same.

PR #839, opened 2022-04-27
Improve CUDAEnsemble's error reporting.

Robadob:
Have added 3 tests to `TestCUDAEnsemble` suite, one for each error level, to confirm it works.

Docs repo PR needs merging at same time: https://github.com/FLAMEGPU/FLAMEGPU2-docs/pull/85

### DIFF
--- a/include/flamegpu/gpu/CUDAEnsemble.h
+++ b/include/flamegpu/gpu/CUDAEnsemble.h
@@ -25,34 +25,47 @@ class CUDAEnsemble {
      * Execution config for running a CUDAEnsemble
      */
     struct EnsembleConfig {
-        // std::string in = "";
         /**
          * Directory to store output data (primarily logs)
+         * Defaults to "" (the working directory, no subdirectory)
          */
         std::string out_directory = "";
         /**
          * Output format
          * This must be a supported format e.g.: "json" or "xml"
+         * Defaults to "json"
          */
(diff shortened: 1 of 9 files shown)


include/flamegpu/gpu/CUDAEnsemble.h
```diff
@@ -72,8 +79,9 @@ class CUDAEnsemble {
      * Execute the ensemble of simulations.
      * This call will block until all simulations have completed or MAX_ERRORS simulations exit with an error
```
ptheywood, line 80: This docstring could probably also do with adjusting, as the statement about when it exits is now also subject to the error level.

Robadob:
> Could do with test(s) for the changes to `CUDAEnsemble::checkArgs`, verifying all documented accepted values work (and set the correct error level) plus atleast one test of bad input (bad integer, bad string, missing value). Short and long forms would also be good.

That's reasonable, didn't think of that. Will do.

include/flamegpu/gpu/CUDAEnsemble.h
Robadob: Not even sure what `MAX_ERRORS` is, don't recall that (there was some prev error tracking stuff which was incomplete (the worker threads never updated the error counter).

Will address.

ptheywood (approved):

PR #841, opened 2022-04-29
Update CUDA CI installation scripts for updated signing keys

ptheywood:
See https://developer.nvidia.com/blog/updating-the-cuda-linux-gpg-repository-key/ for more information


Robadob (approved):
CI is passing again, so I guess it must work.

PR #845, opened 2022-05-06
Set NVRTC gpu-architecture flag to maximum supported version

### LINKED ISSUE #844: RTC models do not compile for unknown future CUDA Architectures
RTC models are compiled for the device's compute capability, i.e when running on a consumer Ampere GPU, nvrtc is passed `--gpu-architecture=compute_86`.

However, if the version of NVRTC does not know about the GPU architecture this will fail to compile, and the user can do nothing about this (other than use a more recent NVRTC)

This means that RTC models will not run on newer GPUs (without using newer features), unlike non-RTC models which will (via PTX embedding / JITing). 

To reproduce this, CUDA 11.0 knows SM_80 but not SM_86, so attempting to run a CUDA 11.0 RTC model on consume ampere will fail RTC compialtion, with an error during RTC compilation such as:

```
Compiler options: --gpu-architecture=compute_86 --generate-line-info -DNDEBUG --std=c++17 --define-macro=SEATBELTS=0 --pre-include=/usr/local/cuda-11.0/include//cuda.h 
```

ptheywood:
Set NVRTC gpu-architecture flag to maximum supported version for NVRTC & device

Closes #844

The maximum compute capability supported by the currently linked NVRTC that is less than or equal to the device's architecture is used for RTC compilation.

This fixes an issue where running an RTC model on consume ampere (SM_86) would fail on CUDA 11.0 and older, which are not aware of SM_86's existence.

CUDA 11.2+ includes methods to query which architectures are supported by the dynamically linked NVRTC (which may add or remove architectures in new releases, and due to a stable ABI from 11.2 for all 11.x releases the linked version can be different than the version available at compile time).
CUDA 11.1 and below (11.1, 11.0 and 10.x currently in our case) do not include these methods, and due to the absence of a stable nvrtc ABI for these versions the known values can be hardcoded at compile time (grim but simple).

A method to select the most appropriate value form an ascending order vector has also been introduced, so this gencode functionality can be programmatically tested without having to predict what values would be appropriate based on the current device and the cuda version used, which is a moving target.

---

CUDA 11.2, `SEATBELTS=OFF` tests pass for SM_86 device
```
[==========] 877 tests from 77 test suites ran. (48705 ms total)
[  PASSED  ] 877 tests.

  YOU HAVE 160 DISABLED TESTS
```

CUDA 10.2, `SEATBELTS=ON` tests pass for SM_86 device (all RTC tests would have failed previously). 

```
[==========] 997 tests from 79 test suites ran. (61473 ms total)
[  PASSED  ] 997 tests.

  YOU HAVE 40 DISABLED TESTS
```


src/flamegpu/util/detail/JitifyCache.cu
```diff
@@ -315,12 +317,20 @@ std::unique_ptr<KernelInstantiation> JitifyCache::compileKernel(const std::strin
...
+    if (nvrtcArchitectures.size()) {
+        int currentDeviceIdx = 0;
+        if (cudaSuccess == cudaGetDevice(&currentDeviceIdx)) {
+            int arch = compute_capability::getComputeCapability(currentDeviceIdx);
+            int maxSupportedArch = compute_capability::selectAppropraiteComputeCapability(arch, nvrtcArchitectures);
+            // only set a nvrtc compilation flag if a usable value was found
+            if (maxSupportedArch != 0) {
+                options.push_back(std::string("--gpu-architecture=compute_" + std::to_string(maxSupportedArch)));
+            }
+            // Optionally we could error here, but this *should* never occur
```
Robadob, line 331: That's what an `assert(false)` or `InternalError` is for. Adding the check would be harmless additional safety.

src/flamegpu/util/detail/compute_capability.cu
```diff
@@ -47,6 +52,56 @@ bool compute_capability::checkComputeCapability(int deviceIndex) {
...
+            // Return the populated std::vector, this should be RVO'd
+            return nvrtcSupportedArchs;
+        }
+    }
+    // If any of the above functions failed, we have no idea what arch's are supported, so assume none are?
+    return {};
+// Older CUDA's do not support this, but this is simple to hard-code for CUDA 11.0/11.1  (and our deprected CUDA 10.x).
+// CUDA 11.1 suports 35 to 86
+#elif (__CUDACC_VER_MAJOR__ == 11) && __CUDACC_VER_MINOR__ == 1
+    return {35, 37, 50, 52, 53, 60, 61, 62, 70, 72, 75, 80, 86};
```
Robadob, line 78: These are kinda grim, but not sure of a nicer option (besides just returning 35 by default).

Robadob (approved):
Looks like a solid PR. Haven't ran it locally, but assume your tests are sufficient.

src/flamegpu/util/detail/JitifyCache.cu
ptheywood: Added the `assert(false)`, users wouldn't be able to catch the exception so continuing with generic arch RTC seems like the better choice.

src/flamegpu/util/detail/compute_capability.cu
ptheywood: I aggree it is grim, but there is no way to programatically access request these values from nvrtc at runtime for these CUDA versions. 
Providing the full vector is the better choice, so that more recent features are used where appropriate.

Issue #816, opened 2022-03-03
DeviceAPI type checking support


Robadob:
CUDA does not support `typeid()` or `std::type_index`, this prevents use of the common type validation that is used all around the host and description APIs.

At current, we limit DeviceAPI/curve to checks of the type size (using `sizeof()`). However, the most common types are all 4 bytes, so this isn't too helpful. In some cases, you will be implicitly casting to the correct type within an agent function, and receive a compilation warning. However, this is not guaranteed, e.g. `float` being incorrectly accessed as `unsigned int`.

It would be possible to implement a more manual form of `typeid()` using templating:

```c
template <typename T>
struct cuda_typeid {
    // Unique id for the specified type
    static constexpr unsigned char id = 0;
};

template <>
struct cuda_typeid<char> { static constexpr unsigned char id = 1; };
template <>
struct cuda_typeid<bool> { static constexpr unsigned char id = 2; };
template <>
struct cuda_typeid<uint8_t> { static constexpr unsigned char id = 3; };
template <>
struct cuda_typeid<int8_t> { static constexpr unsigned char id = 4; };
template <>
struct cuda_typeid<uint16_t> { static constexpr unsigned char id = 5; };
template <>
struct cuda_typeid<int16_t> { static constexpr unsigned char id = 6; };
template <>
struct cuda_typeid<uint32_t> { static constexpr unsigned char id = 7; };
template <>
struct cuda_typeid<int32_t> { static constexpr unsigned char id = 8; };
template <>
struct cuda_typeid<uint64_t> { static constexpr unsigned char id = 9; };
template <>
struct cuda_typeid<int64_t> { static constexpr unsigned char id = 10; };
template <>
struct cuda_typeid<float> { static constexpr unsigned char id = 11; };
template <>
struct cuda_typeid<double> { static constexpr unsigned char id = 12; };
```

This could also be expanded to include string literals of type names, for the seatbelts errors.

As this is templatery, we'd need to update both curve, and the description API. So the mapping could be captured at both ends (or an update to curve which maps `std::type_index` to `cuda_typeid::id` when filling the table).

PR #846, opened 2022-05-06
Shared Memory Curve

### LINKED ISSUE #560: CUDA 11.3 Register Usage
Register usage for CUDA 11.3 appears to be significantly higher than previous cuda versions, especially for the iteration kernel in boids bruteforce. It is unclear where this register usage regresion has come from.

```
cmake .. -DCUDA_ARCH=70 -DSEATBELTS=OFF -DUSE_NVTX=ON
make -j 24 flamegpu2 boids_bruteforce
ncu --set=full -f -o 11-x ./bin/linux-x64/Release/boids_bruteforce -t -s 2
```

| CUDA Version | Reg/thread |
| ------------ | ---------- |
| 11.0         | 60         |
| 11.1         | 60         |
| 11.2         | 70         |
| 11.3         | 170        |

This is building for SM 70, as of 70c2e17.

LTO brings it down a little, but not significanlty (~156 ish) 

This may need to turn into an nvcc bug report when we run out of things to check. 

CUDA 11.3 introduces a way to dump the device callgraph at link time (the following cmake). This doesn't provide any useful information, just showing that the kernel is using 170 reg/thread (its 2 sub-calls both use < 30 reg, so its not a sub call issue.
```cmake
    add_link_options("$<DEVICE_LINK:SHELL:-Xnvlink -dump-callgraph>")
```

Additionally verbose ptx generation can be enabled, which outputs register use for each device function compiled, but this doesn't seem to provide much more information.

```cmake
    add_compile_options("$<$<COMPILE_LANGUAGE:CUDA>:SHELL:-Xptxas -v>")
```

### LINKED ISSUE #571: Use `__shared__` memory for non-rtc cuRVE
Currently, parts of cuRVE are stored in `__constant__` memory. 

We can instead store these in `__shared__` memory, and use a block stride loop + sync to read all the values from global memory into shared memory once per block.

This should improve performance (slightly, based on CINECA prototyping).

It may also have/enable some other advantages:
+ by removing device symbols from the library, removing RDC will become simpler if we wish to pursue this.
+ Rather than having a single device copy of curve for a given device context, by having a sepeare bit of global memory per simulation (or potentially per agent function in a layer) then serialisation can be reduced (in a single simulation, and in ensembles).
+ The maximum number of variables in curve could potentially be removed / increased / made dynamic, although this may harm performance. Templating may enable this with less of a performance compromise.

Robadob:
- [x] Curve to device ptr
- [x] Environment to device ptr
- [x] Curve to per-agent function
- [x] Curve to shared memory
- [x] Check/update pyflamegpu tests

Closes #560 
Closes #571


ptheywood:
As of [2a8f983](https://github.com/FLAMEGPU/FLAMEGPU2/pull/846/commits/2a8f983ca7278f9d9d38610337858731058de156), this is looking on track to resolve the register use in CUDA 11.3+. 

CUDA 11.6, SM_86, `boids_bruteforce` inputdata is using 46 reg/thread, compared to `145` reg/thread for current master. 

```
ptxas info    : Used 4 registers, 352 bytes cmem[0]
ptxas info    : Compiling entry function '_ZN8flamegpu22agent_function_wrapperI14inputdata_implNS_17MessageBruteForceENS_11MessageNoneEEEvjPKNS_6detail5curve5Curve10CurveTableEPKcjjjjPjjPKvSE_P17curandStateXORWOWSC_SC_SC_' for 'sm_86'
ptxas info    : Function properties for _ZN8flamegpu22agent_function_wrapperI14inputdata_implNS_17MessageBruteForceENS_11MessageNoneEEEvjPKNS_6detail5curve5Curve10CurveTableEPKcjjjjPjjPKvSE_P17curandStateXORWOWSC_SC_SC_
    0 bytes stack frame, 0 bytes spill stores, 0 bytes spill loads
ptxas info    : Used 46 registers, 456 bytes cmem[0]

```

I've set the circles benchmark going on a 3080 + titan v in CUDA 11.2 so we can see the performance loss in that benchmark from using global curve only as an intermediate step (will be a while before they complete).

ptheywood:
> I've set the circles benchmark going on a 3080 + titan v in CUDA 11.2 so we can see the performance loss in that benchmark from using global curve only as an intermediate step (will be a while before they complete).

This has resulted in an `invalid configuration argument` error during the comm-radius portion of the benchmark, on a 3080 using CUDA 11.2. Run 235/240 within that benchmark. 

```
2022-05-10T21:03:36Z: 235/240: circles_spatial3D 64000 40.000000 64000.000000 1.000000 40.000000 0
2022-05-10T21:03:36Z: 235/240: circles_spatial3D 64000 40.000000 64000.000000 1.000000 40.000000 0
terminate called after throwing an instance of 'flamegpu::exception::CUDAError'
  what():  /home/ptheywood/code/flamegpu/FLAMEGPU2-circles-benchmark/build-11-2-globalcurve/_deps/flamegpu2-src/include/flamegpu/gpu/detail/CUDAErrorChecking.cuh(27): CUDA Error: /home/ptheywood/code/flamegpu/FLAMEGPU2-circles-benchmark/build-11-2-globalcurve/_deps/flamegpu2-src/include/flamegpu/runtime/HostAgentAPI.cuh(899): invalid configuration argument
Aborted (core dumped)
```

Robadob:
Have updated `DeviceEnvironment::getProperty()`, such that the array element version requires the array length as a template argument. This is only used for validation internally, so it literally does nothing without seatbelts, however it provides much better consistency with the rest of the API (agent/message reads) where it is necessary.

I feel to enforce this consistency further, I will need to apply the same change to `HostAPI`/`CUDASimulation` environment methods.

Robadob:
As of fb6c322, have built ran test suite in all 3 main configs on Windows (Debug + Release with/without seatbelts). USE_GLM enabled too, in each case all tests pass.

Robadob:
Final 2 commits (denoted `?`) are optional queries.

ptheywood:
CUDA 11.4 A100 circles benchmark runs going for the last 3 commits (plus an alt version of the smem commit). Should finish overnight. 

Will show:

+ if shared is an improvment, (it should be for brute at least). 
+ if const restrict has a meaningful improvement (though as long as its not a degredation marking const restrict is valid anyway)
+ Whether smaller curve improves performance (partially occupancy related to smem use which circles might not be impacted by anyway, partially smaller curve table might be better anyway.

ptheywood:
A100 Data for the vairous subcommits, with alpha.2 v100 data for reference.
Some series are fully occluded by others. Previously discussed on slack. 

Shos it fixes the 11.3+ issue, and offers a good performance advantage.

fixed-density--agent_count--step-s--model--all
fixed-density--agent_count--step-s--model--zoomed


V100 data showing the < 11.3 performance relative to the alpha.2 base case. 

Global was a degredation for brute, but the smem 512 version is an improvement 

fixed-density--agent_count--step-s--model--all
fixed-density--agent_count--step-s--model--zoomed

src/flamegpu/gpu/CUDASimulation.cu
```diff
@@ -1636,12 +1562,14 @@ void CUDASimulation::initialiseSingletons() {
...
         if (DEVICE_HAS_RESET_CHECK == DEVICE_HAS_RESET_FLAG) {
             // Device has been reset, purge host mirrors of static objects/singletons
-            detail::curve::Curve::getInstance().purge(stream_0);
+            for (auto& ca : agent_map) {
+                ca.second->purgeCurve();
+            }
             if (singletons) {
                 singletons->rng.purge();
                 singletons->scatter.purge();
+                // @todo environment purge? Is this whole block still required?
```
ptheywood, line 1571: Does this need doign now?

ptheywood (changes requested):
I believe `include/flamegpu/util/StringUint32Pair.h` can now be deleted, as you've removed all use of it (I think). 

Otherwise looks good but I've left a few comments on lines that stood out. Most of them don't need anything doing, but it might be nice to address if not too much faff, or for the `@todo`'s promoting to issues might be worthwhile. 

I went through relatively quick though so could have missed some bits.

Once done + a little history cleaning it'll be good to go IMO

include/flamegpu/runtime/detail/curve/DeviceCurve.cuh
```diff
@@ -0,0 +1,389 @@
...
+class DeviceCurve {
+ public:
+    typedef Curve::Variable          Variable;                    // !< Typedef for cuRVE variable handle
+    typedef Curve::VariableHash      VariableHash;                // !< Typedef for cuRVE variable name string hash
+    typedef Curve::NamespaceHash     NamespaceHash;               // !< Typedef for cuRVE variable namespace string hash
+    static const int MAX_VARIABLES = Curve::MAX_VARIABLES;        // !< Default maximum number of cuRVE variables (must be a power of 2)static const VariableHash EMPTY_FLAG = Curve::EMPTY_FLAG;
+    static const VariableHash EMPTY_FLAG = Curve::EMPTY_FLAG;
+    static const int UNKNOWN_VARIABLE = -1;                       // !< value returned as a Variable if an API function encounters an error
+
+    // private:  // @todo TEMPORARY TO COMPILE PROGRESS
```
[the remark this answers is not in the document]
Robadob: Ah when I hadn't refactored env manager, i had to expose some internals. This shouldn't be necessary any more

include/flamegpu/runtime/detail/curve/HostCurve.cuh
```diff
@@ -0,0 +1,138 @@
...
+    friend class flamegpu::CUDAAgent;
+    /**
+     * Host and device storage of curve table
+     */
+    CurveTable h_curve_table, *d_curve_table;
+    /**
+     * Internal tracking of type_index of stored variables
+     * Not currently support in device code, but maintained incase future-support is added
+     */
+    // std::type_index h_curve_table_ext_type[MAX_VARIABLES]; //@todo How to init this safely?
```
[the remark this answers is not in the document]
Robadob: Yes and No. `std::type_index` doesn't have a default consructor, so I can't create an array of it.

I might follow it up in a later commit if I address #816

src/flamegpu/gpu/CUDASimulation.cu
Robadob: Good question.

Probably not, because if device gets reset after `CUDASimulation` has been init, there's probably going to be far more wrong than just the environment.

include/flamegpu/runtime/detail/curve/Curve.cuh
```diff
@@ -0,0 +1,100 @@
+#ifndef INCLUDE_FLAMEGPU_RUNTIME_DETAIL_CURVE_CURVE_CUH_
+#define INCLUDE_FLAMEGPU_RUNTIME_DETAIL_CURVE_CURVE_CUH_
+
+#ifndef __CUDACC_RTC__
+#include <string>
+#endif
+/**
+ * @file   Curve.cuh
+ * @author Paul Richmond (p.richmond@sheffield.ac.uk) http:// www.paulrichmond.staff.shef.ac.uk/
+ * @date   Feb 2017
```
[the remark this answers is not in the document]
Robadob: Happy to remove it, just left in legacy stuff

src/flamegpu/gpu/CUDASimulation.cu
Robadob: Have removed the comment, but will leave this unresolved incase you feel different.

Robadob: Yeah I think the whole `CUDASimulation::initialiseSingletons()` could do with a small rework, all? of the CUDA device reset stuff is nolonger relevant.

Robadob: Have removed all trace of purge, but left in auto `cudaDeviceReset()`, this will require an extra review from you.

ptheywood (approved):
Tests all pass under linux (SEATBELTS=OFF) and profiling still works so the device resetting is still solid. 

The other changes all look good. I'll let you merge in case you want to tidy history (the individual commits did make it very easy to review the changes though, so that's appreciated)

Robadob:
> Tests all pass under linux (SEATBELTS=OFF) and profiling still works so the device resetting is still solid.
> 
> The other changes all look good. I'll let you merge in case you want to tidy history (the individual commits did make it very easy to review the changes though, so that's appreciated)

The only real caveat with device resetting is that i've removed tracking of which devices `CUDASim` is init on.

So if someone has 2 independent `CUDASim` on different device, only the 2nd to end will be reset.

Likewise, if they have a `CUDASim` alive, when their `CUDAEnsemble` exits, if it's on same device as ensemble was using, it will be reset

PR #848, opened 2022-05-09
Fix `DeviceMacroProperty::operator+=(double)` for SM < 60 GPUs

### LINKED ISSUE #847: `DeviceMacroProperty::operator+=` does not compiler for SM < 60
`DeviceMacroProperty<T, I, J, K, W>::operator+=(const T& val)` uses `    atomicAdd(this->ptr, val);` internally. 

For `double` precision floating point numbers, `atomicAdd` is only implemented in CUDA for compute capability 60 devices (pascal) and newer, where the underlying hardware instruction was first implemented. 
This also requires CUDA >= ~8 IIRC, but that is not an issue for us. 

The CUDA documentation includes a reference implementation of `atomicAdd(double*, double)` using `atomicCAS`. This is much, much slower than the hardware instruction (especially when there is high atomic contention) but it is the only way to implement it for SM < 60. 

E.g. from the [CUDA 11.6 Documenation B.14](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#atomic-functions):
```
#if __CUDA_ARCH__ < 600
__device__ double atomicAdd(double* address, double val)
{
    unsigned long long int* address_as_ull =
                              (unsigned long long int*)address;
    unsigned long long int old = *address_as_ull, assumed;

    do {
        assumed = old;
        old = atomicCAS(address_as_ull, assumed,
                        __double_as_longlong(val +
                               __longlong_as_double(assumed)));

    // Note: uses integer comparison to avoid hang in case of NaN (since NaN != NaN)
    } while (assumed != old);

    return __longlong_as_double(old);
}
#endif
```

If the test suite had included use of doubles here, this would have been caught by CI. 

I've lazily added this to the `DeviceMacroPropertyTest.add` test in the `macroprop-addfp64` branch to demonstate this ([logs, valid for 90 days?](https://github.com/FLAMEGPU/FLAMEGPU2/actions/runs/2293382996)), which when targetting SM < 60 produces an error such as:

```
FLAMEGPU2/include/flamegpu/runtime/utility/DeviceMacroProperty.cuh(274): error: no instance of overloaded function "atomicAdd" matches the argument list
            argument types are: (double *, const double)
          detected during instantiation of "flamegpu::DeviceMacroProperty<T, I, J, K, W> &flamegpu::DeviceMacroProperty<T, I, J, K, W>::operator+=(const T &) [with T=double, I=1U, J=1U, K=1U, W=1U]" 
FLAMEGPU2/tests/test_cases/runtime/test_device_macro_property.cu(95): here

1 error detected in the compilation of "FLAMEGPU2/tests/test_cases/runtime/test_device_macro_property.cu"
```

We could just drop the reference implementation into the `DeviceMacroProperty` header outside of the `flamegpu` namespace, but if this is done anywhere else it would be multiply defined. 
Usign the anon namespace instead would allow this to coexist with other implementations, but would require it a little bit of macro use. I'm not sure which would be the cleaner solution.

ptheywood:
+ Modifies a test to demonstrate the compilation error (and a subsequent test to split out a previously shared agent fn)
+ Adds the `atomicAdd(double*, double) implementation using atomicCAS, based on the cuda docs reference.
  + Some linter suppression due to `atomicCAS(uint64_t*, ...)` not compiling. 3 suppressions is cleaner than 1 suppression + many reinterprets.

Closes #847

### DIFF
--- a/include/flamegpu/runtime/utility/DeviceMacroProperty.cuh
+++ b/include/flamegpu/runtime/utility/DeviceMacroProperty.cuh
@@ -5,6 +5,30 @@
 #include <limits>
 #include <algorithm>
 
+/**
+ * CUDA does not implement atomicAdd(double*, double) for pre-pascal GPUs, which do not have the underlying hardware instruction.
+ * A CAS based software implementation is required instead, which will be much slower.
+ * This implementation is based on the reference implementation prodived by the CUDA toolkit documentation.
+ */
+#ifdef __CUDACC__
+#include <cuda_runtime.h>
+// Needs to be mutually exclusive with definitions in CUDA's sm_60_atomic_functions.h
+#if defined(__CUDA_ARCH__) && __CUDA_ARCH__ < 600
+__device__ __forceinline__ double atomicAdd(double* address, double val) {
+    // cpplint enforces uint64_t, but atomicCAS is implemented for unsigned long long int
+    unsigned long long int* address_as_ull = reinterpret_cast<unsigned long long int*>(address);  // NOLINT(runtime/int)
+    unsigned long long int old = *address_as_ull;  // NOLINT(runtime/int)
+    unsigned long long int assumed = old;  // NOLINT(runtime/int)
+    do {
+        assumed = old;
+        old = atomicCAS(address_as_ull, assumed, __double_as_longlong(val + __longlong_as_double(assumed)));
+    // Note: uses integer comparison to avoid hang in case of NaN (since NaN != NaN)
+    } while (assumed != old);
+    return __longlong_as_double(old);
+}
+#endif  // __CUDA_ARCH__ < 600
+#endif  // __CUDAACC__
+
 namespace flamegpu {
 
 /**


ptheywood:
Squash merging is recommended, the separate commits are to demonstrate the error via CI

tests/test_cases/runtime/test_device_macro_property.cu
```diff
@@ -87,10 +87,12 @@ TEST(DeviceMacroPropertyTest, DISABLED_ReadWrite) {
 
 FLAMEGPU_AGENT_FUNCTION(Init_add, flamegpu::MessageNone, flamegpu::MessageNone) {
     FLAMEGPU->environment.getMacroProperty<unsigned int>("int").exchange(1);
+    FLAMEGPU->environment.getMacroProperty<double>("double").exchange(1);
     return flamegpu::ALIVE;
 }
 FLAMEGPU_AGENT_FUNCTION(Write_add, flamegpu::MessageNone, flamegpu::MessageNone) {
     FLAMEGPU->environment.getMacroProperty<unsigned int>("int") += FLAMEGPU->getVariable<unsigned int>("a");
+    FLAMEGPU->environment.getMacroProperty<double>("double") += static_cast<double>(FLAMEGPU->getVariable<unsigned int>("a"));
```
Robadob, line 95: If you're going to update the test, atleast update it fully so the result is checked too.

Robadob (approved):
Haven't ran it again, but you fixed the 2 failing tests I DM'd you about on slack so should be good.

PR #849, opened 2022-05-09
Enable -lineinfo for MinSizeRel and RelWithDebInfo

### LINKED ISSUE #798: CMake CUDA flags for MinSizeRel RelWithDebInfo
We may wish to adjust our CMake to set certain values for `MinSizeRel` and `RelWithDebInfo` builds which are currently only set for Debug or Release builds as appropriate. 

I.e. the default value of SEATBELTS, use of `-lineinfo` and `-G` etc. 

This is unlikely to effect most users (who will likely use Debug or Release exclusively) but would be nice to improve (at least `RelWithDebInfo`). 

We probably always want to set `-lineinfo`, unless `-G` is set. 

It's less clear when `(_)DEBUG`/`(_)NDEBUG` should be set which might need some thought / reading. 

The easiest way to find current non-default cmake config specific options is to search for `CONFIG:`. Due to IDE style tools such as Visual Studio, we should only use generator expressions for this type of thing rather than using `CMAKE_BUILD_TYPE` anywhere.

ptheywood:
Enable -lineinfo for MinSizeRel and RelWithDebInfo

Includes a couple of other very minor CMake tweaks

Unfortunately cannot use the multi-argument form of `$<CONFIG:` as this requires CMake >= 3.19

Closes #798

### DIFF
--- a/cmake/common.cmake
+++ b/cmake/common.cmake
@@ -43,13 +43,7 @@ include(${CMAKE_CURRENT_LIST_DIR}/cxxstd.cmake)
 
 # Set a default build type if not passed
 get_property(GENERATOR_IS_MULTI_CONFIG GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
-if(${GENERATOR_IS_MULTI_CONFIG})
-    # CMAKE_CONFIGURATION_TYPES defaults to something platform specific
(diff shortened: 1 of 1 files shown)


mondus (approved):
Beautiful soup

PR #850, opened 2022-05-09
Drop CUDA 10.x and C++14

### LINKED ISSUE #611: Drop CUDA 10.x support
As CUDA 11.x has been available for for 15 months now,  it is avialable in most places we need it to be (although there are still a few HPC systems which we use where it is not quite available). 

Although I have generally wanted to keep CUDA 10.x support in the past for compatability, I'm becoming more and more keen on dropping 10.x support for several reasons.

+ C++ 17 only
    + For use of actual `<filesystem>
    + `constexpr` enhancements, including `if constexpr`
        + should allow us to replace some macros with namespaced constexpr's without any runtime cost.
    + simplifed nested namespaces, i.e. `namespace flamegpu::util {` doesn't require two expressions.
    + `<any>`, removing the need for `flamegpu::util::Any`
    + `static_assert` changes that might remove the need for `flamegpu::util::detail::StaticAssert`
    + Probably others, see https://en.cppreference.com/w/cpp/17
    + Remove some complicated CMake logic choosing which c++ version to specify (which is duplicated across 2 repos)
+ We can drop a bunch of CUDA 10.x workarounds / warnings in our CMake
+ Reduces the build matrix required by CI (or atleast shifts it).

If we can, the time to do this is before we have a full non-prerelease release.

Outstanding HPC sites i'm aware of that we have access to and FLAME GPU 2.x is already / expected to be used on:

+ [ ] TUoS Bessemer
    + Driver for 11.0 deployed, but nvcc module was never made available.
    + ShARC recently updated to 11.3 + relevant cudnn. Will chase this for Bessemer too. 
+ [ ] N8CIR Bede
    + Available in RHEL8 test partitiions. Should be deployed in general within a few months. 
+ [ ] Prometheus (plgrid)
    + Unsure on what CUDA version is available, although likely 10+ for volta GPUs.

ptheywood:
Remove the deprecated support for CUDA 10.x and C++17, upgrading the minimum versions to:

+ GCC >= 8.1
+ CUDA >= 11.0
+ Visual Studio 2019

This also slightly reduces our CI workload, and removes some workarounds for older compilers from CMake. 

By removing C++14 support, we can also make use of C++17 features

+ [x] `std::filesystem`
+ [ ] `std::any`
+ [ ] `std::from_chars`
+ [ ] `std::string_view`?

This will also allow the use of structured bindings to simplify map iteration etc moving forwards.

Requires a matching PR in the visualisation repository.

+ [ ] `FLAMEGPU/FLAMEGPU2-visualiaser#@todo`

Closes #611.


Robadob:
>  std::any

Fairly sure there's a manual implementation of any somewhere, probably env
manager.

ptheywood:
> Fairly sure there's a manual implementation of any somewhere, probably env
manager.

Yep, the plan is to replace it.

ptheywood:
I've deffered `std::any` to a future PR, as it would be very conflicty with ongoing other work for now (which opting into c++17 features might make nicer). 
The visualiser repo hash could be updated to the merged hash for the matching PR, however CI builds fine without the matching change (as it's removal not addition).

The "best" merge order would be, this PR, the vis PR, then another Pr updating the hash, but that's a lot of effort for very low risk of harm.

ptheywood:
> I guess `recursive_create_dir()` from the deleted util/filesystem header wasn't actually being used and was inherited from visualiser.

It was, but `std::filesystem::create_directories` is a direct replacement (compared to `std::filesystem::create_directory` which would fail if the parent directory does not exist).

Not sure if it existed in the experimental/tr2 versions or not.

PR #851, opened 2022-05-10
Add CMake option for verbose ptxas output

ptheywood:
Adds the `VERBOSE_PTXAS` cmake configuration option, which enabled verbose PTXAS output during compilation when opted in. 

Most users will not need this, but as I keep having to make this change to look at reg usage during compilation, it's now a PR.


Robadob (approved):
Seems fine, haven't tested it.

More general option would be to expose CMake vars to manually specify additional args to each of the compiler stages.

Robadob (commented):
https://cmake.org/cmake/help/latest/command/mark_as_advanced.html

Possibly worthwhile (for CMake GUI users).

PR #853, opened 2022-05-11
Functional but hacky initial implementation of Schelling's segregation model

MILeach:
This works but currently relies on some questionable code, specifically using cuda thread index to allow an unusual access pattern to a message list.


examples/schelling_segregation/src/main.cu
```diff
@@ -0,0 +1,357 @@
...
+     * Movement resolution submodel
+     */
+
+    flamegpu::ModelDescription submodel("plan_movement");
+    submodel.addExitCondition(movement_resolved);
+    {
+        // Environment
+        {
+            flamegpu::EnvironmentDescription& env = submodel.Environment();
+            env.newProperty<unsigned int>("spaces_available", 0);
```
Robadob, line 162: If this is intended to be user configurable/accessible, you will need to define it in the parent model and map it too.

examples/schelling_segregation/src/main.cu
```diff
@@ -0,0 +1,357 @@
...
+    FLAMEGPU->setVariable<unsigned int>("available", my_availability);
+
+    return flamegpu::ALIVE;
+}
+
+FLAMEGPU_AGENT_FUNCTION_CONDITION(is_available) {
+    return FLAMEGPU->getVariable<unsigned int>("available");
+}
+FLAMEGPU_AGENT_FUNCTION(output_available_locations, flamegpu::MessageNone, flamegpu::MessageArray) {
+    FLAMEGPU->message_out.setIndex(FLAMEGPU->getThreadIndex());
```
Robadob, line 55: Why is `getThreadIndex()` still required? I doubt Paul will like this being used.

examples/schelling_segregation/src/main.cu
```diff
@@ -0,0 +1,357 @@
...
+
+    // Bid for that location
+    FLAMEGPU->message_out.setKey(selected_location - 1);
+    FLAMEGPU->message_out.setVariable<flamegpu::id_t>("id", FLAMEGPU->getID());
+    FLAMEGPU->message_out.setVariable<unsigned int>("type", FLAMEGPU->getVariable<unsigned int>("type"));
+    return flamegpu::ALIVE;
+}
+
+FLAMEGPU_AGENT_FUNCTION(select_winners, flamegpu::MessageBucket, flamegpu::MessageArray) {
+    // First agent in the bucket wins
```
Robadob, line 84: More expensive, but technically more 'fair' to have agents roll a random float which they attach to the message, and the highest/lowest float wins.

examples/schelling_segregation/src/main.cu
Robadob: I guess it's because you only have a random selection of locations available, which aren't a contiguous block. So you need a way agents can identify valid spots, and this is a fairly simple method.

Not sure of a better approach off the top of my head, maybe one for discussion in the meeting.

examples/schelling_segregation/src/main.cu
```diff
@@ -0,0 +1,357 @@
...
+    unsigned int my_next_type = ((my_type != UNOCCUPIED) && isHappy) ? my_type : UNOCCUPIED;
+    FLAMEGPU->setVariable<unsigned int>("next_type", my_next_type);
+    FLAMEGPU->setVariable<unsigned int>("movement_resolved", (my_type == UNOCCUPIED) || isHappy);
+    unsigned int my_availability = (my_type == UNOCCUPIED) || (isHappy == 0);
+    FLAMEGPU->setVariable<unsigned int>("available", my_availability);
+
+    return flamegpu::ALIVE;
+}
+
+FLAMEGPU_AGENT_FUNCTION_CONDITION(is_available) {
```
Robadob, line 51: Might benefit from a comment above this, to denote the remaining functions are from the iterative submodel.

(My initial thought was to split the entire submodel into a seperate header or source file, like I've done in primage to aid readability, but probably overkill for the small model (and I doubt I did it for sugar scape).

Robadob (commented):
You can probably rename the PR, it's nolonger particularly hacky

examples/schelling_segregation/src/main.cu
Robadob: *On reviewing, I don't think it is, so this can be marked resolved.*

examples/schelling_segregation/src/main.cu
```diff
@@ -0,0 +1,357 @@
...
+#include <cstdio>
+#include <cstdlib>
+#include <fstream>
+#include <array>
+
+#include "flamegpu/flamegpu.h"
+
+// Configurable properties
+constexpr unsigned int GRID_WIDTH = 100;
+constexpr float THRESHOLD = 0.70;
```
Robadob, line 11: `0.70f` (This is the Windows CI fail)

examples/schelling_segregation/src/main.cu
Robadob: Having thought about this, I guess the 'proper' way would be for the `spaces_available` property to be binned, and a `spaces_available` brute force message list created.

Spaces available output a message which only contains their ID, Agents then grab the list, check it's length and generate a random number to grab the message at that index. *(Brute force messages don't currently have direct access like an array, so would need iterating, but there's no practical reason it couldn't be added).*

One benefit of this 'proper' version is that it would better facilitate agents with more specific preferences, as the `spaces_available` message could be improved to list any number of features of the space which the agent must approve of. (Although we might want to think about a more generalised version of Spatial messaging for multi-dimensional data if this were to be a real thing).

examples/schelling_segregation/src/main.cu
```diff
@@ -0,0 +1,357 @@
...
+            agent.newVariable<unsigned int, 2>("pos");
+            agent.newVariable<unsigned int>("type");
+            agent.newVariable<unsigned int>("next_type");
+            agent.newVariable<unsigned int>("happy");
+            agent.newVariable<unsigned int>("available");
+            agent.newVariable<unsigned int>("movement_resolved");
+
+            // Functions
+            auto& outputLocationsFunction = agent.newFunction("output_available_locations", output_available_locations);
+            outputLocationsFunction.setMessageOutput("available_location_message");
```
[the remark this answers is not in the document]
ptheywood: It'll depend on how complex the function with the branching is, how big the populations are, if there's room for concurrency by splitting the pop, and how grouped together the agents which follow each branch would be (plus other things). It's something that we should probably investigate at some point to make recommendations about when to use it (if it follows a sane pattern wrt the abov epoints)

ptheywood:
A revised / FLAME GPU 2.0.0-rc compatible version of this is in FLAMEGPU/ABM_Framework_Comparisons. 


https://github.com/FLAMEGPU/ABM_Framework_Comparisons/blob/FLAMEGPU2/FLAMEGPU2/src/schelling.cu

The size will want to be reduced.

PR #856, opened 2022-05-12
Update CI to use CUDA 11.7

ptheywood:
Updates the recent CUDA CI jobs to use 11.7(.0) which has been released. This is just to ensure we don't have any 11.7 specific issues. 

Also updates the list of known CUDA windows urls to be a little more thorough (with 11.5.1 and the release of windows 11 being the most recent change in pattern)

CUDA 11.7 still distributes Thrust/CUB 1.15, which cannot be used on Windows so must be blocked via CMake.


Robadob (approved):
Assuming CI passes should be good

PR #863, opened 2022-05-26
Replace std::default_random_engine with std::mt19937_64

### LINKED ISSUE #754: Do not use std::default_random_engine
`std::default_random_engine` is implementation defined, so different sequences are used on different platforms which is not ideal.

Most occurences were updated to use `std::mt19937_64`, the 64 bit mersenne twister implementation in the standard library, but there are currenlty 39 occurances in the test suite of the default being used (likely in PR's or based on tests from before the change to 64 bit mt was made). 

These should be updated.

ptheywood:
Closes #754


Robadob (approved):
All changes are the same, so should be safe.

You also confirmed you've ran tests to make sure changing the rng hasn't managed to break any bad tests.

mondus (approved):

PR #864, opened 2022-05-27
Improve how CUDAEnsemble reports failure to find a CUDA device.

### LINKED ISSUE #858: Improve CUDAEnsemble device selection error messages
CUDAEnsemble does not provide useful error messages when no CUDA devices are present or the CUDA runtime cannot be created, instead just throwing the error to the use via `gpuErrchk`

https://github.com/FLAMEGPU/FLAMEGPU2/blob/f70bac09e3985ab7b444680e90c6b142bcdce69f/src/flamegpu/gpu/CUDAEnsemble.cu#L75

This should be improved to match the more complex `CUDASimulation` device selection logic that throws exceptions with useful messages.

https://github.com/FLAMEGPU/FLAMEGPU2/blob/f70bac09e3985ab7b444680e90c6b142bcdce69f/src/flamegpu/gpu/CUDASimulation.cu#L1546-L1575

Robadob:
Pretty simple, @ptheywood identified the same code in `CUDASimulation` in the issue. Also added an early check that requested device IDs are in bounds.

Tested it compiles, but not done any execution (or testing?).

Closes #858

### DIFF
--- a/src/flamegpu/gpu/CUDAEnsemble.cu
+++ b/src/flamegpu/gpu/CUDAEnsemble.cu
@@ -71,14 +71,26 @@ unsigned int CUDAEnsemble::simulate(const RunPlanVector &plans) {
     run_logs.clear();
     run_logs.resize(plans.size());
     // Workout how many devices and runner we will be executing
-    int ct = -1;
-    gpuErrchk(cudaGetDeviceCount(&ct));
+    int device_count = -1;
+    cudaError_t cudaStatus = cudaGetDeviceCount(&device_count);
+    if (cudaStatus != cudaSuccess) {
+        THROW exception::InvalidCUDAdevice("Error finding CUDA devices!  Do you have a CUDA-capable GPU installed?, in CUDAEnsemble::simulate()");
+    }
+    if (device_count == 0) {
+        THROW exception::InvalidCUDAdevice("Error no CUDA devices found!, in CUDAEnsemble::simulate()");
+    }
+    for (const int &id : config.devices) {
+        if (id >= device_count) {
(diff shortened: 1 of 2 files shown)


ptheywood (approved):
Testing for no devices wouldn't really be fesible as a unit test, though it would be feasible to check for a too high device number (by requesting num devices +1, or by testing an arbitrarily high number. With MIG this would need to be > 16 * 7 if hardcoded). It's not really neccesary though. 

There's some logic in CUDAEnsemble I'd missed preivously, the CLI catches the device count rather than later on, so via the `-d` flag this `printf` is hit rather than the new exception.
Setting the config value programatially would still trigger the exception though (but that would require a test rahter than me just running an example)

https://github.com/FLAMEGPU/FLAMEGPU2/blob/51bd3e37c244a768aa4e43384cc1b3f1aceb2699/src/flamegpu/gpu/CUDAEnsemble.cu#L260-L265

When running with `CUDA_VISIBLE_DEVICES=` it triggers the new user-friendly exception:

```
CUDA_VISIBLE_DEVICES= ./bin/Release/ensemble 
terminate called after throwing an instance of 'flamegpu::exception::InvalidCUDAdevice'
  what():  /home/ptheywood/code/flamegpu/FLAMEGPU2/src/flamegpu/gpu/CUDAEnsemble.cu(77): Error finding CUDA devices!  Do you have a CUDA-capable GPU installed?, in CUDAEnsemble::simulate()
```

Robadob:
Added (and tested) a new test for out of bounds device ID.

ptheywood (approved):

mondus (approved):

PR #866, opened 2022-06-02
Add Visual Studio 2022 to the CI build matrix.

Robadob:
Well this works.
Had to fix a single new warning (due to an `assert()` statement making some vars unused in Release mode).
- [x] Need to decide if we want CUDA 11.7 VS2022 & VS2019, or just VS2022 with suffice (as we have CUDA 11.0 VS 2019). The latter seems preferable to me, given sometimes the dependency fetches can be temperamental, so less points of failure seems like a good thing. But will leave the call to @ptheywood.
- [ ] Need to squash this PR.

### DIFF
--- a/.github/workflows/Windows-Tests.yml
+++ b/.github/workflows/Windows-Tests.yml
@@ -25,8 +25,8 @@ jobs:
         cudacxx:
           - cuda: "11.7.0"
             cuda_arch: "35"
-            hostcxx: "Visual Studio 16 2019"
-            os: windows-2019
+            hostcxx: "Visual Studio 17 2022"
+            os: windows-2022
           - cuda: "11.0.3"
             cuda_arch: "35"
             hostcxx: "Visual Studio 16 2019"
--- a/.github/workflows/Windows.yml
+++ b/.github/workflows/Windows.yml
@@ -35,7 +35,7 @@ jobs:
             os: windows-2022
           - cuda: "11.7.0"
             cuda_arch: "35"
-            hostcxx: "Visual Studio 17 2019"
+            hostcxx: "Visual Studio 16 2019"
             os: windows-2019
           - cuda: "11.0.3"
             cuda_arch: "35"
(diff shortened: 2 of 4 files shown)


ptheywood:
2019 + 11.0 and 2022 + 11.7 seems fine to me given the windows CI is slow / more tempremental. 

Will be worth updating the manually invokable windows tests CI to also fire off a 2022 job. 

I'll make that change now and push to this branch.

ptheywood (approved):
Approved subject to squashing once CI is happy (hopefully it will be). 

Manually invoked the windows test CI too (which may potentially raise warnings if any VS2022 specific warnings are present in the test suite): 

https://github.com/FLAMEGPU/FLAMEGPU2/actions/runs/2530289219

ptheywood:
> > Manually invoked the windows test CI too (which may potentially raise warnings if any VS2022 specific warnings are present in the test suite):
> 
> There are a handful of warnings which Matt promised to fix, I've fixed them in a different PR [here](https://github.com/FLAMEGPU/FLAMEGPU2/pull/873/commits/5b84d2984722fe3ecf8715982304b7cae981c1a2) (It's the curand one), not yet merged.

It won't be reported as a failure on this PR's checks (I don't think) so shouldn't block merging, but will let us test 2022 more easily in the future / any branches rebased onto this.

ptheywood:
The manually invoked CI caught one additional warning in the test suite  only on the 2022 CI, though it may not be the only one depending on file ordering etc. 

> [build-tests-windows (11.7.0, OFF, Release, windows-2022): tests/test_cases/runtime/test_host_agent_creation.cu#L548](https://github.com/FLAMEGPU/FLAMEGPU2/commit/afe0f734133e5d8e396cb54c532918dfd6b4754c#annotation_3795864097)
'j': local variable is initialized but not referenced [D:\a\FLAMEGPU2\FLAMEGPU2\build\tests\tests.vcxproj]

It might be worth manually building the test suite under VS2022 to check if / when someone has a local install handy and a bit of time.

Robadob:
>  VS2022 to check if / when someone has a local install handy

Might be a while.

PR #883, opened 2022-07-08
Network message

Robadob:
# Stage 1 - Static environment graph
- [x] Description
- [x] Host Set
- [x] Host Get
- [ ] Host Misc (e.g. from file)
- [x] Agent Get
- [ ] Agent Traverse
- [ ] *(internal)* Proper backing graph data structure/sort
- [x] Device Curve
- [ ] RTC Curve
- [ ] Submodel support

# Stage 2  - Messages for the graph
- [ ] Everything


Robadob:
I don't really like how inside `EnvironmentDirectedGraph::CUDABuffers`, a reference to the map of agents is held, so that when buffers are resized curve can be updated in each agent function. Something about this feels wrong.

Alternate options being:
* A central set of `HostCurve` objects is maintained in parallel which can be shared.
* Everytime an agent function is loaded, it re-sets the graph ptrs in curve, although they won't change beyond init in most cases.

Robadob:
Current bug, when master model causes graph buffers to be allocated, submodel curve is not updated. This is a side-effect of how, the `EnvironmentDirectedGraph` tracks the agent map of it's owning model.

In general, I think this means agent access to directed graphs from submodels does not currently work.

Robadob:
Py Access violation isn't present with Python 3.10.4 (as opposed to 3.7.6). So calling it a Python bug and moving on.

Robadob:
Latest commit adds support for iterating edges which leave a specific vertex.

In completing this I fixed/identified a couple of bugs leading to two tings needing fixing.
* After edges are sorted to create the PBM, the host copy of the edges needs to be updated
* Several tests are now failing, due to the edge reorder
```
[  FAILED  ] 3 tests, listed below:
[  FAILED  ] TestEnvironmentDirectedGraph.TestDeviceGetResetGet
[  FAILED  ] TestEnvironmentDirectedGraph.SubModelSet_MasterModelDeviceGet
[  FAILED  ] TestEnvironmentDirectedGraph.MasterModelSet_SubModelDeviceGet
(and probably others after the first issue is addressed).
```

Robadob:
I built pre-rebase version too and it had the same test failure, so will fix that next.

Robadob:
This is now passing CI again. There are a handful of test failures (within the graph test suite). To the best of my understanding, these are a result of edges being sorted due to the later CSR changes. This is alluded to by one of the squashed commit messages.

Robadob:
Have now rebased the branch vs master (post file shuffle). I tested that I was able to build the test suite and pyflamegpu before resolving the final conflict. So the prior 5 commits likely don't build and will want squashing before this is merged (but I think that's a long way off).

Robadob:
Have reduced failing tests to `TestEnvironmentDirectedGraph.TestHostSetGet`, not immediately clear why this is still failing as it uses the same agent/host functions as I corrected for the previous tests. There may be an actual bug.

Robadob:
All current `TestEnvironmentDirectedGraph` tests are now passing. This was solved by defining edge source-dest pairs in the same order they would be sorted into.

It's assumed subsequent tests will cover the traversal of graphs, so the actual sort will be tested implicitly.

However, we may wish to remove indexable edge access in future, as the sort will lead to confusion??

Robadob:
Looking at `EnvironmentDirectedGraphDevice.cuh`, it would appear alongside edge sorting, support for device side traversal was added via the method `outEdges()`. When passed a vertex ID, it is supposed to return an iterator of the edges available.

Will next work on adding some tests for this.

There are some *todo* in the code related with how to handle SEATBELTS for these methods.

Robadob:
RTC now works, but need to think about `SEATBELTS` for PBM access (in both C/RTC). There's the below mock inside `EdgeFilter`s constructor.

```
            // Todo, how will we verify that vertexID is in range?}
            // If key is in bounds
            // if (beginKey >= metadata->min && endKey < metadata->max && beginKey <= endKey) {
            //    bucket_begin = metadata->PBM[beginKey - metadata->min];
            //    bucket_end = metadata->PBM[endKey - metadata->min];
            // }
```

We don't currently have `metadata` for the graph on device. It's not clear to me why `min` would be required either, we simply need to check vertexID is less than the number of vertices.


This needs some thought, e.g. how does vertex access check vertex id is in range.

Robadob:
Both test suites pass (Windows/Release/SEATBELTS=ON)

Following the removal of nested classes, it might be appropriate for me to split the test suites. However it only appears it would be split into host/device at current. I'm not too eager to do that, as there currently aren't too many tests.

Robadob:
ID currently exists as an automatically added Vertex variable. But it's not used for anything?

Robadob:
The HostAPI auto sync could be improved.

If the user calls
```
std::shared_ptr<flamegpu::detail::CUDAEnvironmentDirectedGraphBuffers> map = FLAMEGPU->environment.getDirectedGraph("map").getCUDABuffers();
```

The graph will be sync'd, prior to them accessing the CUDA buffers (ignoring the fact this doesn't mark the graph for rebuild). This however would require some proper internal tracking to ensure it persists to the end of the host fn.

Robadob:
I don't think it validates that edge src/dest are within range of valid vertex indices.

Robadob:
Deprecated by https://github.com/FLAMEGPU/FLAMEGPU2/pull/1089, to avoid breaking FRE dependency.

PR #887, opened 2022-07-27
Incremental pyflamegpu/codegen builds

ptheywood:
Fixes incremental builds when `swig/python/codegen` files are changed.
Adds a new intermediate `pyflamegpu_swig` target - the expensive bit
Adds a custom target `pyflamegpu`, which builds the wheel (and venv)
Dependency relationships set up to correctly perform incremental builds.


ptheywood:
Manually tested under windows and linux. Would be good if @mondus could test locally (and will be worth me re-checking linux when I'm back in it after some windows specific fixes that could have unintentionally broken linux)

ptheywood:
Workaround for non-multiconfig generators < CMake 3.20 has been fixed, so CI will now pass. 

**But** This will still not build under windows (or eclipse) with CMake <= 3.20, so a better fix needs to be found (i.e. grim CMake). Not going to get around to that today.

ptheywood:
@mondus This now includes the early error for CMake < 3.20 using multi-config generators when `BUILD_SWIG_PYTHON=ON`. 

Current CI failures are due to a CMake 3.24 change emitting a new developer warning which our CI gets intentionally angry about. It should only impact the manylinux job too currently (as regular gha images haven't updated CMake yet)

It requires a vis repo change, so that will be it's own PR, that the `python-ast` branch will want rebasing agaisnt eventually .

mondus (approved):
Happy with this having done a walk through with Pete.

PR #896, opened 2022-08-06
Visualise Random Seed and Environment Properties

Robadob:
A draft of it now works

F1 for the existing debug menu, with random seed added
image

F2 for environment properties which update with the model (albeit no thread safety reading the data ptrs from vis thread)
image

- [ ] Seems worth hiding ones that begin with `_` given they're supposed to be internal.
- [ ] I think if they have a huge array it will wrap, but should probably test this (I may need to specify max width of the box)
- [ ] I need to add some kind of scroll if they have loads (e.g. primage).
- [ ] Will need to update CMake to point to new Vis hash when that is ready (reason CI is failing).

Matching vis PR: https://github.com/FLAMEGPU/FLAMEGPU2-visualiser/pull/94

### DIFF
--- a/include/flamegpu/gpu/CUDASimulation.h
+++ b/include/flamegpu/gpu/CUDASimulation.h
@@ -51,6 +51,9 @@ class CUDASimulation : public Simulation {
     friend class HostAgentAPI;
     friend class SimRunner;
     friend class CUDAEnsemble;
+#ifdef VISUALISATION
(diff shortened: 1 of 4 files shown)


Robadob:
#911 (ImGui user interfaces) has superseded this more basic option. Can always add an extra UI listing all environment properties if there's demand for it.

PR #904, opened 2022-08-15
Visualiser: Support for keyframe-animated models.

Robadob:
Todo:
* Why does [this](https://github.com/FLAMEGPU/FLAMEGPU2-visualiser/blob/127af121f6046163cfb64923cf096f3ee2d74fa0/src/flamegpu/visualiser/shader/Shaders.cpp#L518) need to be called twice? (The VAD is zero'd if this line is removed)
* Write some docs
* Implement a ped model
* Update this PR to point to updated vis version

Vis PR: https://github.com/FLAMEGPU/FLAMEGPU2


Robadob:
Current state of the ped nav:
https://youtu.be/PkDOlpK9OAM

Obvious bugs right now.
* Some agents at the curved entrance get stuck facing backwards, this could be due to NaN somehow.
* There are rectangular areas on the two sides, where agents just stay within the rectangle, unclear of the cause.
* Avoid weight seems to have negligible impact.

Missing features
* Can't yet visualise the navmap arrows
* Haven't got support to add UI elements for updating the environment properties.

Robadob:
https://youtu.be/Xf5ZdwtlyLg

ptheywood (approved):

PR #910, opened 2022-08-26
Agent functions conditions in native python

### LINKED ISSUE #909: Support for Agent Function conditions in pure/agent Python
Currently not a supported feature (noted in the docs). Would require support for agent device functions so not considerably different to standard agent functions.

mondus:
Fixes #909 

Requires the following
 - [x] Support within the code generator
 - [x] Unit testing in the code generator
 - [x] Integration test
 - [x] Documentation PR (https://github.com/FLAMEGPU/FLAMEGPU2-docs/pull/112)

### DIFF
--- a/swig/python/codegen/codegen.py
+++ b/swig/python/codegen/codegen.py
@@ -710,14 +710,14 @@ class CodeGenerator:
 
     def _FunctionDef(self, t):
         """
-        Checks the decorators of the function definition much must be either 'pyflamegpu.agent_function' or 'pyflamegpu.device_function'.
+        Checks the decorators of the function definition much must be either 'pyflamegpu.agent_function', 'pyflamegpu.agent_function_condition' or 'pyflamegpu.device_function'.
         Each is then processed in a different way using a specific dispatcher.
         Function calls are actually checked and only permitted (or user defined) function calls are supported.
         """
         self.write("\n")
         # check decorators
         if len(t.decorator_list) != 1 or not isinstance(t.decorator_list[0], ast.Attribute):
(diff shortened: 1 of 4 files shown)


Robadob:
As mentioned in meeting, AgentFunctionCondition's don't have access to any setters (any methods that can change state). Their API object is instead [`ReadOnlyDeviceAPI`](https://github.com/FLAMEGPU/FLAMEGPU2/blob/7e8d7b0ae14d1333802c3f1d4c0d695071edf8fd/include/flamegpu/runtime/DeviceAPI.cuh#L39)

It doesn't look like there's anything to handle this gracefully, or is the plan to leave users at the whim of C++ errors (I'm not sure how bad they would be, probably just method x does not exist).

mondus:
> As mentioned in meeting, AgentFunctionCondition's don't have access to any setters (any methods that can change state). Their API object is instead [`ReadOnlyDeviceAPI`](https://github.com/FLAMEGPU/FLAMEGPU2/blob/7e8d7b0ae14d1333802c3f1d4c0d695071edf8fd/include/flamegpu/runtime/DeviceAPI.cuh#L39)
> 
> It doesn't look like there's anything to handle this gracefully, or is the plan to leave users at the whim of C++ errors (I'm not sure how bad they would be, probably just method x does not exist).

The code generator will not distinguish between the Device API or Read only API. It will just check against a list of API functions so will raise compilation errors if a user tries to set a variable. It is not a "proper" static analyser as it is a tree walker that doesn't retain state so there are other examples where people may call functions out of context that would raise compilation errors. E.g. You can call wrapped message functions on not wrapped implementations. 

The overall ambition of the code generator is to translate correct code (and filter out anything obviously wrong) but it is not expected to catch all errors that a compiler would. This would be done via Jitify 2 when we could link compile errors back to python code lines.

ptheywood (approved):

PR #911, opened 2022-08-30
ImGui support for bespoke interfaces

Robadob:
**Will require updating the CMake vis hash after Vis PR is merged.**

Most development detail can be found in the Vis PR.

I'm happy that this is ready.

Several potential features I've not bothered to implement:
* Allowing users to change the colour scheme (currently it's fixed to ImGui::Dark).
* Allowing users to customise the format string for environment property printing (`%g` is used for floating-point, but this can create unwanted high precision numbers in rare cases)
* Allowing users to manually position panels within the window (they can move them after sim has started)
* Allowing users to mark panels begin collapsed.

Vis PR: https://github.com/FLAMEGPU/FLAMEGPU2-visualiser/pull/100
Docs Pr: https://github.com/FLAMEGPU/FLAMEGPU2-docs/pull/110
Ped Nav Example Pr: https://github.com/FLAMEGPU/FLAMEGPU2-pedestrian_navigation-example/pull/1
Ped Nav Example Video: https://youtu.be/9jPfBs81XmE

### DIFF
--- a/include/flamegpu/visualiser/ModelVis.h
+++ b/include/flamegpu/visualiser/ModelVis.h
@@ -12,6 +12,7 @@
 #include "flamegpu/visualiser/AgentVis.h"
 #include "flamegpu/visualiser/StaticModelVis.h"
 #include "flamegpu/visualiser/LineVis.h"
+#include "flamegpu/visualiser/PanelVis.h"
 #include "flamegpu/visualiser/color/AutoPalette.h"
 #include "flamegpu/visualiser/config/ModelConfig.h"
 
@@ -178,6 +179,13 @@ class ModelVis {
      */
     LineVis newPolylineSketch(float r, float g, float b, float a = 1.0f);
     /**
+     * Add a customisable user interface panel to the visualisation
+     *
+     * Each panel can be moved around the screen/minimised with custom elements representing environment properties added
+     * @param panel_title The string that will be visible in the title of the panel
+     */
+    PanelVis newUIPanel(const std::string& panel_title);
+    /**
      * Sets the visualisation running in a background thread
      */
     void activate() {
@@ -202,10 +210,18 @@ class ModelVis {
      */
     bool isRunning() const;
     /**
+     * Random seed has changed
+     */
+    void updateRandomSeed();
+    /**
      * Updates all agent renders from corresponding
      * @param sc Step count, the step count value shown in visualisation HUD
      */
     void updateBuffers(const unsigned int &sc = UINT_MAX);
+    /**
+     * Singletons have init, so env props are ready to grab
+     */
+    void registerEnvProperties();
 
  private:
     /**
@@ -238,6 +254,10 @@ class ModelVis {
      * Pointer to the visualisation
      */
     std::unique_ptr<FLAMEGPU_Visualisation> visualiser;
+    /**
+     * Only need to register env properties once
+     */
+    bool env_registered = false;
 };
 
 }  // namespace visualiser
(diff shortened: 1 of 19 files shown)


include/flamegpu/visualiser/PanelVis.h
```diff
@@ -0,0 +1,291 @@
...
+#include "flamegpu/visualiser/config/ModelConfig.h"
+
+namespace flamegpu {
+namespace visualiser {
+
+/**
+ * This class serves as an interface for managing an instance of PanelConfig
+ * It allows elements to be specified for a UI panel for the visualisation
+ */
+class PanelVis {
```
ptheywood, line 19: Not super keen on the class name (don't like `Vis`, but it matches exisitng use so not going to fight it). 

`VisPanel` would make more snse though? otherwise given the existance of `ModelVis` it implies its currently visualising a panel? Feel completely free to ignor this.

include/flamegpu/visualiser/ModelVis.h
```diff
@@ -160,7 +161,7 @@ class ModelVis {
      * @param modelPath Path of the model on disk
      * @param texturePath Optional path to a texture fore the model on disk
      */
-    StaticModelVis addStaticModel(const std::string &modelPath, const std::string &texturePath = "");
+    StaticModelVis newStaticModel(const std::string &modelPath, const std::string &texturePath = "");
```
ptheywood, line 164: This needs to be listed as a braking change in the changelog for the next release. Probably better to do it now rather than hope it gets picked up when manually making the next changelog (which will be enourmous).

Also needs updating in the docs PR, but I've changed my review on that to reflect this already.

The change itself is a good one for consistency though, and now is the time to make breaking changes.

src/flamegpu/visualiser/ModelVis.cpp
```diff
@@ -126,7 +142,12 @@ void ModelVis::updateBuffers(const unsigned int &sc) {
         visualiser->releaseMutex();
     }
 }
-
+void ModelVis::updateRandomSeed() {
+    if (visualiser) {
+        // Yolo thread safety, shouldn't matter if random seed is printed wrong for a single frame
```
ptheywood, line 147: :vomiting_face: 

(I agree with the sentiment in the interest of time, but even still we should atleast make an issue so we can point at it if needed).

ptheywood (approved):
> Allowing users to change the colour scheme (currently it's fixed to ImGui::Dark).
> Allowing users to customise the format string for environment property printing (%g is used for floating-point, but this can create unwanted high precision numbers in rare cases)
> Allowing users to manually position panels within the window (they can move them after sim has started)
> Allowing users to mark panels begin collapsed.

Some / all of these should be promoted to issues, as they do all sound useful but probably shouldn't block this.

---

The API changes etc all look fine, and feel free to ignore my nitpicking.

There is an absence of tests though for the new classes (though vis code is probably missing tests in general, possibly for a good reason). 

We should atleast open an issue about testing the vis (if I haven't already made one in the past)

---


As you've noted, the hash needs updating prior to merge too (so the approval is subject to that)

Robadob:
> Some / all of these should be promoted to issues, as they do all sound useful but probably shouldn't block this.

https://github.com/FLAMEGPU/FLAMEGPU2-visualiser/issues/105

include/flamegpu/visualiser/ModelVis.h
Robadob: image
Already was marked as breaking, will update docs.

Robadob:
This should be ready for merge once CI passes, along with the docs PR. I have updated the linked visualiser to match the merged VIS PR.

After which, the ped example will want updating to point to latest FGPU2 hash.

PR #914, opened 2022-08-31
Correctly set visual studio filters for the split pyflamegpu targets

ptheywood:


Robadob (approved):
image
Tested it with a quick `ZERO_CHECK`

Issue #915, opened 2022-08-31
Native Python (codegen) errors with python 3.7


ptheywood:
Using python 3.7, under linux or windows the new `codegen` tests fail:

```
../tests/swig/python/codegen/test_codegen.py ......F........F.....FFFF..F...............FF.FF.
../tests/swig/python/codegen/test_codegen_integration.py .F
```

Current officially supported python versions are 3.7+, which we are aiming to also support. 3.7 goes EOL 2023-06-27, so ideally we should still support it. 

This will be blocking a new alpha release until either it is fixed (or disabled for python 3.7).   

Python 3.8 and 3.9 have been tested under linux and do not reproduce the errors.

---

[Full log (gist)](https://gist.github.com/ptheywood/8510fbf2dc217246ee80fe4ed15ef76e) log 

An example error:

```
../tests/swig/python/codegen/test_codegen.py:587: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <test_codegen.CodeGenTest testMethod=test_bytes>, source = "b'123'", exception_str = 'Byte strings not supported'

    def _checkException(self, source, exception_str):
        with pytest.raises(pyflamegpu.codegen.CodeGenException) as e:
            tree = ast.parse(source.strip())
            # code generate
            code = pyflamegpu.codegen.codegen(tree)
        if EXCEPTION_MSG_CHECKING:
>           assert exception_str in str(e.value)
E           AssertionError: assert 'Byte strings not supported' in 'Error (1, 0): Bytes function not supported'
E            +  where 'Error (1, 0): Bytes function not supported' = str(CodeGenException('Error (1, 0): Bytes function not supported'))
E            +    where CodeGenException('Error (1, 0): Bytes function not supported') = <ExceptionInfo CodeGenException('Error (1, 0): Bytes function not supported') tblen=10>.value

../tests/swig/python/codegen/test_codegen.py:508: AssertionError

```

ptheywood:
Closed by #917

PR #916, opened 2022-08-31
Swig array element unify

Robadob:
Have tested C++/Py tests.
Added new Py tests for the missing `CUDASimulation` functionality.

Only test failures were in CodeGen, due to #915


ptheywood (approved):
Changes look fine, though it's unclear why PanelVis.cpp is marked as new file in the diff for this PR given it's merging into `imgui`.

Good test additions too, though neither built nor ran them myself.

PR #930, opened 2022-10-13
Windows: CUDAEnsemble can block standby whilst running.

Robadob:
This uses Windows API to notify that threads running models inside `CUDAEnsemble` don't want the PC to enter standby.

It can be disabled by setting `block_standby=false` or passing `--standby` or `-s` at runtime. Though I'm not sure why a user would want to do that.

I don't believe there is a similar system wide power mgmt feature on Linux.

------------

This requires a manual test when I have a computer spare to try it on. It may be better to add it to the parent thread temporarily, rather than to the children threads.

### DIFF
--- a/src/flamegpu/gpu/CUDAEnsemble.cu
+++ b/src/flamegpu/gpu/CUDAEnsemble.cu
@@ -30,12 +30,24 @@ CUDAEnsemble::CUDAEnsemble(const ModelDescription& _model, int argc, const char*
     initialise(argc, argv);
 }
 CUDAEnsemble::~CUDAEnsemble() {
-    // Nothing to do
+// Call this here incase simulate() exited with an exception
+#ifdef _MSC_VER
+    if (config.block_standby) {
+        // Disable prevention of standby
+        SetThreadExecutionState(ES_CONTINUOUS);
+    }
+#endif
 }
 
 
 
 unsigned int CUDAEnsemble::simulate(const RunPlanVector &plans) {
+#ifdef _MSC_VER
+    if (config.block_standby) {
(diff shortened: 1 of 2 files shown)


src/flamegpu/gpu/CUDAEnsemble.cu
```diff
@@ -316,6 +317,13 @@ int CUDAEnsemble::checkArgs(int argc, const char** argv) {
             }
             continue;
         }
+        // -s/--standby Disable the blocking of standby
+        if (arg.compare("--standby") == 0 || arg.compare("-s") == 0) {
```
ptheywood, line 321: I'm not keen on `-s`, as `-s` means steps on the simulation cli. We don't need short options for everything. 

Also seem to have missed adding it to `CUDAEnsemble::printHelp`.

Robadob: Fair, will address both those.

ptheywood (approved):
Looks fine, though do test it actually prevents sleep prior to merging.

Robadob:
> Looks fine, though do test it actually prevents sleep prior to merging.

Yeah will do when I don't need my PC. Tomorrow when I goto git class or weekend when I'm at the gym maybe.

mondus:
More importantly check that the machine recovers from standby afterwards.

Robadob:
I changed around how this was implemented (it now blocks during `CUDAEnsemble::simulate()` rather than the bugged call to mark every runner thread as blocking).

Have tested it works as intended. PC can still go into standby after the call to `simulate()` returns, even if the process is running.

PR #935, opened 2022-10-19
Do not install required pip packages on behalf of users

### LINKED ISSUE #639: Do not install pip packages into the current environment
During CMake configuration with `-DBUILD_SWIG_PYTHON` several pip packages are automatically installed into the current python environment if they are not found.

This might just be the users default python environment.

If so, it is poor form to do this without any kind of prompt to the user. 

I believe this should have been covered by a previous issue, but I couldn't find it. I previously had done some work on performing everything in a venv within the build directory as part of #285, but ran into issues with Windows so stopped working on that if i remember correctly.

Related to #267 and probably others.

### LINKED ISSUE #947: setup.py install is deprecated
Spotted the following deprecation warning during pyflamegpu building:

```
SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.
```

This seems to have been introdued in [setuptools 58.3.0](https://setuptools.pypa.io/en/latest/history.html#v58-3-0) (22 Oct 2021)

> [#917](https://github.com/pypa/setuptools/issues/917): setup.py install and easy_install commands are now officially deprecated. Use other standards-based installers (like pip) and builders (like build). Workloads reliant on this behavior should pin to this major version of Setuptools. See [Why you shouldn’t invoke setup.py directly](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html) for more background.

I'm not sure what the setuptools removal timeline is, but they're on `v65.6.0` now.

---

We either need to fix this, or pin our setuptools as suggested?

---

We currently invoke `setup.py bdist_wheel` via our cmake, to produce the wheels, which impliclty calls install internally. 

Using [`build`](https://pypa-build.readthedocs.io/en/stable/) instead the equivalent call would (probably) be `python3 -m build --wheel`. The change would also require us to add the pyhton package `build` as a checked python dependency. There might also need to be some tweaks to the CMake for byproduces etc for correct cleaning.
`--wheel` is required as we don't care for source distributions `sdist` (default is sdist and bdist). 

`build` simply requires `Python >=3.6`, so should be fine to adopt, and is provided/maintained by pypa, so shoudl be supported/maintained for a decent chunk of time.

It'll be easiest to make this change post #935

ptheywood:
Emit a CMake Fatal Error during configuraition for missing required python packages, rather than installing them into the users environment.

Wheel, setuptools and optionall venv are requred.

Installing into --user automatically would fail in some cases, and installing not into user into a not-generated venv is bad form.

Closes #639

Also switches from the deprectaed `setup.py install` to use `build --wheel`. 
Closes #947.

### DIFF
--- a/.github/workflows/CMake.yml
+++ b/.github/workflows/CMake.yml
@@ -104,8 +104,7 @@ jobs:
       if: ${{ env.PYTHON != '' && env.BUILD_SWIG_PYTHON == 'ON' }}
       run: |
         sudo apt-get install python3-venv
-        python3 -m pip install --upgrade wheel
-        python3 -m pip install --upgrade setuptools
+        python3 -m pip install --upgrade wheel build setuptools
 
     - name: Install Visualisation Dependencies
       if: ${{ startswith(env.OS, 'ubuntu') && env.VISUALISATION == 'ON' }}
--- a/.github/workflows/Draft-Release.yml
+++ b/.github/workflows/Draft-Release.yml
@@ -103,8 +103,7 @@ jobs:
       if: ${{ env.PYTHON != '' && env.BUILD_SWIG_PYTHON == 'ON' }}
       run: |
         sudo apt-get install python3-venv
-        python3 -m pip install --upgrade wheel
-        python3 -m pip install --upgrade setuptools
+        python3 -m pip install --upgrade wheel build setuptools
 
     - name: Install Visualisation Dependencies
       if: ${{ startswith(env.OS, 'ubuntu') && env.VISUALISATION == 'ON' }}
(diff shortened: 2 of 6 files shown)


Robadob (approved):
Looks sensible

ptheywood:
> Would we benefit from a requirements.txt or similar?

We could, but things like `venv` are optional (and not always installable via pip). It would only contain `wheel` and `setuptools`, and I'd be reluctant to pin versions given the broad range of python version support (and subsequent dependabot spam). 

Individual examples might then have additional dependencies, including `pflamegpu` but that is not installable via name via pip anyway, so would just cause use of requirements.txt to fail if included.

I'm inclined to leave it for the future (though it might not hurt to open an issue?)

mondus (changes requested):
Can you add a brief sentence in README.md about where cmake will search for python and that cmake can be run from inside a virtual environment with the dependencies installed. Seems like this PR would be a good place for this to go in as its adding python discrepancies.

mondus (approved):

PR #937, opened 2022-10-19
Remove boids_bruteforce_dependency_graph

### LINKED ISSUE #874: Remove `boids_bruteforce_dependency_graph` example
The [`boids_bruteforce_dependency_graph`](https://github.com/FLAMEGPU/FLAMEGPU2/tree/master/examples/boids_bruteforce_dependency_graph) example is essentially a duplicate of `boids_bruteforce` but using the dependency graph execution ordering. 

This can be removed, with the other boids examples updated to use the (ideally/eventually) preffered dependency approach.

ptheywood:
Reduce the number of boids examples by replacing the original layer-based boids example with the dependency graph based example. 

Also updates the spatial3D boids exmaple to use the dependency graph.

Closes #874


Robadob (approved):
Looks reasonable, haven't tested it (assuming you have).

PR #940, opened 2022-10-21
Updated python boids model examples

mondus:
Renamed both examples, made th native Python version wrapped and added UI elements to visualisation.


Robadob (changes requested):
If you're adding a model, it should probably simply be included with the rest of the stock models. See [this PR](https://github.com/FLAMEGPU/FLAMEGPU2-visualiser/pull/91) for guidance.

P.S. I expect this may conflict with #937

examples/python_boids_spatial3D_wrapped_native/boids_spatial3D.py
```diff
@@ -1,13 +1,12 @@
-!# /usr/bin/env python3
```
ptheywood, line 1: Why have you removed the shebang?. This means you can't invoke the py file directly, `./boids_spatial3D.py`, but have to call it via `python3 boids_spatial3D.py`.

examples/python_boids_spatial3D_wrapped_native/boids_spatial3D.py
```diff
@@ -281,7 +279,7 @@ def inputdata(message_in: pyflamegpu.MessageSpatial3D, message_out: pyflamegpu.M
 
 env = model.Environment()
 # Population size to generate, if no agents are loaded from disk
-env.newPropertyUInt("POPULATION_TO_GENERATE", 4000)
```
ptheywood, line 284: we probably don't want to commit to 100k agents by default. Debug builds and smaller devices will suffer.

examples/swig_native_python_boids_spatial3D/venv.bat
```diff
@@ -1,12 +0,0 @@
-@echo off
```
ptheywood, line 1: I assume you didn't mean to actually just delete this file, rather you didnt force add the script in `python_boids_spatial3D_wrapped`, as `.bat` are gitignored.

ptheywood (changes requested):
The example itself looks fine, though I've not tested it myself.

Some minor comments, which will be easy to resolve. 

Do we want to think about splitting the examples dir into `CXX` and `python` subfolders or similar? `python_boids_spatial3D_bounded` is getting pretty long, and it might make things a bit clearer too? Sortof related to #603.

examples/python_boids_spatial3D_wrapped_native/boids_spatial3D.py
mondus: Because it stops the python file being executed in my windows Python environment using `python file.py`.

examples/swig_native_python_boids_spatial3D/venv.bat
mondus: Do we really want a platform specific batch file? Would instructions not be better in the file header?

examples/python_boids_spatial3D_wrapped_native/boids_spatial3D.py
mondus: Even a laptop GPU can handle 100k agents. I don't want all our demos to be small and unimpressive. I will however add a comment that this should be reduced for debug etc.

examples/python_boids_spatial3D_wrapped_native/model/Attribution.txt
```diff
@@ -0,0 +1,2 @@
+The 3D bird models are key frames extracted from the following 3D model provided under CC-BY
```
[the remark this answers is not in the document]
mondus: Fixed

mondus:
"Do we want to think about splitting the examples dir into CXX and python subfolders or similar?" @ptheywood 

I think treat this as a separate issue as examples could certainly do with a  tidy before release.

examples/python_boids_spatial3D_wrapped_native/boids_spatial3D.py
ptheywood: The lowest compute capability mobile GPU we support (SM35) is the 920M, a 33W, 384 core, 14.4Gb/s 733 GFLOPS device. It will not deal with a 100k agent sim, even if it only invovlves spatial comms (boids flocking leads to relatively dense comms). 

That is the most extreme case, but given this boids example is intended to be a better than real time visualisation, we should care about perf on a more recent but not great GPU, not everyone has a 3080.
Realistically Pascal is the current minimum arch to care about, my old 1070 for instance was a 15SM device, so 30k resident threads maximum, but achieved ~20k at the time (curve changes may have altererd this). 

I'm not saying it should be 4000 agents, but it probably shouldn't be 100,000.

examples/swig_native_python_boids_spatial3D/venv.bat
ptheywood: I don't see the need, but @Robadob wanted it (as mentioned on slack). You have added one for the other python example.

mondus:
> If you're adding a model, it should probably simply be included with the rest of the stock models (unless it's super high poly). See [this PR](https://github.com/FLAMEGPU/FLAMEGPU2-visualiser/pull/91) for guidance.
> 
> P.S. I expect this may conflict with #937

I think its fine to use an external model. Its good to have an example that does this. No issue with conflicts on #937 which is the C++ example not Python.

examples/python_boids_spatial3D_wrapped_native/boids_spatial3D.py
mondus: Fixed the typo and added back in both Python examples

examples/swig_native_python_boids_spatial3D/venv.bat
mondus: Batch file removed and readme added to each python example

examples/python_boids_spatial3D_bounded/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+or just
+
+`./boids_spatial3D.py`
+
+in Linux.
+
+# Visualisation
+
+If running with visualisation enabled then passing the argument `--steps 0` will ensure the simulation runs indefinitely. E.g.
```
Robadob, line 25: Alternatively, specifying 100 steps would allow you to explore the final simulation state after 100 steps have completed.

examples/python_boids_spatial3D_wrapped_native/model/Attribution.txt
```diff
@@ -0,0 +1 @@
+The 3D bird models (bird_a.obj and bird_b.obj) are key frames extracted from "Flying Bird" 3D model (https://sketchfab.com/3d-models/flying-bird-eb843194e06d429ebef7dd4aa7e265c1) by "sandeep.s" licenced under CC-BY 4.0 (https://creativecommons.org/licenses/by-nd/2.0/)
```
Robadob, line 1: > extracted from "Flying Bird"

Then you link to a "No Derivatives license". You can't modify it and redistribute.

P.S. You state CC BY 4.0, but link to CC BY-ND 2.0

Robadob: (This is why for the "stuntplane" stock model I just spent 30 minutes following a blender tutorial. So it would be free from any licenses.

Robadob: Appears if you click the license thing on sketchfab it takes you to `https://creativecommons.org/licenses/by/4.0/`, so it's not the ND version but you should fix the link.

```suggestion
The 3D bird models (bird_a.obj and bird_b.obj) are key frames extracted from "Flying Bird" 3D model (https://sketchfab.com/3d-models/flying-bird-eb843194e06d429ebef7dd4aa7e265c1) by "sandeep.s" licenced under CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)
```

examples/python_boids_spatial3D_bounded/README.md
mondus: added

examples/python_boids_spatial3D_wrapped_native/model/Attribution.txt
mondus: Not sure how I did this but have fixed this.

Robadob (approved):
Should be fine, I'm erring on the side of Pete regarding agent count though. The visualisations feel more useful with small populations (as you can actually make out the model detail whilst seeing the whole population).

P.S. A I said before I expect this may conflict with #937

examples/python_boids_spatial3D_wrapped_native/boids_spatial3D.py
```diff
@@ -353,17 +355,29 @@ def inputdata(message_in: pyflamegpu.MessageSpatial3D, message_out: pyflamegpu.M
...
-    circ_agt.setForwardXVariable("fx")
-    circ_agt.setForwardYVariable("fy")
-    circ_agt.setForwardZVariable("fz")
-    circ_agt.setModel(pyflamegpu.STUNTPLANE)
-    circ_agt.setModelScale(env.getPropertyFloat("SEPARATION_RADIUS") /3.0)
+    boid_agt.setForwardXVariable("fx")
+    boid_agt.setForwardYVariable("fy")
+    boid_agt.setForwardZVariable("fz")
+    #boid_agt.setModel(pyflamegpu.ARROWHEAD) # Alternative simple 3D model for very large pop sizes
+    boid_agt.setKeyFrameModel("./model/bird_a.obj", "./model/bird_b.obj", "wing_animation")
```
ptheywood, line 368: The python script can only be executed from the `examples/python_boids_spatial3D_wrapped_native/` directory, i.e. `model` must be a child of the current directory. 

E.g. executing it from a build directory doesn't work:

```
python3 ../examples/python_boids_spatial3D_wrapped_native/boids_spatial3D.py 
Traceback (most recent call last):
  File "/home/ptheywood/code/flamegpu/FLAMEGPU2/build-vis/../examples/python_boids_spatial3D_wrapped_native/boids_spatial3D.py", line 381, in <module>
    visualisation.activate()
  File "/home/ptheywood/code/flamegpu/FLAMEGPU2/build-vis/lib/Release/python/venv/lib/python3.10/site-packages/pyflamegpu/pyflamegpu.py", line 24109, in activate
    return _pyflamegpu.ModelVis_activate(self)
RuntimeError: /home/ptheywood/code/flamegpu/FLAMEGPU2/build-vis/_deps/flamegpu_visualiser-src/src/flamegpu/visualiser/util/Resources.cpp(82): Resources::locateFile(): File './model/bird_a.obj' could not be found!
```

With `pathlib`, it's not hard to get the abs path to the dir containing the script file

```
import pathlib
pathlib.Path(__file__).parent.resolve()
```

E.g. this works, but could probably be done cleaner.
```suggestion
    script_dir = pathlib.Path(__file__).parent.resolve()
    bird_a = str(script_dir / "model/bird_a.obj")
    bird_b = str(script_dir / "model/bird_b.obj")
    boid_agt.setKeyFrameModel(bird_a, bird_b, "wing_animation")
```

examples/python_boids_spatial3D_wrapped_native/boids_spatial3D.py
ptheywood: 40k runs at ~170 steps per second on my 3060ti. I've not got a laptop with a GPU to test an older / smaller device (had to enable the fps counter and change the background colour to see it). Perf does degrade as the step count progresses, as flocks form and local density increases.

100k slows to ~30sps by 3000 steps, so smaller/older devices definately wouldn't have maintained a useful rate.

examples/python_boids_spatial3D_wrapped_native/boids_spatial3D.py
ptheywood: As an aside, is `model` the best term to use (it's correct, but potentially ambiguous)?

examples/python_boids_spatial3D_wrapped_native/boids_spatial3D.py
mondus: I've changed to 40k. @Robadob I have tweaked parameters to that even at high numbers "you can actually make out the model detail whilst seeing the whole population".

ptheywood:
> P.S. A I said before I expect this may conflict with #937

I think it should be fine, I only modified c++ boids examples in that PR iirc.

ptheywood (approved):

PR #943, opened 2022-10-24
Rename of getThreadIndex to getIndex

### LINKED ISSUE #898: Add Host/Device API agent `getIndex` method
Add a method to get an agents index within a population as part of the FLAME GPU API (for host and device code?)

This will be the index within the current population, so the value is ephemeral and may change for a given agent between functions (i.e. if the order changes due to death / birth / state changes / generic sorting). 

For a CUDA backend, this is effectively the global thread id, but should be possible for non-CUDA implementations too (and the value will be known for agent data to be accessible anyway). 

Needs to clearly be distinct from ID, and needs to be well documented as only valid at the time the method was called. 

---

This will then be used in conflict resolution within the prospective schelling implementation

mondus:
Fixes #898 .

Simple fix by renaming. Not used in any tests only internally in a few places. Testing of this seems like overkill.

Addition of `getIndex` to host API can be dealt with as a separate enhancement issue as it is already achievable using iterator logic.


ptheywood:
This does not close the referenced issue, which is explicilty for a `getIndex` method which is for both the hsot and device components, although it is part of that issue (and the most immediately required one.

Could you remove the `Fixes` statement so it doens't auto close the issue? We can move it in the prioritisaiton board when this is done.

mondus:
> This does not close the referenced issue, which is explicilty for a `getIndex` method which is for both the hsot and device components, although it is part of that issue (and the most immediately required one.
> 
> Could you edit the PR description and remove the `Fixes` statement so it doens't auto close the issue? We can move it in the prioritisaiton board when this is done.

@ptheywood I've just added the host version of this instead along with a test for host side which is less trivial than the device API.

PR #944, opened 2022-10-25
Add Python 3.11 to CI including wheel generation

### LINKED ISSUE #886: Wheel Releases: Remove 3.6, add 3.11
Python 3.6 is no longer supported IIRC, so we probably want to remove this from the `Draft-Release` action workflow.

3.11 will also be releasing ~24th October 2022, at which point it will want adding to the matrix (and testing). We could potentially do an early test to check for compatibility via the release candidate

+ [x] Remove Python 3.6 from the release wheels
+ [x] Add Python 3.11 to the release wheels, after 2022-10-24
+ [x] Add Python 3.11 to the regular CI jobs where appropraite, after 2022-10-24.

ptheywood:
Python 3.11 is now fully released, so adds to Manylinux + Windows CI, and the wheel generation matrix in the draft-release CI. 

Closes #886

### DIFF
--- a/.github/workflows/Draft-Release.yml
+++ b/.github/workflows/Draft-Release.yml
@@ -293,6 +293,7 @@ jobs:
             hostcxx: devtoolset-8
             os: ubuntu-20.04
         python: 
+          - "3.11"
           - "3.10"
           - "3.9"
           - "3.8"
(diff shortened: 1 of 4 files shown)


Robadob (approved):
Not too clear why Ubuntu jumped from 3.8 to 3.11, whilst the rest are 3.10-3.11 jump.

But I assume you'll read this and change it if that's a problem.

ptheywood:
I'd missed changing it at some point in the past, so i made it match the regular windows one of using latest. 

3.8 was neither oldest supported nor newest, which are usually the most interesting in terms of CI.

PR #945, opened 2022-10-25
Update pinned versions of external GitHub Actions

ptheywood:
Update pinned versions of external GitHub Actions to resolve deprecation warnings. 

Node 11 actions are deprecated and will be removed in the future. The v2 versions of actions/checkout etc were node 11, so generate warnings, so updating to newer versions resolves the issue.

This has already been applied to the docs repo (open PR)

### DIFF
--- a/.github/workflows/CMake.yml
+++ b/.github/workflows/CMake.yml
@@ -67,7 +67,7 @@ jobs:
       CMAKE: ${{ matrix.cmake }}
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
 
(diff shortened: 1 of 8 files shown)


ptheywood:
Warnigns hwo sup on the summary of a given action as annotaitons, eg. for a [recent Windows action](https://github.com/FLAMEGPU/FLAMEGPU2/actions/runs/3320353870):

image

The summary for this PR should not show these annotations (if the PR has worked as intended)

Robadob (approved):
Trivial changes and CI still passes, so should be fine.

PR #946, opened 2022-10-25
Removal of Pointless References in Function Args

### LINKED ISSUE #825: Remove pointless references in functions
I was discussing a C++ quirk with a friend, they started berating me over `const bool &` as an argument from a screenshot of the affected code.

> That'll just slow your program down if the function isn't inlined by:
>     - Forcing the argument to be written to memory rather than passed in a register.
>     - Forcing a load and wasting a register on the address.

Should be pretty trivial to replace all constant references of primitive types (e.g. not `std::string`), with some find replace. In general most these uses won't be in perf critical places, but can't hurt.

mondus:
Fixes #825 

Data types fixed

- [x] unsigned int
- [x] int
- [x] bool
- [x] char
- [x] unsigned char
- [x] size_type (unsigned int)
- [x] float
- [x] double
- [x] signed char (we have some use of this)
- [x] EnvironmentManager::size_type
- [x] uint64_t
- [x] int64_t
- [x] uint16_t
- [x] int16_t
- [x] T
- [x] T2 (from use in macro properties)
- [x] cudaStream_t
- [x] size_t

Each items has been searched for using `const type &` as well as `const type&` as linting allows either.

Where const references appear in declarations they have been made primitive (pass by value) type argument and `const` has been removed as it is meaningless. Within definitions `const` primitive types are allowed as this signals an intention that the argument should not be changed in the function body.

### DIFF
--- a/include/flamegpu/gpu/CUDAScatter.cuh
+++ b/include/flamegpu/gpu/CUDAScatter.cuh
@@ -129,7 +129,7 @@ class CUDAScatter {
         const std::vector<ScatterData> &scatterData,
         unsigned int itemCount,
         unsigned int out_index_offset = 0,
-        const bool invert_scan_flag = false,
+        bool invert_scan_flag = false,
         unsigned int scatter_all_count = 0);
     /**
      * Scatters agents from SoA to SoA according to d_position flag as input_source, all variables are scattered
--- a/include/flamegpu/gpu/CUDAScatter.cuh
+++ b/include/flamegpu/gpu/CUDAScatter.cuh
@@ -106,7 +106,7 @@ class CUDAScatter {
         const std::map<std::string, void*> &out,
         unsigned int itemCount,
         unsigned int out_index_offset = 0,
-        const bool invert_scan_flag = false,
+        bool invert_scan_flag = false,
         unsigned int scatter_all_count = 0);
     /**
      * Scatters agents from SoA to SoA according to d_position flag
--- a/include/flamegpu/gpu/CUDAScatter.cuh
+++ b/include/flamegpu/gpu/CUDAScatter.cuh
@@ -99,7 +99,7 @@ class CUDAScatter {
      */
      unsigned int scatter(
         unsigned int streamResourceId,
-        const cudaStream_t &stream,
+        cudaStream_t stream,
         const Type &messageOrAgent,
         const VariableMap &vars,
         const std::map<std::string, void*> &in,
@@ -124,7 +124,7 @@ class CUDAScatter {
      */
     unsigned int scatter(
         unsigned int streamResourceId,
-        const cudaStream_t &stream,
+        cudaStream_t stream,
         const Type &messageOrAgent,
         const std::vector<ScatterData> &scatterData,
         unsigned int itemCount,
@@ -164,7 +164,7 @@ class CUDAScatter {
      */
     unsigned int scatterCount(
         unsigned int streamResourceId,
-        const cudaStream_t &stream,
+        cudaStream_t stream,
         const Type &messageOrAgent,
         unsigned int itemCount,
         unsigned int scatter_all_count = 0);
@@ -180,7 +180,7 @@ class CUDAScatter {
      */
     unsigned int scatterAll(
         unsigned int streamResourceId,
-        const cudaStream_t &stream,
+        cudaStream_t stream,
         const std::vector<ScatterData> &scatterData,
         unsigned int itemCount,
         unsigned int out_index_offset = 0);
@@ -196,7 +196,7 @@ class CUDAScatter {
      */
     unsigned int scatterAll(
         unsigned int streamResourceId,
-        const cudaStream_t &stream,
+        cudaStream_t stream,
         const VariableMap &vars,
         const std::map<std::string, void*> &in,
         const std::map<std::string, void*> &out,
@@ -218,7 +218,7 @@ class CUDAScatter {
      */
     void pbm_reorder(
         unsigned int streamResourceId,
-        const cudaStream_t &stream,
+        cudaStream_t stream,
         const VariableMap &vars,
         const std::map<std::string, void*> &in,
         const std::map<std::string, void*> &out,
@@ -238,9 +238,9 @@ class CUDAScatter {
      */
     void scatterNewAgents(
         unsigned int streamResourceId,
-        const cudaStream_t &stream,
+        cudaStream_t stream,
         const std::vector<ScatterData> &scatterData,
-        const size_t &totalAgentSize,
+        size_t totalAgentSize,
         unsigned int inCount,
         unsigned int out_index_offset);
     /**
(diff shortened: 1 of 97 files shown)


mondus:
There are actually quite a few places inside curve and message iterator functions where this might save a few registers.

Robadob:
> There are actually quite a few places inside curve and message iterator functions where this might save a few registers.

Much of the device API stuff is marked inline/forceinline. So if it is actually being inlined, the function calls are presumably irrelevant.

Robadob:
Windows/Release/SEATBELTS=ON tests pass. I need to look through the code changes before I put a review in though.

include/flamegpu/gpu/CUDAScatter.cuh
```diff
@@ -98,16 +98,16 @@ class CUDAScatter {
...
         const VariableMap &vars,
         const std::map<std::string, void*> &in,
         const std::map<std::string, void*> &out,
-        const unsigned int &itemCount,
-        const unsigned int &out_index_offset = 0,
-        const bool &invert_scan_flag = false,
-        const unsigned int &scatter_all_count = 0);
+        unsigned int itemCount,
+        unsigned int out_index_offset = 0,
+        const bool invert_scan_flag = false,
```
Robadob, line 109: `const bool` here was missed?

include/flamegpu/gpu/CUDAScatter.cuh
```diff
@@ -123,14 +123,14 @@ class CUDAScatter {
...
         const cudaStream_t &stream,
         const Type &messageOrAgent,
         const std::vector<ScatterData> &scatterData,
-        const unsigned int &itemCount,
-        const unsigned int &out_index_offset = 0,
-        const bool &invert_scan_flag = false,
-        const unsigned int &scatter_all_count = 0);
+        unsigned int itemCount,
+        unsigned int out_index_offset = 0,
+        const bool invert_scan_flag = false,
```
Robadob, line 132:
```suggestion
        bool invert_scan_flag = false,
```

include/flamegpu/gpu/CUDAScatter.cuh
Robadob:
```suggestion
        bool invert_scan_flag = false,
```

include/flamegpu/runtime/utility/HostRandom.cuh
```diff
@@ -38,7 +38,7 @@ class HostRandom {
     * @note Available as float or double
     */
     template<typename T>
-    inline T logNormal(const T& mean, const T& stddev) const;
+    inline T logNormal(const T mean, const T stddev) const;
```
Robadob, line 41:
```suggestion
    inline T logNormal(T mean, T stddev) const;
```

include/flamegpu/sim/RunPlanVector.h
```diff
@@ -203,10 +203,10 @@ class RunPlanVector : private std::vector<RunPlan>  {
      * @throws exception::InvalidEnvPropertyType If a property with the name has a type different to T
      * @throws exception::OutOfBoundsException If index is greater than or equal to the length of the environment property array
      * @throws exception::OutOfBoundsException If this vector has a length less than 2
-     * @see setPropertyNormalRandom(const std::string &name, const T &mean, const T &stddev)
+     * @see setPropertyNormalRandom(const std::string &name, T mean, T stddev)
      */
     template<typename T>
-    void setPropertyNormalRandom(const std::string &name, const EnvironmentManager::size_type &index, const T &mean, const T &stddev);
+    void setPropertyNormalRandom(const std::string &name, const EnvironmentManager::size_type index, const T mean, const T stddev);
```
Robadob, line 209:
```suggestion
    void setPropertyNormalRandom(const std::string &name, const EnvironmentManager::size_type index, T mean, T stddev);
```

include/flamegpu/sim/RunPlanVector.h
```diff
@@ -219,7 +219,7 @@ class RunPlanVector : private std::vector<RunPlan>  {
      * @throws exception::OutOfBoundsException If this vector has a length less than 2
      */
     template<typename T>
-    void setPropertyLogNormalRandom(const std::string &name, const T &mean, const T &stddev);
+    void setPropertyLogNormalRandom(const std::string &name, const T mean, const T stddev);
```
Robadob, line 222:
```suggestion
    void setPropertyLogNormalRandom(const std::string &name, T mean, T stddev);
```

include/flamegpu/sim/RunPlanVector.h
```diff
@@ -233,10 +233,10 @@ class RunPlanVector : private std::vector<RunPlan>  {
      * @throws exception::InvalidEnvPropertyType If a property with the name has a type different to T
      * @throws exception::OutOfBoundsException If index is greater than or equal to the length of the environment property array
      * @throws exception::OutOfBoundsException If this vector has a length less than 2
-     * @see setPropertyNormalRandom(const std::string &name, const T &mean, const T &stddev)
+     * @see setPropertyNormalRandom(const std::string &name, T mean, T stddev)
      */
     template<typename T>
-    void setPropertyLogNormalRandom(const std::string &name, const EnvironmentManager::size_type &index, const T &mean, const T &stddev);
+    void setPropertyLogNormalRandom(const std::string &name, const EnvironmentManager::size_type index, const T mean, const T stddev);
```
Robadob, line 239:
```suggestion
    void setPropertyLogNormalRandom(const std::string &name, EnvironmentManager::size_type index, T mean, T stddev);
```

src/flamegpu/util/detail/compute_capability.cu
```diff
@@ -90,7 +90,7 @@ std::vector<int> compute_capability::getNVRTCSupportedComputeCapabilties() {
 
 int compute_capability::selectAppropraiteComputeCapability(const int target, const std::vector<int>& architectures) {
     int maxArch = 0;
-    for (const int &arch : architectures) {
+    for (const int arch : architectures) {
```
Robadob, line 93:
```suggestion
    for (const int &arch : architectures) {
```

Removing the reference here causes the for loop to make copies.

Robadob (changes requested):
Looked through all the changes, added a bunch of suggestions where I think you made mistakes. However, given the length of changes my eyes may have glazed over and missed some.

Doesn't appear you changed any non-const references, which is where I was worried issues may occur.


Additionally you should probably do:
`cudaStream_t` doesn't need to be a reference, it's merely a typedef for a ptr
`size_t` too (`CUDAScatter.cuh:243`)

ptheywood:
Tests all pased for a debug build under linux. I won't do a full review right now, but will take a look when Rob's suggestions have been resolved to see if any bits still jump out.

src/flamegpu/util/detail/compute_capability.cu
mondus: I think a copy in this case is also fine as it is a primitive type. Will commit the suggestion anyway.

mondus:
> Looked through all the changes, added a bunch of suggestions where I think you made mistakes. However, given the length of changes my eyes may have glazed over and missed some.
> 
> Doesn't appear you changed any non-const references, which is where I was worried issues may occur.
> 
> Additionally you should probably do: `cudaStream_t` doesn't need to be a reference, it's merely a typedef for a ptr `size_t` too (`CUDAScatter.cuh:243`)

Additional types added. Tests passing with SEATBELTS.

Robadob (approved):

PR #949, opened 2022-10-27
Remove ensemble warnings

### LINKED ISSUE #675: Ensemble warning by default
When running an ensemble with the default config, a warning is emitted by default:

```
Warning: Only 1 of out_directory and out_format is set, both must be set for logging to commence to file.
```

This is due to `out_directory` defaulting to `""`, and `out_format` defaulting to `json`. 

We could do one or more of the following:

+ Only emit the warning if `out_directory` is unset? 
+ Default `out_format` to `""`
+ Change `out_directory` and `out_format` in the config struct to be a pair or sub-struct (Doesn't help much?) 
    + Swig might require some extra effort
+ Replace the Config struct with a class, containing private members with getters/setters to only allow both to be set as a pair?. 
    + This will add a bunch of faff, and swig might not be too keen on this.

This warning is also output even if `-q` is passed. Do we  want to provide a way to suppress warnings too?  This would likely be much more involved however.

mondus:
Fixes #675 by removing the warning. This can be justified as a warning should not be output for default behaviour (`out_directory="", out_format=""`). Instead no logging will take place. 

If the `out_directory` is set then logging takes place. If `out_directory` is set and the user changes `out_format` to "" (which would previously raise a warning) then an exception will be thrown instead (as it would with any `out_format` which represents an unsupported format). 

The only place where this is less nice is if someone sets  `out_format` but doesn't set `out_directory` in which case there is no warning and no logging. This doesn't seem unreasonable.

### DIFF
--- a/src/flamegpu/gpu/CUDAEnsemble.cu
+++ b/src/flamegpu/gpu/CUDAEnsemble.cu
@@ -144,8 +144,8 @@ unsigned int CUDAEnsemble::simulate(const RunPlanVector &plans) {
     if (!config.out_directory.empty() && !config.out_format.empty()) {
         log_worker = new SimLogger(run_logs, plans, config.out_directory, config.out_format, log_export_queue, log_export_queue_mutex, log_export_queue_cdn,
         step_log_config.get(), exit_log_config.get(), step_log_config && step_log_config->log_timing, exit_log_config && exit_log_config->log_timing);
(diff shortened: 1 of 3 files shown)


Robadob (approved):
Seems fine, other option would have been to change default out dir to `"."`, aka current directory.

Issue #939, opened 2022-10-20
Python Fatal Error during pytest


ptheywood:
@Robadob found that the python test suite would exit with a `Fatal Python error: Aborted` using the current master branch a76da8ca75a543513f5ea6c108a52da6f0c70afc. 

I've reproduced this under linux, for a release builds with and without visualisation enabled.

As rob found, the reported error in `io/test_logging_exceptions.py` is fine in isolation. The `io` suite on its own is OK. 

Running the `gpu` and `io` suites together triggers the error too:

```
python3 -m pytest ../tests/swig/python/gpu/ ../tests/swig/python/io/

... 

../tests/swig/python/io/test_logging_exceptions.py ..Fatal Python error: Aborted

```

I've managed to trigger a `CUDAError` by narrowing the test down further, to `test_cuda_simulation.py` & `test_logging.py`. 

```
python3 -m pytest ../tests/swig/python/gpu/test_cuda_simulation.py ../tests/swig/python/io/test_logging.py 
===================================== test session starts =====================================
platform linux -- Python 3.10.6, pytest-7.1.3, pluggy-1.0.0
rootdir: /home/ptheywood/code/flamegpu/FLAMEGPU2
collected 23 items                                                                            

../tests/swig/python/gpu/test_cuda_simulation.py ......s.............                   [ 86%]
../tests/swig/python/io/test_logging.py ...                                             [100%]

================================ 22 passed, 1 skipped in 0.66s ================================
terminate called after throwing an instance of 'flamegpu::exception::CUDAError'
  what():  /home/ptheywood/code/flamegpu/FLAMEGPU2/include/flamegpu/gpu/detail/CUDAErrorChecking.cuh(28): CUDA Error: /home/ptheywood/code/flamegpu/FLAMEGPU2/src/flamegpu/runtime/utility/EnvironmentManager.cu(149): invalid argument
Aborted (core dumped)
```

Runnint `test_cuda_simulation` and `io` it reproduces the original error message.

```
python3 -m pytest ../tests/swig/python/gpu/test_cuda_simulation.py ../tests/swig/python/io/
===================================== test session starts =====================================
platform linux -- Python 3.10.6, pytest-7.1.3, pluggy-1.0.0
rootdir: /home/ptheywood/code/flamegpu/FLAMEGPU2
collected 30 items                                                                            

../tests/swig/python/gpu/test_cuda_simulation.py ......s.............                   [ 66%]
[...]

PR #950, opened 2022-10-28
Remove automatic `cudaDeviceReset`

ptheywood:
The automatic `cudaDeviceReset` in `CUDASimulation` and `CUDAEnsemble` were breaking the pyhton test suite due to GC of a `CUDASimulation` being triggered after the reset in `CUDAEnsemble::simulate`. 

This PR removes the device resets for now, which will (potentially) break use of cuda-memecheck, profilers etc, but allow the python test suite to run.

The planned more significant fix is to add an implementation agnostic public api method (`cleanup`/`finalise`/`finalize`/`shutdown`) which for the current (only) implemetnation using CUDA would perform a `cudaDeviceReset` for any/all CUDA devices (potentially used, maybe not initially). However, this is non-trivial to implement in practice in a way that can be used in the same scope as a `CUDASimulation` or `CUDAEnsemble` (or any other class which uses cuda in the dtor). 

This PR is to make the python test suite usable again, while another PR (see the `device-reset-method` branch) will add the new method, but that may take some time to correctly implement.

Part of #939.

### DIFF
--- a/include/flamegpu/gpu/CUDASimulation.h
+++ b/include/flamegpu/gpu/CUDASimulation.h
@@ -604,11 +604,6 @@ class CUDASimulation : public Simulation {
     std::unique_ptr<visualiser::ModelVis> visualisation;
 #endif
     /**
-     * This tracks the current number of alive CUDASimulation instances
-     * When the last is destructed, cudaDeviceReset is triggered();
-     */
-    static std::atomic<int> active_instances;
-    /**
      * Returns false if any agent functions or agent function conditions are not RTC
      * Used by constructor to set isPureRTC constant value
      * @param _model The agent model hierarchy to check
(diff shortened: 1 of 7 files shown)


mondus (approved):
Tests seem sensible.

mondus:
@ptheywood Can you run the full test (including Python) suite as I have just rebased this?

ptheywood:
I had previously ran the full  test suite of both, it's how I knew it defiantely worked but will re do it anyway. The only changes since then are the ensemble warnings if I've followed the parent chains correctly.

ptheywood:
Release, seatbelts=ON, CUDA_ARCH=86, CUDA 11.8, GCC 11, python 3.10, VISUALISATION=OFF:

C++: 
```
[==========] 1023 tests from 80 test suites ran. (163873 ms total)
[  PASSED  ] 1023 tests.

  YOU HAVE 40 DISABLED TESTS
```

Python: 
```
========================== 610 passed, 5 skipped in 294.76s (0:04:54) ==========================
Warning: Input file 'test.json' refers to second input file 'invalid', this will not be loaded.
RTC Initialisation Processing time: 0.000000 s
Warning: Input file 'test.json' refers to second input file 'invalid', this will not be loaded.
Step 1 Processing time: 0.000096 s
Warning: Input file 'test.xml' refers to second input file 'invalid', this will not be loaded.
RTC Initialisation Processing time: 0.000000 s
Warning: Input file 'test.xml' refers to second input file 'invalid', this will not be loaded.
Step 1 Processing time: 0.000109 s
```

We do need to remember to do vis builds when implementing the `cleanup` method in case it owns some device memory (need to check).

PR #960, opened 2022-11-03
Verbosity levels

### LINKED ISSUE #677: Simulation/Ensemble verbose/quiet configuration
Currently:
+ Simulations have `-v/--verbose/config().verbose` options, to enable opt-in verbose output (progress per step)
+ Ensembles have `-q/--quiet/config().verbose` options to disable default output (completion of each ensemble).

It would probably be a good idea to try to standardise these. 

One approach would be to add both options to both classes.

+ `quiet` to disable default output
+ `verbose` to enable extra output

With the current direct access to a configuration struct, it would be possible to set both to `true` at the same time, so one would need to override the other (verbose output checks should check for `verbose && !quiet` ?

There are several alternatives:

+ Enum for verbosity level? `DEFAULT`, `QUIET` or `VERBOSE`.
+ Change configuration structs to classes, with methods to control private properties? 
+ use integers to track a level of output, with multiple settings of --verbose flags to make it even more verbose (this would likely be excessive) 

#257 might be worth thinking about at the same time.

### LINKED ISSUE #892: Output RNG seed
When a seed is not provided via the cli (and the user has not manually set one), the seed used for a run is not known. 

This should be output to `stdout` and / or somewhere else so it is not lost in case something interesting occurs in the individual run. It might be wroth just outputting all the time (atleast if verbose)

Will need to think about how to best deal with users being able to programatically set the seed  afte cli parsing has been done.

Will also need to not spam stdout with rng seeds for ensembles (but the seed used to generate rng run plans might be needed still?)  

This could be guarded to only emit for verbose runs, but I'd probably lean towards it being emitted by default (but not for quiet runs).

mondus:
Fixes #677 by adding verbosity levels and consistency between Simulation and Ensembles.

Notes for doc update ate that

 - Timeing (e.g. `-t`) data is also displayed if using `-v`
 - `-q` suppresses any warnings
 - `-q` respected by ensembles
 - `-v` individual sims will not print verbose output and will use default levels unless quiet

All warnings are now to `stderr` as there was inconsisteny with this.

Also closes #892 outputting of sim seed. Does this under verbose output. Verbose output for ensembles outputs the number of concurrent runs.

From the Ensemble example (note: the final two lines are part of the example not simulation outputs):

-q (quiet verbosity)
```
Ensemble init: 450, calculated init 450
Ensemble result: 40244135200, calculated result 40244135200
```

no args (default vebosity)
```
CUDAEnsemble completed 100 runs successfully!
Ensemble init: 450, calculated init 450
Ensemble result: 40244135200, calculated result 40244135200
```

-v (verbose output)
```
FLAME GPU 2.0.0-alpha.3+c524a128 Ensemble configuration
        Concurrent runs 4
CUDAEnsemble completed 100 runs successfully!
Ensemble time elapsed: 0.327000s
Ensemble init: 450, calculated init 450
Ensemble result: 40244135200, calculated result 40244135200
```

Tests have been added in C++ to test expected outputs of Simulate and Ensemble runs. Various other tests and Python tests have been updated. Python doesn't have duplicate tests which explicitly check outputs to stderr and stdout.

### DIFF
--- a/include/flamegpu/defines.h
+++ b/include/flamegpu/defines.h
@@ -17,6 +17,10 @@ constexpr const char* ID_VARIABLE_NAME = "_id";
  * If this value is changed, things may break
  */
 constexpr id_t ID_NOT_SET = 0;
+/**
+* Typedef for verbosity level of the API
+*/
+typedef enum {QUIET, DEFAULT, VERBOSE} verbosity;
 
 typedef unsigned int size_type;
 
--- a/include/flamegpu/defines.h
+++ b/include/flamegpu/defines.h
@@ -20,7 +20,7 @@ constexpr id_t ID_NOT_SET = 0;
 /**
 * Typedef for verbosity level of the API
 */
-typedef enum {QUIET, DEFAULT, VERBOSE} verbosity;
+enum class Verbosity {Quiet, Default, Verbose};
 
 typedef unsigned int size_type;
 
--- a/src/flamegpu/gpu/CUDAEnsemble.cu
+++ b/src/flamegpu/gpu/CUDAEnsemble.cu
@@ -180,12 +180,12 @@ unsigned int CUDAEnsemble::simulate(const RunPlanVector &plans) {
     ensemble_elapsed_time = ensemble_timer.getElapsedSeconds();
 
     // Ensemble has finished, print summary
-    if (config.verbosity > 0) {
+    if (config.verbosity > Verbosity::Quiet) {
         printf("\rCUDAEnsemble completed %u runs successfully!\n", static_cast<unsigned int>(plans.size() - err_ct));
         if (err_ct)
             printf("There were a total of %u errors.\n", err_ct.load());
     }
-    if (config.timing || config.verbosity >= VERBOSE) {
+    if (config.timing || config.verbosity >= Verbosity::Verbose) {
         printf("Ensemble time elapsed: %fs\n", ensemble_elapsed_time);
     }
 
@@ -213,7 +213,7 @@ void CUDAEnsemble::initialise(int argc, const char** argv) {
         exit(EXIT_FAILURE);
     }
     // If verbose, output the flamegpu version and seed.
-    if (config.verbosity == VERBOSE) {
+    if (config.verbosity == Verbosity::Verbose) {
         fprintf(stdout, "FLAME GPU %s\n", flamegpu::VERSION_FULL);
         fprintf(stdout, "Ensemble configuration:\n");
         fprintf(stdout, "\tConcurrent runs: %u\n", config.concurrent_runs);
@@ -297,12 +297,12 @@ int CUDAEnsemble::checkArgs(int argc, const char** argv) {
         }
         // -q/--quiet, Don't report progress to console.
         if (arg.compare("--quiet") == 0 || arg.compare("-q") == 0) {
-            config.verbosity = QUIET;
+            config.verbosity = Verbosity::Quiet;
             continue;
         }
         // -v/--verbose, Report all progress to console.
         if (arg.compare("--verbose") == 0 || arg.compare("-v") == 0) {
-            config.verbosity = VERBOSE;
+            config.verbosity = Verbosity::Verbose;
             continue;
         }
         // -t/--timing, Output timing information to stdout
--- a/src/flamegpu/gpu/CUDAEnsemble.cu
+++ b/src/flamegpu/gpu/CUDAEnsemble.cu
@@ -214,7 +214,8 @@ void CUDAEnsemble::initialise(int argc, const char** argv) {
     }
     // If verbose, output the flamegpu version and seed.
     if (config.verbosity == VERBOSE) {
-        fprintf(stdout, "FLAME GPU %s Ensemble configuration\n", flamegpu::VERSION_FULL);
+        fprintf(stdout, "FLAME GPU %s\n", flamegpu::VERSION_FULL);
+        fprintf(stdout, "Ensemble configuration:\n");
         fprintf(stdout, "\tConcurrent runs: %u\n", config.concurrent_runs);
     }
 }
@@ -354,8 +355,8 @@ void CUDAEnsemble::printHelp(const char *executable) {
     printf(line_fmt, "-c, --concurrent <runs>", "Number of concurrent simulations to run per device");
     printf(line_fmt, "", "By default, 4 will be used.");
     printf(line_fmt, "-o, --out <directory> <filetype>", "Directory and filetype for ensemble outputs");
-    printf(line_fmt, "-q, --quiet", "Don't print information to console");
-    printf(line_fmt, "-v, --verbose", "Print addititional information to console");
+    printf(line_fmt, "-q, --quiet", "Do not print progress information to console");
+    printf(line_fmt, "-v, --verbose", "Print config, progress and timing (-t) information to console");
     printf(line_fmt, "-t, --timing", "Output timing information to stdout");
     printf(line_fmt, "-e, --error <error level>", "The error level 0, 1, 2, off, slow or fast");
 #ifdef _MSC_VER
--- a/src/flamegpu/gpu/CUDAEnsemble.cu
+++ b/src/flamegpu/gpu/CUDAEnsemble.cu
@@ -145,7 +145,7 @@ unsigned int CUDAEnsemble::simulate(const RunPlanVector &plans) {
                 new (&runners[i++]) SimRunner(model, err_ct, next_run, plans,
                     step_log_config, exit_log_config,
                     d, j,
-                    !config.quiet, config.error_level == EnsembleConfig::Fast,
+                    config.verbosity, config.error_level == EnsembleConfig::Fast,
                     run_logs, log_export_queue, log_export_queue_mutex, log_export_queue_cdn, fast_err_detail, TOTAL_RUNNERS);
             }
         }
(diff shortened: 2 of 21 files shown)


src/flamegpu/sim/SimRunner.cu
```diff
@@ -104,7 +106,7 @@ void SimRunner::start() {
             }
             log_export_queue_cdn.notify_one();
             // Print progress to console
-            if (verbose) {
+            if (verbosity == VERBOSE) {
```
Robadob, line 109: This should be "Default" rather than "Verbose" imo.

It' uses `\r` so it just rewrites the same line of output to show progress.
(However, it will write something ugly if output is being piped to file as we don't detect to provide an alt output).

src/flamegpu/sim/SimRunner.cu
```diff
@@ -125,11 +127,13 @@ void SimRunner::start() {
                 }
                 return;
             } else {
-                if (verbose) {
+                // Progress flush
+                if (verbosity == VERBOSE) {
```
Robadob, line 131: If above is changed, this will need updating too.

Robadob:
The changes look good so far, I suggested an adjustment to levels in one place you've done so far.

src/flamegpu/sim/SimRunner.cu
Robadob: I'm almost certain that previously always progress printed by default. It's very useful to know how far along your huge batch ensemble is with processing.

Robadob: Yes, `verbose` would be set `true` unless quiet was enabled.

https://github.com/FLAMEGPU/FLAMEGPU2/blob/0f5a4c597560a6991a9a9b33c730fb9fc18622ad/src/flamegpu/gpu/CUDAEnsemble.cu#L148

Robadob: The outputs should not be "racy", we fixed the issue where progress counter would briefly count backwards (it was printing the id of the last finished job, rather than finished count).

Robadob: I'm personally inclined to say Sim's should always be marked Quiet by ensembles. Their output is kinda useless if it's not associated with a particular run during concurrent ensembles.

Robadob: To achieve what you want, would be to add a programtic option to CUDASimulation which pipes stdout/stderr to file. So a user can review them in isolation after.

src/flamegpu/sim/SimRunner.cu
mondus: Agreed

mondus:
C++ and Python tests passing in Release

Robadob (approved):
Can't spot any problems, couple of thoughts though:

* I'd have probably named the `verbosity` enum `Verbosity` or `verbosity_t` to differentiate it from a variable name.
* `pyflamegpu.DEFAULT` feels a bit too general of a name to embed directly under `pyflamegpu` (but we don't have modules)
* Not clear why `TestRunPlan::noOutputWarning` was removed

ptheywood (changes requested):
In general it's good, just a couple of missed bits (see below) and a few suggestions that should address some of Rob's points too. 

* `CUDASimulation`'s --help does not include `-q/--quiet` (not in the diff so can't add a review comment in the right place easily)
* You've changed `test_cuda_simulation.cu` but not `test_cuda_simulation.py`, so `-q` musnt' be tested in an argparse test there (it does seem to work though, but for consistency it should be there given you added it for ensemble

src/flamegpu/gpu/CUDAEnsemble.cu
```diff
@@ -212,12 +212,11 @@ void CUDAEnsemble::initialise(int argc, const char** argv) {
     if (!checkArgs(argc, argv)) {
         exit(EXIT_FAILURE);
     }
-    /* Disabled as this is printed prior to quiet being accessible 
-    // If verbsoe, output the flamegpu version.
-    if (!config.quiet) {
-        fprintf(stdout, "FLAME GPU %s\n", flamegpu::VERSION_FULL);
+    // If verbose, output the flamegpu version and seed.
+    if (config.verbosity == VERBOSE) {
+        fprintf(stdout, "FLAME GPU %s Ensemble configuration\n", flamegpu::VERSION_FULL);
```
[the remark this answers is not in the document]
mondus: Fixed

include/flamegpu/defines.h
```diff
@@ -17,6 +17,10 @@ constexpr const char* ID_VARIABLE_NAME = "_id";
  * If this value is changed, things may break
  */
 constexpr id_t ID_NOT_SET = 0;
+/**
+* Typedef for verbosity level of the API
+*/
+typedef enum {QUIET, DEFAULT, VERBOSE} verbosity;
```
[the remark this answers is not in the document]
Robadob: From the docs
```
ensemble.Config().error_level = pyflamegpu.CUDAEnsembleConfig.Fast
```
`CUDAEnsembleConfig` is the name of the class which nests the enum, not the name of the enum's class.

src/flamegpu/gpu/CUDAEnsemble.cu
```diff
@@ -212,12 +212,11 @@ void CUDAEnsemble::initialise(int argc, const char** argv) {
...
+        fprintf(stdout, "\tConcurrent runs: %u\n", config.concurrent_runs);
```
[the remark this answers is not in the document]
mondus: Wont fix as its consistent with Simulation Configuration output. I have clarified the previous line as `Ensemble configuration:` so its clearer why this is indented. We may also add more verbose config outputs at a later date.

include/flamegpu/defines.h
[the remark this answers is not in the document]
Robadob: It's defined here, just `enum`, rather than `class enum`, not sure of the difference.
https://github.com/FLAMEGPU/FLAMEGPU2/blob/0f5a4c597560a6991a9a9b33c730fb9fc18622ad/include/flamegpu/gpu/CUDAEnsemble.h#L61

src/flamegpu/gpu/CUDAEnsemble.cu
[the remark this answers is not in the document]
Robadob: Could print device information too, it is verbose.

src/flamegpu/gpu/CUDAEnsemble.cu
```diff
@@ -350,7 +354,8 @@ void CUDAEnsemble::printHelp(const char *executable) {
     printf(line_fmt, "-c, --concurrent <runs>", "Number of concurrent simulations to run per device");
     printf(line_fmt, "", "By default, 4 will be used.");
     printf(line_fmt, "-o, --out <directory> <filetype>", "Directory and filetype for ensemble outputs");
-    printf(line_fmt, "-q, --quiet", "Don't print progress information to console");
+    printf(line_fmt, "-q, --quiet", "Don't print information to console");
```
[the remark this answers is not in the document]
mondus: I have changed this back as it is only progress info which would be output under `DEFAULT`.

include/flamegpu/defines.h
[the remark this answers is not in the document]
mondus: Have used a scoped enum `Verbosity`. Making this `Verbisty_t` will look ugly in SWIG so have encured that it has a fully qualified type name whereever one is declared. E.g. `flamegpu::Verbosity verbosity;`.

The enums have also been renamed to use Title Case this produces nicer SWIG bindings. E.g. `pyflamegpu.Verbosity_Quiet` (or `flamegpu::Verbosity::Quiet` in c++).

This also addresses 2 of robs comments.

mondus:
> * I'd have probably named the `verbosity` enum `Verbosity` or `verbosity_t` to differentiate it from a variable name.
I've used scoped enums and qualified the type wherever used with namespace and scope. Changing the name would make it ugly in SWIG.
> * `pyflamegpu.DEFAULT` feels a bit too general of a name to embed directly under `pyflamegpu` (but we don't have modules)
Resolved by using
Resolved using @ptheywood suggestion of scoped enums.
> * Not clear why `TestRunPlan::noOutputWarning` was removed
It was moved (and extended) into another test file as it was actually testing Ensemble functionality.

mondus:
> I meant enum Verbosity not the quiet/default/verbose version. I can never remember the correct way around for enum and enumerator.

Doesn't matter when using a scoped enum. Just pushed changes now.

ptheywood (approved):
Still haven't ammended the python test suite for `test_cuda_simulation.py` for argparse testing. 

Either add a test there, or open an issue so it gets added eventually.

mondus:
The following tests have been implemented.

`test_cuda_ensemble.cu`
 - [x] C++ Ensemble Tests for --quiet args
 - [x] C++ Ensemble Tests for default verbosity
 - [x] C++ Ensemble Tests for --verbose args
 - [x] C++ Ensemble Tests verbosity output levels
 
`test_cuda_simulate.cu`
 - [x] C++ Simulate Tests for --quiet args
 - [x] C++ Simulate Tests for default verbosity
 - [x] C++ Simulate Tests for --verbose args
 - [x] C++ Simulate Tests verbosity output levels
 
`test_cuda_ensemble.py`
 - [x] Python Ensemble Tests for --quiet args
 - [x] Python Ensemble Tests for default verbosity
 - [x] Python Ensemble Tests for --verbose args
 - [x] Python Ensemble Tests verbosity output levels (NOTE: skipped)
 
`test_cuda_simulate.py`
 - [x] Python Simulate Tests for --quiet args
 - [x] Python Simulate Tests for default verbosity
 - [x] Python Simulate Tests for --verbose args
 - [x] Python Simulate Tests verbosity output levels (NOTE: skipped)
 
Testing python output is problematic #966. @ptheywood Let me know if the skipped tests pass on Ubuntu.

ptheywood (approved):
Non skipped tests all pass under linux, just needs squashing when merging. 

Manually tested the skipped python stdout/err capturing tests and they were misbehaving on linux too, but resolving them will be a future problem.

PR #967, opened 2022-11-07
Warning for unknown args with tests

### LINKED ISSUE #931: CLI: Optionally allow unknown arguments
Current CLI parsing errors if unknown arguments are encountered.

This makes sense for pure use of the provided CLI, but complicates extending the CLI (but not replacing), as the user must build a copy of argv with the user-provided cli options removed.

The simplest solution to this, would be to add a mechanism for a model to skip/disable the unkown argument check during the `checkArgs` methods, so that they can just forward argv/argc on. 

Ideally this would be handeled during #257, but as that is a much larger job to think about it might be worth handling sooner to avoid an API break later on. (though this should be possible without a strict API break via overloading the relevant methods, or storing the setting in a class memebr to avoid having to modify ~4 method prototypes)

mondus:
Fixes #931.

Testing for previous behaviour is tricky as it causes a system exit. There are exit tests available in google test but these don't support "test for no exit". It would be possible to have a friend class test the arg parsing function specifically however an exit from google test would not constitute a pass (it just wouldn't fail as the testing would exit with an error code) so this seems unnecessary.

### DIFF
--- a/src/flamegpu/gpu/CUDAEnsemble.cu
+++ b/src/flamegpu/gpu/CUDAEnsemble.cu
@@ -338,9 +338,7 @@ int CUDAEnsemble::checkArgs(int argc, const char** argv) {
 #endif
             continue;
         }
-        fprintf(stderr, "Unexpected argument: %s\n", arg.c_str());
-        printHelp(argv[0]);
-        return false;
+        fprintf(stderr, "Warning: Unknown argument '%s' passed to Ensemble will be ignored\n", arg.c_str());
     }
     return true;
 }
--- a/src/flamegpu/gpu/CUDAEnsemble.cu
+++ b/src/flamegpu/gpu/CUDAEnsemble.cu
@@ -310,6 +310,11 @@ int CUDAEnsemble::checkArgs(int argc, const char** argv) {
             config.timing = true;
             continue;
         }
+        // -u/--silence-unknown-args, Silence warning for unknown arguments
+        if (arg.compare("--silence-unknown-args") == 0 || arg.compare("-u") == 0) {
+            config.silence_unknown_args = true;
+            continue;
+        }
         // -e/--error, Specify the error level
         if (arg.compare("--error") == 0 || arg.compare("-e") == 0) {
             if (i + 1 >= argc) {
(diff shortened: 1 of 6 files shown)


Robadob (approved):
Can't see anything wrong with it.

ptheywood (commented):
This needs a way to suppress the warning messages via a config property really to close the issue, as otherwise it just adds unsuppressable warning spam to any models wich just want to add to the default CLI (e.g. the use case which led to me opening the issue).

This doesn't need to be done prior to a release however, but either needs the PR changing to not automatically close the issue, or a new issue opening about suppressing the warning.

mondus:
Corresponding docs [PR](https://github.com/FLAMEGPU/FLAMEGPU2-docs/pull/119)

src/flamegpu/sim/Simulation.cu
```diff
@@ -288,9 +293,9 @@ int Simulation::checkArgs(int argc, const char** argv) {
         if (checkArgs_derived(argc, argv, i)) {
             continue;
         }
-        fprintf(stderr, "Unexpected argument: %s\n", arg.c_str());
-        printHelp(argv[0]);
-        return false;
+        // Warning if not in QUIET verbosity or if silnce-unknown-args is set
+        if (!(config.verbosity == flamegpu::Verbosity::Quiet || config.silence_unknown_args))
```
Robadob, line 297: For this to work, a user has to specify `silence_unknown_args`, programmatically or on command line, prior to the unknown argument. Otherwise, we'd need to double parse commandline.

Bit grim.

Robadob: Maybe worth adding a note about that quirk to the docs?

mondus: Its in the docs and the help. "Silence warnings for unknown arguments passed **after** this flag."

src/flamegpu/gpu/CUDAEnsemble.cu
```diff
@@ -310,6 +310,11 @@ int CUDAEnsemble::checkArgs(int argc, const char** argv) {
             config.timing = true;
             continue;
         }
+        // -u/--silence-unknown-args, Silence warning for unknown arguments
+        if (arg.compare("--silence-unknown-args") == 0 || arg.compare("-u") == 0) {
```
ptheywood, line 314: Does this need a short option? `-u` is not the most memorable for `silence  unknown arguments`, and there is a finite set of short options available (not that we wil ever have 52).

mondus: I don't see any harm.

PR #968, opened 2022-11-10
BugFix: Remove redundant code from CUDAFatAgent::addSubAgent()

### LINKED ISSUE #820: Memory on device 0 is always used.
CUDA Context is being created on device 0, before device is switched.

This example causes the below NVIDIA SMI output (note nvidia smi IDs don't match cuda device IDs).
```c++
flamegpu::CUDASimulation sim(model);
sim.CUDAConfig().device_id = 1
sim.applyConfig();
```

image

@ptheywood thinks he previously fixed this and it may have regressed, needs some investigation.

As of [620e904](https://github.com/FLAMEGPU/FLAMEGPU2/commit/620e9047300e0bf2b113655cfd038a240dd664e3) (after the shared mem curve merge) this is still an issue.

Robadob:
The only effect this code had was creation and deletion of CUDAFatAgentStateList objects, when mapping subagents. (std::map::emplace() return value includes a bool which denotes whether the insertion suceeded, in this case the two branches above ensure it will always fail.)

This had the secondary effect of potentially initialising a CUDA context on the wrong CUDA device.

Closes #820

Windows, Debug, Seatbelt=On
```

[==========] 1022 tests from 79 test suites ran. (369557 ms total)
[  PASSED  ] 1022 tests.

  YOU HAVE 53 DISABLED TESTS
```

*I haven't actually checked whether it solves the bug on a multi-device PC, however my understanding of @ptheywood's nsys check and my use of debugger suggest it does.*

### DIFF
--- a/src/flamegpu/gpu/CUDAFatAgent.cu
+++ b/src/flamegpu/gpu/CUDAFatAgent.cu
@@ -72,9 +72,6 @@ void CUDAFatAgent::addSubAgent(
             states.emplace(sub_state, cloned_state);
             states_unique.insert(cloned_state);
         }
-        // allocate memory for each state list by creating a new Agent State List
-        AgentState state = {mappedAgentCount, s};
-        states.emplace(state, std::make_shared<CUDAFatAgentStateList>(description));
     }
     // Handle agent variables
     for (auto &state : states_unique) {


ptheywood (approved):
This fixes the issue shown by running sugarscape with `-d 1` in a multi-gpu system (by watching nvidia-smi separately).

```
|=============================================================================|
|    0   N/A  N/A      1434      G   /usr/lib/xorg/Xorg                  4MiB |
|    1   N/A  N/A      1434      G   /usr/lib/xorg/Xorg                 12MiB |
|    1   N/A  N/A      1476      G   /usr/bin/sddm-greeter              18MiB |
|    1   N/A  N/A    203168      C   ./bin/Release/sugarscape          336MiB |
+-----------------------------------------------------------------------------+
```

Compared to the current state of master 

```

```

The existing test (`timeCUDASimulationContextCreationTest`) misses this as it always tests device 0, and doesn't create a submodel either. 
We could ammend it to use the -1th device, but even then it would need  a submodel for that to be caught, (and then everything else instanciating too in case there are other similar issues), and it would still only catch it in a multi-gpu system, so not worth ammending, especially as the test itself is a little fluffy (timing based, no api methods of detecting if a context has been established for a given device afiak)

PR #969, opened 2022-11-10
Handling NaN/Inf in JSON Output

### LINKED ISSUE #762: Handling NaN/Inf in JSON Output
Based on an issue Kenneth was having, it would appear that under FGPU2, if a float (or array) containing `NaN`, is exported that particular value (or array) is simply omitted. Leading to invalid JSON, and in the case of the array loss of data.

It could be better if we handle this, by replacing `NaN` with 0 and writing a warning to stderr. Same probably applies to `INF` too, though I've never achieved that by mistake.

(Or we work out how to enable RapidJSON to break standards and output NaN, but iirc there's a bunch of compiler compatibility issues trying to enable that).

mondus:
Closes #762 

Uses RapidJSONs write and parse flags to relax the JSON standard and output `nan` and `inf` type values. Not technically part of the JSON spec but this is the way in which RapidJSON deals with it. Alternative would be two write `nan` and `inf` as 0 but this would prevent fully serializing and loading a model. 

Requires changes to and test for;

- [x] State agent variables
- [x] State Environment Variables
- [ ] Logging: Wont implement (see below)
 
NOTE: Testing json log outputs is very involved and requires writing a new JSON parser so it wont be done as part of this fix. Currently non out of logging tests test the actual file contents. A new issue (#970 ) for loading logs from JSON in log objects has been opened as a future feature than would then permit this to be tested.

### DIFF
--- a/src/flamegpu/io/JSONLogger.cu
+++ b/src/flamegpu/io/JSONLogger.cu
@@ -277,12 +277,12 @@ void JSONLogger::logCommon(const RunLog &log, const RunPlan *plan, bool doLogCon
     rapidjson::StringBuffer s;
     if (prettyPrint) {
         // rapidjson::Writer doesn't have virtual methods, so can't pass rapidjson::PrettyWriter around as ptr to rapidjson::writer
-        rapidjson::PrettyWriter<rapidjson::StringBuffer>* writer = new rapidjson::PrettyWriter<rapidjson::StringBuffer>(s);
+        auto writer = new rapidjson::PrettyWriter<rapidjson::StringBuffer, rapidjson::UTF8<>, rapidjson::UTF8<>, rapidjson::CrtAllocator, rapidjson::kWriteNanAndInfFlag>(s);
         writer->SetIndent('\t', 1);
         logCommon(writer, log, plan, doLogConfig, doLogSteps, doLogExit, doLogStepTime, doLogExitTime);
         delete writer;
     } else {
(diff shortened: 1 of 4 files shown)


src/flamegpu/io/JSONLogger.cu
```diff
@@ -277,12 +277,12 @@ void JSONLogger::logCommon(const RunLog &log, const RunPlan *plan, bool doLogCon
     rapidjson::StringBuffer s;
     if (prettyPrint) {
         // rapidjson::Writer doesn't have virtual methods, so can't pass rapidjson::PrettyWriter around as ptr to rapidjson::writer
-        rapidjson::PrettyWriter<rapidjson::StringBuffer>* writer = new rapidjson::PrettyWriter<rapidjson::StringBuffer>(s);
+        auto writer = new rapidjson::PrettyWriter<rapidjson::StringBuffer, rapidjson::UTF8<>, rapidjson::UTF8<>, rapidjson::CrtAllocator, rapidjson::kWriteNanAndInfFlag>(s);
```
Robadob, line 280: Ah so this is the secret template args to enable NAN, I couldn't even figure out a combination that built the one or two times I looked into it.

Robadob (approved):
Changes look as I would expect, haven't tested it though.

src/flamegpu/io/JSONLogger.cu
mondus: Yeah the docs are not particularly clear on this.

ptheywood:
This should include a change to the docblocks / documentation statign that output JSON is RapdiJson's "relaxed" and can include nan etc.

PR #973, opened 2022-11-11
Prevent messages persisting between  iterations

ptheywood:
+ [x] Add test(s) which detect messages persisting from one iteration to the next
+ [ ] Prevent messages persisting from one iteration to the next (via resetting at the end of the iteration)


ptheywood:
Full test suite passes for release with belts under linux:

```
[==========] 1032 tests from 80 test suites ran. (165641 ms total)
[  PASSED  ] 1032 tests.

  YOU HAVE 40 DISABLED TESTS
```

If any models rely on persistence then this will break them however.

Robadob:
> If any models rely on persistence then this will break them however.

Yeah I don't agree with this change. Add a HostAPI method for users to manually purge message lists, and bin this PR.

ptheywood:
This is now ready for review, as a PR which makes non-persistent messages default, with opt-in to persistence for each message list.

Adds C++ and python tests for new features, but only checks persistence behaves correctly for brute force messages and bucket messages (in the interest of time, they are not simple tests).

PR #974, opened 2022-11-11
Add Device Reset Method

ptheywood:
+ [x] Adds a public API method for users to reset all cuda devices (but named agnostically
+ [x] Call it in the examples, and in the test suite so they can be profiled/memchecked
+ [x] Add test to show this works with a single cudaSimulation
+ [x] Add python test(s) to show this works
+ [x] Wrap `cudaFree` in dtors to ensure the pointer being freee'd is a valid device pointer
+ [ ] Safe cudaStream destruction in `CUDASimulation`
  + Checking th explicit return code works some of the time, but  some build configs is a segfault. Unsure of a better way to do this right now.
+ [ ] Wider testing of dtor's after cleanup, to ensure there's no unwrapped CUDA.


ptheywood:
This now is in a state where the cleanup method is callable from python and c++, but is only explciltly tested against CUDASimulation objects.

The workaround for context checking prior to stream destruction needs some polish, and could be wrapped into a re-usable method too. 
Ideally there could be extra checks all over our codebase to produce more useful exceptions than the implicit CUDA  errors, but that's a much more substatntial task than just fixing current dtor use.

ptheywood:
This is now ready for review / could do with both test suites running on windows.

I'm not convinced by the name anymore `cleanup` implies that all global state will be cleaned up, but all it does is reset devices? Not the end of the world. 

The new cuda util header at `flamegpu::util::detail::cuda` is maybe a bit deep, but that might just fold into the other open namespacing tidying issue that needs looking at / thinking about. Just `flamegpu::detail::cuda` might  be nicer? using `cuda` as the file name is borderline sketchy too though, given some debuggers use of filename only (i.e if it was `cuda.h` rather than `cuda.cuh`, which is not far off).

Robadob:
Windows/Release/Seatbelts=ON
```
[==========] 1041 tests from 82 test suites ran. (158219 ms total)
[  PASSED  ] 1041 tests.

  YOU HAVE 40 DISABLED TESTS
  ```

Robadob (approved):
The new methods look sensible. (`util/cleanup.h`, `util/detail/cuda.cuh`)

I haven't done an exhaustive check that you've replaced all prior usages though. I assume you used grep or similar.

PR #975, opened 2022-11-15
Automatically trigger __disown__() when Python hostfns are bound to models.

### LINKED ISSUE #498: SWIG Host Callbacks going out of scope
When setting up standalone primage, I define submodel inside a seperate file/function. This causes the python defined exit condition to go out of scope, which leads to an access violation when CUDASimulation is iteration exit callbacks.

This error was actually silent until I attached visual studio debugger. So the program would either exit early or get stuck calling the parent models step function instead of the exit condition forever.

We need a better way of preferably making a deep copy of the callback handlers. Or if that's not possible, atleast detecting when python GCs them.

---

Moving the definition of the ExitCallBack object from method scope to file scope appears to have fixed it. But this obviously isn't a fix, just a work around, as we can't detect when they are passed out of scope.

Robadob:
I wasn't able to trigger the exception, but added a try-catch regardless.

I think I have removed all usages of `__disown__()` from the Python tests, checked using `grep -rnw "tests/" -e "disown"`.

Closes #498

### DIFF
--- a/swig/python/flamegpu.i
+++ b/swig/python/flamegpu.i
@@ -472,6 +472,34 @@ class FLAMEGPURuntimeException : public std::exception {
 %feature("director") flamegpu::HostFunctionCallback;
 %feature("director") flamegpu::HostFunctionConditionCallback;
 
+
+// Automatically disown all passed host functions, to prevent them going out of scope too early
+// This still leaves a potential race condition if stateful information is stored in a host function instance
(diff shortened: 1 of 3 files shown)


ptheywood:
```
617 passed, 11 skipped in 310.61s
```

Linux python tests all pass under linux (Release Seatbelts=ON)

ptheywood (approved):

PR #952, opened 2022-10-28
AgentData reference removal test/example

Robadob:
Needs some review/refinement

#466

PR #980, opened 2022-11-16
Fix Python Tests

Robadob:
Broken by #952

Have implemented a suitable workaround for message descriptions (https://github.com/FLAMEGPU/FLAMEGPU2/issues/982), whilst we consider our options. It should be fine to deploy the next release with this workaround in place.


ptheywood (approved):
Fixes the test suite for me, and seems ok as a (hopefully) shorter term workaround. 

```
617 passed, 11 skipped
```

PR #981, opened 2022-11-17
Add swig wrapper dependencies on flamegpu header files

### LINKED ISSUE #979: Swig CMake incremental build missing dependency?
When I split the swig building into multiple CMake targets for better incremetnal builds, I missed adding a depenceny on the `PYTHON_SWIG_TARGET_NAME` target to make it depend on the pyflamegpu header files? 

I.e. currently it is possible to get `flamegpuPYTHON_wrap.cxx` file compilation errors when the header files have changed on disk, but CMake thinks it's up to date. 

It might be a bit of a timestamp race to achieve this, via cancelling builds and checking out between branches

I think this is because `swig_add_library` depends only on `${SWIG_INPUT_FILE_NAME}.i` with a link depenceny on `flamegpu`, so the cxx generation is only tied to the `.i` timestamp and not also the header files. Just passing the list of header files as SOURCES too should resolve this based on the UseSwig docs?

ptheywood:
Generation of the swig wrapper file was not dependent on flamegpu header files, so changes to header files might not cause a regeneration of the wrapper.

CMake 3.22+ would be required for automatic dependency generation via Swigs -M option, but we want to maintain 3.18 minimum.

Fortunately as we have an explciit list of our header files anyway its easy enough to explicitly list these via teh DEPENDS swig source property

Closes #979


Robadob (approved):
Tested it, forced rebuild of flamegpu.i after tweaking a random lib header

PR #983, opened 2022-11-17
Runplanvector redux

### LINKED ISSUE #681: RunPlanVector potential changes/enhancements
While testing / working with RunPlans and RunPlanVectors several potential improvements came up, but to avoid blocking bugfixes they're getting their own issue.

+ [ ]`RunPlanVector::setPropertyUniformDistribution` has a potentially misleading/ambiguos name, which is similar to setPropertyUniformRandom`, which uses a Unform distribution. It should probably be renamed. It's more of a LERP imo, but there may be a more canonical name for this for similar use-cases.
+ [ ] `operator==` would be useful to compare `RunPlan` instances (and potentially `RunPlanVector`s) against one another. greatly simplifying some tests
+ [ ] `RunPlanVector::at(size_t index)`. I.e. expose the vectors .at methods, which is `operator[]` with bounds checking.

Robadob:
C and Python test suites both pass on Windows/Release/Seatbelts=ON

Closes #681

### DIFF
--- a/examples/ensemble/src/main.cu
+++ b/examples/ensemble/src/main.cu
@@ -60,9 +60,9 @@ int main(int argc, const char ** argv) {
     {
         runs.setSteps(STEPS);
         runs.setRandomSimulationSeed(12, 1);
-        runs.setPropertyUniformDistribution<int>("init", 0, 9);
-        runs.setPropertyUniformDistribution<int>("init_offset", 1, 0);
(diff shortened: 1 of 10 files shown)


mondus (changes requested):
Suggested change for `setPropertyUniformDistribution` should be `setPropertyLerpRange`.

PR #988, opened 2022-11-21
Fix compilation on Bessemer (GCC11?)

Robadob:
The crux of the issue was that `ModelDescription.h` includes `CUDASimulation.cuh` and `CUDAEnsemble.cuh`. This has the knock-on effect of causing cuda headers to get included in a wide number of places unnecessarily, and for some reason Bessmer/GCC11 causes curand to include cuda headers as C rather than C++.

Many other files were relying on the `ModelDescription.h` include hierarchy, so some additional includes had to be added elsewhere.


ptheywood (approved):
Still builds locally with CUDA 11.4 and GCC 9 (GCC 11.3.0 available from the ubuntu repos is too new for CUDA 11.4, so couldn't repro the original issue locally anyway. 

This might be further justification for #78.

PR #990, opened 2022-11-21
Replace NVTX macros with constexpr + namespaced methods

### LINKED ISSUE #984: Use `if constexpr` for namespaced `NVTX_PUSH` etc
Since we're c++17, `if constexpr` is avialable for use, this can be used to replace existing preproc macros for `NVTX_RANGE/PUSH/POP`, to avoid symbols outside of the namespace. 

Efffectively we currently have code similar to 

```
#include <string>

#define USE_NVTX

#ifdef USE_NVTX
#define NVTX_PUSH(label) (nvtx_push(label))
#else
#define NVTX_PUSH(label)
#endif

void nvtx_push(const char * label) {
    printf("%s\n", label);
}

int nvtx_mock_test(int num) {
    NVTX_PUSH("test");
    std::string x = "s";
    return num * num;
}




```
Which produces optimised (-O3) assembly when USE_NVTX is defined:

```
nvtx_push(char const*):
        jmp     puts
.LC0:
        .string "test"
nvtx_mock_test(int):
        push    rbx
        mov     ebx, edi
        mov     edi, OFFSET FLAT:.LC0
        imul    ebx, ebx
        call    puts
        mov     eax, ebx
        pop     rbx
        ret
```

And when not defined
```
nvtx_push(char const*):
        jmp     puts
nvtx_mock_test(int):
        imul    edi, edi
        mov     eax, edi
        ret
```

https://godbolt.org/z/741no85Gd

Instead, we can declare the enablement/disablement of NVTX inside the namespace, and just call a method:

```
#include <string>

namespace f {
constexpr bool USE_NVTX=true;
void nvtx_push(const char * label) {
    if constexpr (USE_NVTX) {
        printf("%s\n", label);
    }
}

} // namespace f

// Type your code here, or load an example.
int nvtx_mock_test(int num) {
    f::nvtx_push("test");

    std::string x = "s";
    return num * num;
}


```

Optimised assembly when enabled:

```
f::nvtx_push(char const*):
        jmp     puts
.LC0:
        .string "test"
nvtx_mock_test(int):
        push    rbx
        mov     ebx, edi
        mov     edi, OFFSET FLAT:.LC0
        imul    ebx, ebx
        call    puts
        mov     eax, ebx
        pop     rbx
        ret
```

Optimised assembly when disabled: 

```
f::nvtx_push(char const*):
        ret
nvtx_mock_test(int):
        imul    edi, edi
        mov     eax, edi
        ret
```

https://godbolt.org/z/4jdq7K5x9

I.e. runtime cost will be the same as when using the macro, but we won't be polluting the (global) preprocessor

(we will still have to pass the CMake option to the preprocessor to set the initial vlaue, or configure a file, and the `if constexpr` will need to be in a header file not a source file)

---

Unfortunatley, we can't do this for our other macro functions. `THROW`, `DTHROW`, `gpuErrchk` etc make use of `__LINE__` / `__FILE__`, so the preproc needs to put those in place (`std::source_location` is c++20), and it doesn't make sense for return types (`FLAMEGPU_INNIT_FUNCTION` etc).

ptheywood:
+ Replaces NVTX_PUSH(label) with flamegpu::util::nvtx::push(label)
+ Replaces NVTX_POP() with flamegpu::util::nvtx::pop()
+ Replaces NVTX_RANGE(label) with flamegpu::nvtx::Range r(label)
+ Adds cosntexpr bool flamegpu::nvtx::ENABLED
+ Wraps flamegpu::util::nvtx with swig, exluding range
  + pyflamegpu.NVTX_ENABLED, pyflamegpu.nvtx_push(label), pyflamegpu.nvtx_pop()

Should have the same (C++) runtime cost in optimisaed builds

Closes #984

### DIFF
--- a/examples/circles_bruteforce/src/main.cu
+++ b/examples/circles_bruteforce/src/main.cu
@@ -65,8 +65,8 @@ FLAMEGPU_STEP_FUNCTION(Validation) {
     printf("%.2f%% Drift correct\n", 100 * driftDropped / static_cast<float>(driftDropped + driftIncreased));
 }
 int main(int argc, const char ** argv) {
-    NVTX_RANGE("main");
-    NVTX_PUSH("ModelDescription");
+    flamegpu::util::nvtx::Range range{"main"};
+    flamegpu::util::nvtx::push("ModelDescription");
     flamegpu::ModelDescription model("Circles BruteForce");
 
     const unsigned int AGENT_COUNT = 16384;
@@ -114,14 +114,14 @@ int main(int argc, const char ** argv) {
         layer.addAgentFunction(move);
     }
 
-    NVTX_POP();
+    flamegpu::util::nvtx::pop();
 
     /**
      * Create Model Runner
      */
-    NVTX_PUSH("CUDASimulation creation");
+    flamegpu::util::nvtx::push("CUDASimulation creation");
     flamegpu::CUDASimulation  cudaSimulation(model, argc, argv);
-    NVTX_POP();
+    flamegpu::util::nvtx::pop();
 
     /**
      * Create visualisation
--- a/examples/diffusion/src/main.cu
+++ b/examples/diffusion/src/main.cu
@@ -44,8 +44,8 @@ FLAMEGPU_EXIT_CONDITION(stable_temperature) {
 int main(int argc, const char ** argv) {
     const unsigned int SQRT_AGENT_COUNT = 200;
     const unsigned int AGENT_COUNT = SQRT_AGENT_COUNT * SQRT_AGENT_COUNT;
-    NVTX_RANGE("main");
-    NVTX_PUSH("ModelDescription");
+    flamegpu::util::nvtx::Range range{"main"};
+    flamegpu::util::nvtx::push("ModelDescription");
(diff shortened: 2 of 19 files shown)


ptheywood:
Ranges can be generated from python now: 

```
nsys profile -o test python3 -m pytest ../tests/swig/python/util/test_nvtx.py 
```

image

Might need some tweaking if / when we have a static installable build of the c++ library, as the ENABLED property is just fed from the macro.

mondus (changes requested):
Suggest to disable `Range` in Python to avoid potential Python garbage collection problems.

ptheywood:
Now re-adjusted to:

+ uses a namespace again, rathe rthan class
+ lowercase `range`
+ does not wrap range, python is now `nvtx_push` via rename (due to use of namespace not class)
+ fixed a lifetime bug in C++ (need to assign to a variable for scoping to be useful, another minor downside compared to the macro). 


image

Needs a squash, but subject to windows specific warnings about unused variables / masking this should be good to re-review.

Robadob (approved):
Changes look good, will merge after CI is happy.

PR #991, opened 2022-11-21
CMake option/cache variable naming

### LINKED ISSUE #448: Centralise where CMake options are converted to compiler options
Various CMake options (e.g. `SEATBELTS`, `VISUALISATION` etc) enable/disable compiler flags in multiple places. This scattered nature makes it easy to miss changing all of them should they change.

It would be more maintainable if there were a central function/s to handle converting these CMake options to the relevant compiler flags.

Initially it was thought that would be possible by just making a macro that calls the `add_compile_definitions(<definition>)` with the appropriate definitions, however SWIG instead requires.

```
set_property(TARGET ${PYTHON_MODULE_NAME} APPEND PROPERTY SWIG_COMPILE_DEFINITIONS <definition>)
set_property(TARGET ${PYTHON_MODULE_NAME} APPEND PROPERTY SWIG_GENERATED_COMPILE_DEFINITIONS <definition>)
```

### LINKED ISSUE #732: Modernise GLM CMake
The GLM CMake was not modernised during the big CMake modernisation, so it still relies on global cmake variables rather than using a header only target. 

This should be improved to match other header only targets.

Unfortunately, recent GLM seems to have dropped the use of version numbers / hasn't made a new versioned relese since april 2020, with almost 200 commits since this release, so using CMake to check the correct GLM was found might not be viable.

In addition, the vis repo GLM should be modenised together with this. Ideally with both repos requireing the same glm version, and only downloading it once. 

Alonside the use of targets, the main flamegpu library must also inform RTC of include paths. This is currently handled by a macro definiation `GLM_PATH`, forwarding the include path from CMake. This is non portable, so we need to address this (part of #260, maybe #607). 

Further more, 2.0.0a1 `USE_GLM` builds only work when visualiseation is enabled. This will be fixed  quickly in the short term with a very minor change, to be improved at a later date.

+ [ ] Modernise GLM use, to use an `imported interface` target (see `FindJitify.cmake`)
+ [ ] Ideally find existing GLM versions and check if it is sufficiently new. This may not be possible as GLM seems to have stopped making new versioned releases
+ [ ] Update vis repo GLM to match
+ [ ] Use the same GLM in both repositories if required. I.e. only download if GLM was not already in _deps. Without usable version numbers it will be hard to support this and allow use-provided glm
+ [ ] Optionally: Expand CI to include a USE_GLM=ON build(s) to ensure it works in the future.

### LINKED ISSUE #837: CMake: Use `cmake_dependent_option` to change default `BUILD_` flags
When including FLAME GPU 2 in other repositories, users only want the library itself, rather than the tests and examples etc by default. 

E.g. in the template repository

```
    set(NO_EXAMPLES ON CACHE INTERNAL "-")
    set(BUILD_TESTS OFF CACHE BOOL "-")
    mark_as_advanced(FORCE BUILD_TESTS)
```
[FLAMEGPU/FLAMEGPU2-example-template cmake/flamegpu2.cmake:L41-L43](https://github.com/FLAMEGPU/FLAMEGPU2-example-template/blob/8b8c6075ccde39fb720f5b64bfd013566d31328f/cmake/flamegpu2.cmake#L41-L43)

These options should be on by default for standalone FLAME GPU 2 CMake configurations however.

It should be possible to achieve this behaviour with `cmake_dependent_option`, checking if the main CMAKE Project Name is the same as the main `CMakeLists.txt` `PROJECT NAME`

Something like:
```
cmake_dependent_option(BUILD_TESTS "Build tests" ON "CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME" OFF)
```

From an CMake inclusion perspective, it might be better to also rename these options when included, so they are prefixed with FLAMEGPU, but I'm less sure/confident about if that's sensible or not.

I.e. when configuring the main FLAME GPU 2 repository, use `-DBUILD_TESTS=ON`, but when configuring a project which `add_subdirectory(FLAMEGPU)` it should be `-DFLAMEGPU_BUILD_TESTS` so it doesn't conflict with local tests? This is a bigger change

### LINKED ISSUE #976: Switch to CMake CUDA_ARCHITECTURES
CMake 3.18 introduced [`CMAKE_CUDA_ARCHITECTURES`/`CUDA_ARCHITECTURES`](https://cmake.org/cmake/help/latest/prop_tgt/CUDA_ARCHITECTURES.html) for CMake-native gencode options setting. 

At the time we didn't adopt this due to supporting older CMake versions, and (iirc) the original version might not have embedded PTX. 

It would be better to adopt the use of this rather than our `CUDA_ARCH` cmake variable for consistency with other CMake >= 3.18 projects, and to reduce the confusing duplicate options from the CMake GUI. 

CMake 3.24 also adds `native` as an option to just build for the cuda arch in the current machine, which might be useful (but shouldn't be the default IMO / can't be if we don't want to bump our cmake version)  

---

+ [x] Check if this embeds PTX / if we could make it embed PTX for future arch JITTING
+ [x] Replace use of `CUDA_ARCH` with `CMAKE_CUDA_ARCHITECTURES`/`CUDA_ARCHITECTURES`
+ [x] Figure out how to deal with setting a default value for this if not specified (/ seeing what CMake's default does / if it is sensible)
+ [x] Update readme

ptheywood:
Refactor CMake: Option names, use CMAKE_CUDA_ARCHITECTURES, general improvments

+ Adds top-level CMakeLists.txt detection to the root CMakeLists.txt
+ Prefixes CMake options / CACHE variables with FLAMEGPU_ to cooperate with others, including:
  + BUILD_EXAMPLE_<example> -> FLAMEGPU_BUILD_EXAMPLE_<example> as dependent options
  + FLAMEPGU_BUILD_ALL_EXAMPLES -> FLAMEGPU_BUILD_ALL_EXAMPLES
    + Defaults to ON only if the main FLAMEGPU CMakeLists is the top level, removing the need for NO_EXAMPLES
  + BUILD_SWIG_PYTHON -> FLAMEPGU_BUILD_PYTHON
  + USE_NVTX -> FLAMEGPU_ENABLE_NVTX
  + USE_GLM -> FLAMEGPU_ENABLE_GLM
  + CURAND_ENGINE -> FLAMEGPU_CURAND_ENGINE, including a dropdown in the GUI
  + Removes USE_NVCC_THREADS
  + NVCC_THREADS -> FLAMEGPU_NVCC_THREADS which defaults to 2, and handles being set to OFF correctly.
  + VERBOSE_PTXAS -> FLAMEGPU_VERBOSE_PTXAS
  + BUILD_TESTS -> FLAMEGPU_BUILD_TESTS
  + BUILD_TESTS_DEV -> FLAMEGPU_BUILD_TESTS_DEV
  + EXPORT_RTC_SOURCES -> FLAMEGPU_RTC_EXPORT_SOURCES
  + RTC_DISK_CACHE -> FLAMEGPU_RTC_DISK_CACHE
  + WARNINGS_AS_ERRORS -> FLAMEGPU_WARNINGS_AS_ERRORS
  + SEATBELTS -> FLAMEGPU_SEATBELTS
  + BUILD_API_DOCUMENTATION -> FLAMEGPU_BUILD_API_DOCUMENTATION
  + VISUALIASTION -> FLAMEGPU_VISUSALISATION
    + And associated sub-optons FLAMEGPU_VISUALISATION_ROOT etc.
  + USE_GTEST_DISCOVER -> FLAMEGPU_ENABLE_GTEST_DISCOVER
+ Switch to CMAKE_CUDA_ARCHITECTURES from CUDA_ARCH with custom default setting
+ Mark many CMake Cache varaibles as advanced to exclude them from the default CMake GUI
+ Improves cpplint CMake logic
+ Move definition location of some CMake options, so they are pulled into individual examples
+ Prefix Cmake functions and macros with flamegpu_ and rename to all lower case, including:
  + add_flamegpu_executable -> flamegpu_add_executable
+ Use target-specific c++17 specification
+ Correct cmake_minimum_required statements and add a <polciy_max> of 3.25
+ Modernise GLM CMake
+ De-duplicate compiler definition for the pyflamegpu target, using properties from the flamegpu target
+ Reduce use of cmake variables from the visualiser CMake, instead request a list of (dll) files to copy is needed
+ set target compile options on targets where possible
+ Prefixes preprocessor macros with FLAMEGPU

Closes #837
Closes #976
Closes #448 
Closes #732


ptheywood:
The CMake GUI for a configuraiton with python, nvtx, tests and CUDA_ARCH=86 under linux currently looks like:  

image

Which can be grouped

image

And ADVANCED exposes a bunch of options which most users shouldn't mind. 

For now, I've left CUDA_ARCH alone, as it might be dissapearing (#976). 

Similar changes need making to the vis repo, and the docs repo will need a tweak as well. 

I've potentially still missed some CACHE variables / conditional options that are not in the top level CMake or other obvious locations, so I need to hunt those down. 

---

@Robadob - any thoughts on `FLAMEGPU_ENABLE_GLM` instead of `USE_GLM` (and any other general cmake option name's that you'd like changing while i'm at it)?
Seems to be a mix of `ENABLE` vs `USE` in others, and apparently I was leaning towards `ENABLE` yesterday, maybe because it enables functionality rather than using it internally?

Renaming `BUILD_SWIG_PYTHON` to `FLAMEGPU_BUILD_PYTHON`is also maybe contentious, but we have no immediate plans for an alterante python implementation, which would 100% be a new major version anyway so I think its fine / slightly less typing for me (Though I've setup `CMakeUserPresets.json` which is quite nice to use) 

---

I Might renable the NVCC_THREADS related options too, so they are neighbours in the list, not 100%.

Robadob:
> @Robadob - any thoughts on FLAMEGPU_ENABLE_GLM instead of USE_GLM (and any other general cmake option name's that you'd like changing while i'm at it)?
Seems to be a mix of ENABLE vs USE in others, and apparently I was leaning towards ENABLE yesterday, maybe because it enables functionality rather than using it internally?

No preference.

ptheywood:
> P.S. Not seen that grouping before.

It's a checkbox next to the Advanced checkbox, just seems to group cache vars by splitting upto the first `_`.

> The [cmake-gui(1)](https://cmake.org/cmake/help/latest/manual/cmake-gui.1.html#manual:cmake-gui(1)) tool can display options in groups defined by their prefix, so it makes sense for third parties to ensure that they use a self-consistent prefix.
[docs](https://cmake.org/cmake/help/latest/guide/user-interaction/index.html#setting-build-variables)

ptheywood:
Configuring individual examples from the individual example directory, has the following advanced grouped GUI entry, this is due to where options / cache variables are mostly set (the root CMakeLists.txt).

`CUDA_ARCH` doesn't appear eitherm thgouh that might not be a problem longer term. 

image


This is a little complicated though, as to what options should be available in this case. 



A lot of options e.g. python/tests/exampeles aren't relevant, but other options are:

+ `FLAMEGPU_ENABLE_NVTX`
+ `FLAMEGPU_EXPORT_RTC_SOURCES`
+ `FLAMEGPU_CURAND_ENGINE`
+ `FLAMEGPU_NVCC_THREADS`
+ `FLAMEGPU_ENABLE_NVCC_THREADS
+ `CUDA_ARCH` (for now)

`FLAMEGPU_BUILD_API_DOCUMENTATION` should probably be off by default? though its not part of all so maybe just advanced? 

I guess these options should be deinfed in `src/CMakeLists.txt` or `cmake/common.cmake` or their own cmake file?. It's a little bit chicken and egg, as they're internal subprojects, external projects which add_subdirectory() the main flamegpu repo will get them already (i.e. standalone).

ptheywood:
Testing in Windows was worth it as I had missed an important line of CMake when copying the dlls. 

Vis builds tested on python and windows, vis and non vis on linux have been tested. 

Also updated the freshly merged CTest logic in, making sure things are marked as advanced etc too. 

---

This is now ready for review, but prior to merge does need the vis hash updating after https://github.com/FLAMEGPU/FLAMEGPU2-visualiser/pull/110 has been merged.

A fresh build directory / configure is probably worthwhile if you want to see the correct state of the CMake UI after this change. 

It might actually be worth prefixing compile time defines with `FLAMEGPU_` too, given they are global scope, and `VISUALISATION` is required for use in model code. Can't provide a constexpr instead, as the API doesn't exist if `VISUALISATION` is off (and changing that is a load of faff).

Robadob (approved):
Changes seem thorough and worthwhile.

Nothing obvious stands out as missing, I'm not in a position to do a local build but CI seems happy.

ptheywood:
I've now also prefixed the preprocessor macros with FLAMEGPU, to avoid windows.h MAX style issues.

I've also made the preproc definitions CMake target public where they are required in flamegpu headers (so they match the static lib, even if the user would attempt to redefine them). 

And also made a minor fix to CMAKE_CUDA_ARCHITECTURES, as in the vis repo. 

Commits are still individual for now, but will manually squash once the vis hash is finalised.

Robadob (approved):

mondus (approved):

ptheywood:
This should be in a state to be merged once CI passes, however the telemetyr PR has slightly broken python testing, when running the tests on `../tests/swig` not `../tests/swig/python`, which worked before. Probably because of `conftest.py` somehow? 

```
python3 -m pytest ../tests/swig/
========================================================================================================= test session starts ==========================================================================================================
platform linux -- Python 3.10.6, pytest-7.2.0, pluggy-1.0.0
rootdir: /home/ptheywood/code/flamegpu/FLAMEGPU2
collected 659 items                                                                                                                                                                                                                    

../tests/swig/python/test_version.py .                                                                                                                                                                                           [  0%]
../tests/swig/python/codegen/test_codegen.py ..................................................                                                                                                                                  [  7%]
../tests/swig/python/codegen/test_codegen_integration.py .
```

It just stops there rather than progressing, with no errors etc. 


Running via `python3 -m pytest ../tests/swig/python/` behaves correctly.

PR #992, opened 2022-11-22
CUDAEnsemble truncate log files.

### LINKED ISSUE #818: CUDAEnsemble log files always append
Noticed by Kenneth, if you run an ensemble with logging twice in a row all the files will be appended, rather than truncated.

The fix for step logs, should be rather simple change the last parameter [here](https://github.com/FLAMEGPU/FLAMEGPU2/blob/3b3406e7ba94f45314405ad0e85fd010ab9b6854/src/flamegpu/sim/SimLogger.cu#L96) to `true`, and step log files will always truncate when logged via `CUDAEnsemble`.

Exit log files are a bit more complicated, as they are intended to append. So it would be necessary to delete any exit log files at the expected locations before outputting any logs.

There's probably some discussion to be had, over whether it would make more sense to purge the entire output dir before logging, to remove any confusion if the two runs create different files.

Robadob:
Ensemble and simulation now have the argument `--truncate` or config variable `truncate_log_files`.

If `true`, this causes all output files to overwrite existing files (and in the case of ensemble exit logs, it will simply delete files at the start).

If `false` a `FileAlreadyExists` exception will be raised. The existence of files is checked at the start of calls to `simulate()` in the case of ensemble runs to ensure early exit, rather than after waiting for compute. Simulation's however perform this as the time of logging (to avoid repetition, due to those methods being possible to call in isolation).

- [x] CUDAEnsemble
- [x] CUDASimulation
- [x] Tests

Closes #818

### DIFF
--- a/include/flamegpu/exception/FLAMEGPUException.h
+++ b/include/flamegpu/exception/FLAMEGPUException.h
@@ -399,6 +399,10 @@ DERIVED_FLAMEGPUException(VersionMismatch, "Versions do not match");
  */
 DERIVED_FLAMEGPUException(InvalidFilePath, "File does not exist.");
 /**
+ * Defines an error reported when an output file already exists, so would be overwritten
+ */
+DERIVED_FLAMEGPUException(FileAlreadyExists, "File already existst.");
+/**
  * Defines an exception indicating that the flamegpu::util::detail::Timer has been used incorrectly.
  */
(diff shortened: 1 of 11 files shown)


ptheywood (approved):
Looks good, both test suites pass under linux. 

I've added 2 tests for `--truncate` argparsing to python, just to make sure it's available and the config properties can be accessed.

PR #994, opened 2022-11-23
Remove resolved / no longer relevant known issues from the readme

ptheywood:
Spotted these no longer  relevant known issues in the readme, which have either been fixed or are no longer relevant due to CUDA/CMake minimum version changes.


Robadob (approved):
Didn't even realise we had these listed in README.

PR #997, opened 2022-11-24
Breaking Change: Remove "callback" from Python host function related methods/classes.

Robadob:
This better unifies it with the C API.

Also fixed a bug, whereby the `__disown__()` fix (#975), had failed to address `addExitCondition()`.

Still want to find a way to remove, or make private `LayerDescription::_addHostFunction()`.


Robadob:
Appears some changes in this PR have not worked as intended and  have broken the `__disown__()` fix (appears the add init/step/exit function methods are now treated as overloads.

e.g.

```py
    def addInitFunction(self, *args):
        r"""
        *Overload 1:*

        Adds an init function to the simulation
        Init functions execute once before the simulation begins
        :type func_p: void
        :param func_p: Pointer to the desired init function
        :raises: exception::InvalidHostFunc If the init function has already been added to this model description
        Notes: Init functions are executed in the order they were added to the model

        |

        *Overload 2:*

        Adds an init function callback to the simulation. The callback objects is similar to adding via addInitFunction
        however the runnable function is encapsulated within an object which permits cross language support in swig.
        Init functions execute once before the simulation begins
        :type func_callback: :py:class:`HostFunctionCallback`
        :param func_callback: Pointer to the desired init function callback
        :raises: exception::InvalidHostFunc If the init function has already been added to this model description
        Notes: Init functions are executed in the order they were added to the model
        """

        try:
            func_callback = func_callback.__disown__()
        except:
            pass


        return _pyflamegpu.ModelDescription_addInitFunction(self, *args)


```

Robadob:
Ah think I've solved the issue. I had ignored the functions incorrectly specifying `FLAMEGPU_HOST_FUNCTION_POINTER` rather than the more specific init/step/exit aliases.

include/flamegpu/model/LayerDescription.h
```diff
@@ -219,19 +223,20 @@ class LayerDescription : public CLayerDescription {
      * @throw exception::InvalidHostFunc If the function has already been added to the layer
      * @note ONLY USED INTERNALLY AND BY PYTHON API - DO NOT CALL IN C++ BUILD
      */
-    void _addHostFunctionCallback(HostFunctionCallback *func_callback);
+    inline void addHostFunction(HostFunctionCallback *func_callback);
```
ptheywood, line 226: Just checking this is still intentionally `HostFunctionCallback` in the c++ portion (This is the first occurance in the diff). 

Its fine it if is, just thought i'd highlight it.

ptheywood (approved):
Looks good, pytest all passes under linux. 

Raised one comment about a c++ class with Callback in the name still incase it was missed rather than intentionally still there, but its fine if it is intentional.

Robadob:
Assuming CI passes this can be merged.

There's something I don't think should compile (a returned struct before it's used by a prototype, so size unknown), but it worked for me locally.

PR #999, opened 2022-11-25
Expose JitifyCache::clearDiskCache() to Python API

Robadob:
Additionally made the method static as it does not require the instance.


Robadob:
Todo: Create a public API wrapper for clearing cache.

ptheywood (approved):

PR #1002, opened 2022-11-29
Duplicate several C ensemble tests for RTC.

### LINKED ISSUE #835: C++ RTC Ensemble Test(s)
The C++ test suite does not test ensembles which make use of RTC kernels. 

This may be implicitly covered by any python tests of the ensembles, and duplicating all tests to be rtc/non rtc would be very time consuming, but there is explicit RTC ensemble code so this should be in the c++ test suite, checking the simualtions within an ensemble behave as intended.

This should include test(s) with submodels involved, as #843 fixed an issue encountered by ensembles containing submodels. (RTC and non RTC ensemeble submodel test(s) would be a good addition)

Robadob:
Closes #835

Windows/Debug
```
[----------] 29 tests from TestCUDAEnsemble (5541 ms total)

[==========] 29 tests from 1 test suite ran. (5543 ms total)
[  PASSED  ] 29 tests.
```


ptheywood (approved):
1055 passes under linux.

PR #1007, opened 2022-12-08
Filename shuffle

### LINKED ISSUE #550: Class / Namespaces to Rename / Move

Robadob:
Requires fully CI build to check it works

- [x] Lots of files have moved into detail and need adding to the namespace.
- [x] Some test suites should now be moved to match the new file structure?
- [x] Remove `CUDASimulation::getAgent()` (it returns type `detail::AgentInterface` it was added for hacky reasons to access agent data on the device which are probably nolonger relevant.)

Closes #550


Robadob:
Just building test suite locally, but I think this is ready to merge.

Robadob:
Windows/Debug/Seatbelts=ON test suite built locally.

```
[==========] 1063 tests from 82 test suites ran. (379961 ms total)
[  PASSED  ] 1063 tests.

  YOU HAVE 53 DISABLED TESTS
```

(After correcting an issue in `JitifyCache::getIncludeDir()` that had been broken whilst updating include dirs with find-replace.)

ptheywood (approved):
File moves look good, Tests and pytest all pass under linux. 

~Should be fine to merge once the pytests are restructured similar to the cxx tests.~

Issue #646, opened 2021-08-18
Test suite compilation issues with multiple CUDA_ARCH


ptheywood:
When building the c++ tests suite with many cuda architectures enabled, significant amounts of host resource can be consumed, potentialyl causing compilation failures (I.e. CI errors during release CI)

This was demonstrated by Windows CUDA 11.4 build with `CUDA_ARCH="35 52 60 70 80"` triggering an out of memory issue when device-linking the test suite.

```
     1>C:/Users/RUNNER~1/AppData/Local/Temp/tmpxft_00000814_00000000-8_tests.device-link.fatbin.c(9423890): fatal error C1060: compiler is out of heap space [D:\a\FLAMEGPU2\FLAMEGPU2\build\tests\tests.vcxproj]
```

Additionally, with linux CI a `Segmentation fault` was encountered during test suite compilation, which is likely related.

For now on CI this has (hopefully) been addressed by building fewer cuda architectures when building the test suite, and reducing the number of threads used by `nvcc` when using the `--threads` parameter.

Longer term we have several options:

+  Reduce the amount of code being device linked somehow, so fatbin linking is not generating 9 million LOC
+ Split the C++ test sute into multiple binaries, which can then be orchestrated by a new `tests` binary, or through `ctest` (see #267 for more information, and #285 for a very stale attempt at adding ctest)

ptheywood:
@Robadob has encountered heap size related compialtion errors when building a debug configuration of the test suite under windows. Splitting the c++ test suite and orchestrating via ctest would be one way to address the challenge (also addressing this issue)

Robadob:
Visual Studio provides both x86 and x64 linkers. By default it uses x86 (and is supposed to switch to x64 if the heap is exceeded).
There are alot of circle 2013 stack overflow posts saying how you can specify to use x64, this CMake var corresponds to that method. There's no recent discussion, so it's possible it nolonger has an impact in newer visual studios.

https://cmake.org/cmake/help/latest/variable/CMAKE_VS_PLATFORM_TOOLSET_HOST_ARCHITECTURE.html

PR #1008, opened 2022-12-08
Install python dependencies in windows draft-release CI

ptheywood:
These were missed previously, but automatically installed if missing until a recent-ish change, and as this CI job doesn't run too often it was overlooked.

### DIFF
--- a/.github/workflows/Draft-Release.yml
+++ b/.github/workflows/Draft-Release.yml
@@ -236,6 +236,14 @@ jobs:
       uses: actions/setup-python@v4
       with:
         python-version: ${{ env.PYTHON }}
+    
+    - name: Install python dependencies
+      if: ${{ env.PYTHON != '' && env.FLAMEGPU_BUILD_PYTHON == 'ON' }}
+      run: |
+        python3 -m pip install --upgrade wheel build setuptools
+
+    - name: Add custom problem matchers for annotations
+      run: echo "::add-matcher::.github/problem-matchers.json"
 
     # Must pass -G -A for windows, and -DPython3_ROOT_DIR/-DPYTHON3_EXECUTABLE as a github action workaround
     - name: Configure cmake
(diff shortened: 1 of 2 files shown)


ptheywood:
Still 1 failure on the manual run of the draft-ci workflow, for the CUDA 11.8 build on a windows-2019 (i.e. VS 2019) machine, due to errors from the CUB/Thrust version used when building the flamegpu target.
```
2022-12-08T17:41:57.8397026Z      1>D:\a\FLAMEGPU2\FLAMEGPU2\build\_deps\thrust-src\dependencies\cub\cub\iterator\../util_device.cuh(569): error #20011-D: calling a __host__ function("cudaDeviceSynchronize") from a __host__ __device__ function("cub::SyncStream") is not allowed [D:\a\FLAMEGPU2\FLAMEGPU2\build\FLAMEGPU\flamegpu.vcxproj]
2022-12-08T17:41:57.8397763Z          
```

Unsure why this configuration results in errors, when the VS2022 + CUDA 11.8 regular CI is fine with current master, and VS2019 + CUDA 11.9 is unhappy. 

https://github.com/FLAMEGPU/FLAMEGPU2/actions/runs/3650583144/jobs/6166748488

It's found CUB/Thrust 1.14.0.0, which is the same as the 11.0 build, so the source should be the same so this makes very little sense.

Robadob (approved):
Changes look sane assuming the values provided are right.

I presume you've done a draft release with this branch to check?

ptheywood:
> I presume you've done a draft release with this branch to check?

Yes, but it has also modified the workflow triggers so that pull requests which modify `Draft-Release.yml` will trigger it, hence its doing 56 checks in total. 

This should help with keeping it correct / up to date given the infrequency of it being executed.

Robadob:
> Yes, but it has also modified the workflow triggers so that pull requests which modify Draft-Release.yml will trigger it, hence its doing 56 checks in total.

I saw that change, but hadn't clocked it's impact.

Robadob:
That one CI failure is out of heap space building tests -_-

> `C:/Users/RUNNER~1/AppData/Local/Temp/tmpxft_00000dd0_00000000-8_tests.device-link.fatbin.c(6894951): fatal error C1060: compiler is out of heap space [D:\a\FLAMEGPU2\FLAMEGPU2\build\tests\tests.vcxproj]`

ptheywood:
Yep its a known issue already, but previously 3 architectures was ok.  https://github.com/FLAMEGPU/FLAMEGPU2/issues/646 

I've turned it down to 2 arch's (min, max) from 3, (min, mid, max) which hopefully will link. 

Longer term we can either investigave msvc's flags related to setting heap size, or we can just split the test suite now that we have ctest integration. Just moving the reduction compilation unit to it's own binary would probably do the job to be honest.

PR #1015, opened 2022-12-12
CUDA 12.0

### LINKED ISSUE #1009: CUDA 12 support
CUDA 12 has now been released, although we can't opt-in to new features, we should make sure that FLAMEGPU 2 is usable with CUDA 12. 

It removes support for SM35, so CI jobs will need to use that as the oldest arch rather than SM 35. 

For release wheels, this will add an extra layer of release wheels however, but due to stabillity guarantees and CUDA 12.x build should be binary compatible with other CUDA 12.0 builds. 

The only potential issue I can see from the change log is the CUB/Thrust bump from 1.15.0 to 2.0 (we currently pull 1.14 as our own minimum). The simplest thing to do is probably required 2.x, and pull it back into CUDA 11.x, which I expect will still work. This will also solve some other warning based issues and macro based issues not fixed till 1.16.0

+ [x] Add 12.0 regular CI jobs (so 11.0, 11.8 and 12.0 to cover all min/major versions of 11.x and 12.x, arguably we could drop 11.8 if CI time is getting too big). SM 35 for 11.x, SM 52 for 12.x
+ [x] Add 12.0 wheels to the draft release matrix, for all major arch's. This will double our wheel count however. Long term we probably jsut want one 11.2+ wheel and one 12.0+ wheel. (11.0 is only provided for colab currently, otehrwise 11.2+ has abi stability).
+ ~Potentially upgrade CUB minimum to 2.0, if required (or desired)~
  + Deffered to #1021 
+ [x] `__init__.py` has some dll specific logic which is likely to need adjusting for 12.x. 
  + We correcntl implemented the > 11 branch of this already / Nvidia did what they said they would
+ [x] Run all tests on linux
+ [ ] Run all tests on windows 

This will be done after rc.0, but we probably want to do this for an rc.1 / for the final build to get 12.0 wheels out. 

The thrust/cub repo's have PRs in place to make the 2.0 releases from the cuda toolkit in place, so we can switch to that when it is out. Some tweaks do appear needed for the CMake call to for libcudacxx to be found in CUDA 11.x builds

### LINKED ISSUE #1019: Wheel CUDA architectures / PTX embedding.
Since the change to `CMAKE_CUDA_ARCHITECTURES`, specfying multiple cuda architectures as integers will embed ptx for each architecture. I.e. 

`35;52;60;70` translates to `35-virtual;35-real;52-virtual;52-real;60-virtual;60-real;70-virtual;70-real`, `virtual` is ptx embedding.

Previously, the old `CUDA_ARCH` behaviour was equivalent to `35-real;52-real;60-real;70-real;70-virtual`, i.e. only embed ptx for the most recent arch. This is almost equivalent to the recent nvcc `all-major` option, with one key exception: SM 50 devices would not be usable, as no older ptx is embeded, and there is no real arch for an older device with the same major architecture value. 

In practice, this should be `50` instead of `52` (though there are not many sm 50 devices, so it's not been an actual issue so far). 

We should probaly do 2 things for rc.1/release:

+ [ ] Replace 52 with 50
+ [ ] Switch to only embedding ptx for the final arch, either use `all-major` if CMake is recent enough, or use `35-real;50-real;60-real;70-real;70-virtual` for the wheels to reduce compilation time and binary file size.

ptheywood:
CUDA 12 has been released, this PR:

+ Adds CI builds of CUDA 12, using SM 50 as the minimum arch
+ Updates the fallback arch detection to remove SM35 if CUDA >= 12.0
+ Updates release wheels to be 11.2 and 12.0. 
  + 11.0 is not strictly required anymore, as 11.2 has much wider compatibility and 11.0 was only produced for google colab, which is now 11.2.
+ Updates CUB/Thrust finding logic to be more robust
+ Enables Wreorder on linux for CUDA 11.3+, and on windows for CUDA 11.5+, as these versions have usable suppressions for CUB/Thrust <= 2.1.0

Includes not directly related fixes, highlighted by new warnings:

+ Incorrect if statement in test_device_agent_creation.cu
+ Fix array comparison in MessageSpatial2DDevice.cuh
+ Fix licence badge in README.md
+ Makes finding rapidjson after fetching more specific.

Tested by: 

+ [x] C++ and python tests on linux using CUDA 12.0
+ [x] C++ and python tests on windows using CUDA 12.0
+ [x] CI demonstration of thrust / cub suppression changes: 
  + [Windows - 8 builds](https://github.com/FLAMEGPU/FLAMEGPU2/actions/runs/3714572192)
  + [Ubuntu - 10 builds](https://github.com/FLAMEGPU/FLAMEGPU2/actions/runs/3714572197)

Closes #1009 
Closes #1019

### DIFF
--- a/.github/scripts/install_cuda_windows.ps1
+++ b/.github/scripts/install_cuda_windows.ps1
@@ -2,23 +2,23 @@
 ## Constants
 ## -------------------
 
-# Dictionary of known cuda versions and thier download URLS, which do not follow a consistent pattern :(
-# From 11.0, the download URLs have changed as verisoning within CUDA has changed. 
-# The toolkit verison is separate from the individual versions of tools, i.e. toolkit 11.1.0 contains cudart 11.1.74, but CUPTI 11.1.69  
+# Dictionary of known cuda versions and thier download URLS, which do not follow a consistent pattern
+# From 11.0, the download url/toolkit version is separate from the cudart version.
+# Releases since 11.5.1 (including 11.4.4) use `windows` rather than `win10` in the uri, due to windows 11 inclusion
 $CUDA_KNOWN_URLS = @{
-    "8.0.44" = "http://developer.nvidia.com/compute/cuda/8.0/Prod/network_installers/cuda_8.0.44_win10_network-exe";
-    "8.0.61" = "http://developer.nvidia.com/compute/cuda/8.0/Prod2/network_installers/cuda_8.0.61_win10_network-exe";
-    "9.0.176" = "http://developer.nvidia.com/compute/cuda/9.0/Prod/network_installers/cuda_9.0.176_win10_network-exe";
-    "9.1.85" = "http://developer.nvidia.com/compute/cuda/9.1/Prod/network_installers/cuda_9.1.85_win10_network";
-    "9.2.148" = "http://developer.nvidia.com/compute/cuda/9.2/Prod2/network_installers2/cuda_9.2.148_win10_network";
-    "10.0.130" = "http://developer.nvidia.com/compute/cuda/10.0/Prod/network_installers/cuda_10.0.130_win10_network";
-    "10.1.105" = "http://developer.nvidia.com/compute/cuda/10.1/Prod/network_installers/cuda_10.1.105_win10_network.exe";
-    "10.1.168" = "http://developer.nvidia.com/compute/cuda/10.1/Prod/network_installers/cuda_10.1.168_win10_network.exe";
-    "10.1.243" = "http://developer.download.nvidia.com/compute/cuda/10.1/Prod/network_installers/cuda_10.1.243_win10_network.exe";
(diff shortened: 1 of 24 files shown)


ptheywood:
This has passed all tests for both suites on windows (debug) and linux (release) using CUDA 12. 

Could tidy history a little more and keep separate commits, but will wait for feedback to avoid triggering another 65 CI builds (or just squash it all). 

Only thing really worth thinking about is the dropping of the 11.0 wheels. It's a worthwhile tradeoff IMO. 11.2 and 12 provides the maximum compatibility for the minimum number of wheels. 11.0 users can still build from source, and 11.0 wheels are no longer required for collab.

Robadob (approved):
Changes seem thorough, I can't spot any issues looking at the changelog and your desc says you've tested on Windows so I haven't bothered doing so myself.

PR #1028, opened 2023-01-11
BugFix: Curl would write to stdout/err if old

### LINKED ISSUE #1027: Telemetry: Curl errors being output to stderr/stdout
When telemetry is enabled, but the version of curl available at runtime is rather old (centos 7 hpc) it emits errors to stderr/stdout, which is shown to the user, for each failed attempt.

This was intended to be silent? 

```
curl: option --connect-timeout: expected a proper numerical parameter
curl: try 'curl --help' or 'curl --manual' for more information
```

Currently there is no checking of curl availability or if the version is sufficiently new, with the original intent being to just silently fail iirc.

Robadob:
@ptheywood Had this on HPC with Curl 7.29.0 (circa Feb 2013)
```
curl: option --connect-timeout: expected a proper numerical parameter
curl: try 'curl --help' or 'curl --manual' for more information
```

This attempts to pipe output of the system call to `/dev/null`, haven't tested it though.

Closes #1027

### DIFF
--- a/src/flamegpu/detail/TestSuiteTelemetry.cpp
+++ b/src/flamegpu/detail/TestSuiteTelemetry.cpp
@@ -17,11 +17,15 @@ bool TestSuiteTelemetry::sendResults(std::string reportName, std::string outcome
     // generate telemetry data
     std::string telemetry_data = flamegpu::io::Telemetry::generateData(reportName, telemetry_payload);
     // send telemetry
-    flamegpu::io::Telemetry::sendData(telemetry_data);
+    bool telemetrySuccess = flamegpu::io::Telemetry::sendData(telemetry_data);
     // print telemetry payload to the user if requested
     if (verbose) {
-        fprintf(stdout, "Telemetry packet sent to '%s' json was: %s\n", flamegpu::io::Telemetry::TELEMETRY_ENDPOINT, telemetry_data.c_str());
-        fflush(stdout);
+        if (telemetrySuccess) {
+            fprintf(stdout, "Telemetry packet sent to '%s' json was: %s\n", flamegpu::io::Telemetry::TELEMETRY_ENDPOINT, telemetry_data.c_str());
+            fflush(stdout);
+        } else {
+            fprintf(stderr, "Warning: Usage statistics for Test suite failed to send.\n");
+        }
     }
     return true;
 }
--- a/src/flamegpu/io/Telemetry.cpp
+++ b/src/flamegpu/io/Telemetry.cpp
@@ -8,6 +8,7 @@
 #include <iostream>
 #include <cstdio>
 #include <cctype>
+#include <sstream>
 
 #include "flamegpu/version.h"
 
@@ -255,9 +256,18 @@ bool Telemetry::sendData(std::string telemetry_data) {
 #else
     null = "/dev/null";
 #endif
-    std::string curl_command = "curl -s -o " + null + " --connect-timeout " + std::to_string(CURL_CONNECT_TIMEOUT) + " --max-time " + std::to_string(CURL_MAX_TIME) +  " -X POST \"" + std::string(TELEMETRY_ENDPOINT) + "\" -H \"Content-Type: application/json; charset=utf-8\" --data-raw \"" + telemetry_data + "\"";
+    std::stringstream curl_command;
+    curl_command << "curl";
+    curl_command << " -s";
+    curl_command << " -o " << null;
+    curl_command << " --connect-timeout " << std::to_string(CURL_CONNECT_TIMEOUT);
+    curl_command << " --max-time " << std::to_string(CURL_MAX_TIME);
+    curl_command << " -X POST \"" << std::string(TELEMETRY_ENDPOINT) << "\"";
+    curl_command << " -H \"Content-Type: application/json; charset=utf-8\"";
+    curl_command << " --data-raw \"" << telemetry_data + "\"";
+    curl_command << " > " << null << " 2>&1";
     // capture the return value
-    if (std::system(curl_command.c_str()) != EXIT_SUCCESS) {
+    if (std::system(curl_command.str().c_str()) != EXIT_SUCCESS) {
         return false;
     }
 
--- a/src/flamegpu/simulation/CUDAEnsemble.cu
+++ b/src/flamegpu/simulation/CUDAEnsemble.cu
@@ -240,15 +240,15 @@ unsigned int CUDAEnsemble::simulate(const RunPlanVector &plans) {
         #endif
         // generate telemetry data
         std::string telemetry_data = flamegpu::io::Telemetry::generateData("ensemble-run", payload_items);
-        // send
-        if (!flamegpu::io::Telemetry::sendData(telemetry_data)) {
-            if ((config.verbosity > Verbosity::Verbose)) {
+        // send the telemetry packet
+        bool telemetrySuccess = flamegpu::io::Telemetry::sendData(telemetry_data);
+        // If verbose, print either a successful send, or a misc warning.
+        if (config.verbosity >= Verbosity::Verbose) {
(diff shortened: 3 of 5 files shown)


ptheywood:
Related issues is #1027

ptheywood:
Running boids 2D with this build on bessemer in verbose mode claims to have sent the telemetry packet successfully, with no error produced. 

```
Telemetry packet sent to 'https://nom.telemetrydeck.com/v1/' json was: [{\"isTestMode\":\"false\",...
```

but I'm not seeing the associated data in telemetry deck. 

Might also be worth just changing the timeouts to `1` to avoid locale issues too (previously suggested by Rob?). Needs more testing though so not going to approve / reject yet.

Robadob:
> in verbose mode claims to have sent the telemetry packet successfully, with no error produced.

Yeah didn't realise I was supposed to handle that. Might be a way to detect the return code?

ptheywood:
> Yeah didn't realise I was supposed to handle that. Might be a way to detect the return code?

The function you modified is already set up to return false if the return code is non zero, so there wasn't an expectation that you would need to change anything tbh. 

After checking out master again, curl output was still there, was was the packet sending success so it was broken before too.

Robadob:
curl always returns exit success?

Probably need to play with manually calling curl and checking err code to find out how to solve that

ptheywood:
https://github.com/FLAMEGPU/FLAMEGPU2/blob/cfbe3c39b597781f9ca5d4f216cf100c580a8492/src/flamegpu/simulation/CUDASimulation.cu#L1251-L1258

It's set up that it should output a warning that it failed to send, and then it will also state the packet was sent anyway (its not an `else`). 

The same is probably true in Ensembles too (and test suite?), but I've not checked. 

Curl does return `130` when otuputting the `expected a proper numerical parameter`, so the `sendData` method is not returning flase when it should be.

Robadob:
> Curl does return 130 when otuputting the expected a proper numerical parameter, so the sendData method is not returning flase when it should be.

Are you sure there's not another arg in there e.g. `-o null` which is breaking the return code? Bc the code looks pretty clear as written (unless Paul's original stuff is somehow wrong).

ptheywood:
Stanage includes a too recent version of curl (7.83 from 2022) so doens't reproduce this issue, will need to test it out on bessemer or sharc.

ptheywood:
Running this on a K80 node in ShARC runs without error, claiminng the telemetry packet was sent. 

```
 ./bin/Release/boids_spatial3D -s 10 -t -v
FLAME GPU 2.0.0-rc.1+a089dd1
Simulation configuration:
        Random Seed: 1681820561
        Steps: 10
RTC Initialisation Processing time: 0.000000 s
Init Function Processing time: 0.000000 s
Processing Simulation Step 0
Step 0 Processing time: 0.001603 s
Processing Simulation Step 1
Step 1 Processing time: 0.000979 s
Processing Simulation Step 2
Step 2 Processing time: 0.000945 s
Processing Simulation Step 3
Step 3 Processing time: 0.001001 s
Processing Simulation Step 4
Step 4 Processing time: 0.000948 s
Processing Simulation Step 5
Step 5 Processing time: 0.000940 s
Processing Simulation Step 6
Step 6 Processing time: 0.000956 s
Processing Simulation Step 7
Step 7 Processing time: 0.000955 s
Processing Simulation Step 8
Step 8 Processing time: 0.000941 s
Processing Simulation Step 9
Step 9 Processing time: 0.000904 s
Exit Function Processing time: 0.000000 s
Total Processing time: 0.011201 s
Telemetry packet sent to 'https://nom.telemetrydeck.com/v1/' json was: [{\"isTestMode\":\"false\",\"appID\":\"94AC5E3F-F674-4E29-BF87-DAF4BA7F8F79\",\"clientUser\":\"FdhkyoR6s81wBsuf7vauwUVrTFYXN2pWymZa\",\"sessionID\":\"\",\"type\":\"simulation-run\",\"payload\":[\"GPUDevices:TeslaK80\",\"SimTime(s):0.011201\",\"Visualisation:false\",\"appVersion:2.0.0-rc.1\",\"appVersionFull:2.0.0-rc.1+a089dd1\",\"appVersionPatch:0\",\"appVersionPreRelease:rc.1\",\"buildNumber:a089dd1\",\"majorMinorSystemVersion:2.0.0\",\"majorSystemVersion:2\",\"operatingSystem:Linux\"]}]
```

However this is not actually sent the telemetry packet, as there are no matching signals in the `Recent Signals` tab of telemetry deck (There are some successfully K80 simualtion runs from today, but they don't match up with my runs (time or user identifier, or payload contents). 

`std::system`'s return value is implementation defined, but in this case it is returning `512`, so the `Telemetry::sendData` is returning false, but the message is still output due to the conditional logic on printing the warnings in `cudaEnsemble.cu` and `cudaSimulation.cu`. 

I.e. it should be `>=` for the failed to send warning, and it should only emit the telemetry packet info if it did not fail.

```c++
// send
        if (!flamegpu::io::Telemetry::sendData(telemetry_data)) {
            if ((getSimulationConfig().verbosity > Verbosity::Verbose))
                fprintf(stderr, "Warning: Usage statistics for CUDASimulation failed to send.\n");
        }
        // print
        if ((getSimulationConfig().verbosity >= Verbosity::Verbose)) {
            fprintf(stdout, "Telemetry packet sent to '%s' json was: %s\n", flamegpu::io::Telemetry::TELEMETRY_ENDPOINT, telemetry_data.c_str());
        }
```

Ideally we would still like this to actually send data, but even if we address the current error of `--connect-timeout` not liking decimal values, the version of curl present doesn't understand `--data-raw` either, which was added to curl in 7.43.0. `--data` is present, but this has special case handling of `@` unlike `--data-raw`, which we won't want to be included.

CURL 7.43 was relased in 2015. This is sufficinetly old in my opinion that we don't need to try and make this work, especially as curl 7.29.0 is known to have 72 CVEs, and 7.43.0 has 71.


--- 

I'll rebase this PR and patch the telemetry messaging logic, so that it stioll consumes the curl errors (added by this PR) but correctly outputs warnings / success messages.

ptheywood:
This now behaves as intended on HPC:
+ in verbose mode: emitting a (vague) warning if it fails to send, and not emitting the success message
+ Not in verbose mode: say nothing either way.

```
 ./bin/Release/boids_spatial3D -s 10 -t -vFLAME GPU 2.0.0-rc.1+58d29c6
Simulation configuration:
        Random Seed: 1681824248
        Steps: 10
RTC Initialisation Processing time: 0.000000 s
Init Function Processing time: 0.000000 s
Processing Simulation Step 0
Step 0 Processing time: 0.001325 s
Processing Simulation Step 1
Step 1 Processing time: 0.000823 s
Processing Simulation Step 2
Step 2 Processing time: 0.000852 s
Processing Simulation Step 3
Step 3 Processing time: 0.000861 s
Processing Simulation Step 4
Step 4 Processing time: 0.000853 s
Processing Simulation Step 5
Step 5 Processing time: 0.000834 s
Processing Simulation Step 6
Step 6 Processing time: 0.000809 s
Processing Simulation Step 7
Step 7 Processing time: 0.000857 s
Processing Simulation Step 8
Step 8 Processing time: 0.000856 s
Processing Simulation Step 9
Step 9 Processing time: 0.000841 s
Exit Function Processing time: 0.000000 s
Total Processing time: 0.009946 s
Warning: Usage statistics for CUDASimulation failed to send.
```

Also tested locally, and it correctly sends the packet and emits a success message if verbose if CURL is working. 

Also fixed a bug in test suite telemetry, where the outcome was inverted (due to return code 0 being implicitly falsey)

ptheywood:
Manually squashed / tidied the history. Good to go, subject to CI (and maybe a windows run, but should be fine).

mondus (approved):
Trusting Pete to approve this.

Issue #1035, opened 2023-01-25
Telemetry Payload errors


ptheywood:
The telemertry payload has a number of (potential errors) from a quick look at some data:

+ [ ] pyflamegpu is never reported - this is (probably) due to the `#ifdef SWIG` being in a source file not a header file, and SWIG is only defined when building the pyflamegpu interface. 
+ [x] `buildNumber` has been reported as `buildNumber` rather than an actual hash. This value is set at CMake configuration time from a call to git, which should be on the path at configure time otherwise an error should occur so not sure how this is happening.
  + This appears to be an issue with the manylinux CI jobs not getting the build number at cmake time, so manylinux workflows need adjusting to fix the git related warning.
  + Fixed by #1036

Robadob:
To catch python, the most sensible option is probably to augment the constructor for `ModelDescription` or `CUDASimulation`/`CUDAEnsemble`, to catch SWIG (pass a bool to the normal constructor defined private). Lazy version would be to detect use of RTC.

Robadob:
Marked this planned as it would be useful to identify C/Python breakdown if we're going to be publishing Python specific paper.

Robadob:
Bit more testing this morning/yesterday, can't do the same trick as the visualiser for `CUDASimulation::simulate()` as it's a virtual function it cannot be inlined. Needs some more thought.

PR #1036, opened 2023-01-25
BugFix: ManyLinux CI was not setting BuildNumber inside generated bin

Robadob:
Fix the build number issue inside ManyLinux CI.

Also pre-emptively apply the fix to most of the other CI jobs (those which do compilation/CMake test)

Relates #1035

Error:
image

Fixed version: (albeit the `FLAMEGPU_VERSION_BUILDMETADATA:` statement has been removed that was for debugging.)

image
image

### DIFF
--- a/.github/workflows/CMake.yml
+++ b/.github/workflows/CMake.yml
@@ -142,6 +142,10 @@ jobs:
         make
         sudo make install
 
+    # This pre-emptively patches a bug from ManyLinux where git dir is owned by diff user, blocking buildnumber generation
+    - name: Enable git safe-directory
+      run: git config --global --add safe.directory $GITHUB_WORKSPACE
+
     - name: Configure cmake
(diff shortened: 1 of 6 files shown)


ptheywood (approved):
Looks good. We should be able to double check this when the CI has run as the artifacts should be available for download, your testing did show it would fix it so it should be fine.

ptheywood:
Tested the manylinux wheel generated by the pull_request triggered draft_ci workflow:

```bash
python3 -m pip install pyflamegpu-2.0.0rc1+cuda120-cp310-cp310-linux_x86_64.whl 
python3 -c "import pyflamegpu; print(pyflamegpu.VERSION_BUILDMETADATA)"
06b722f
```

The hash is the post-merge hash due to the pull_request triggger, but actual releases will behave as intended.

PR #1041, opened 2023-02-19
Visualiser: Allow agent states to be hidden.

Robadob:
This solution is confined to the fgpu2 main library, the visualiser is not aware of the existence of hidden agent states.

Based on Jack A's suggestion. May want to add a note in the docs too.


ptheywood (approved):
Untested but looks fine. 

Does highlight we have a lack of tests for the vis API though.

Robadob:
To be merged once CI passes.

PR #1046, opened 2023-03-11
BugFix: ModelVis::setBeginPaused() now ensures pause is on step 0.

Robadob:
This PR will need updating to match the merged vis PR hash.

### DIFF
--- a/cmake/dependencies/flamegpu2-visualiser.cmake
+++ b/cmake/dependencies/flamegpu2-visualiser.cmake
@@ -7,7 +7,7 @@ include(FetchContent)
 cmake_policy(SET CMP0079 NEW)
 
 # Set the visualiser repo and tag to use unless overridden by the user.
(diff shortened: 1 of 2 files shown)


ptheywood (approved):
`game_of_life` was paused on step 0 10/10 times, compared to 7/10 times previously under linux.

Fine to merge once vis hash is updated (does feel it might be easier to just merge the repos given how many changes require updates to both).

Robadob:
> does feel it might be easier to just merge the repos given how many changes require updates to both

I understand your point, but I don't mind the additional hassle because I like the clear separation. Ideal world the visualiser API would reach a stable state and no further changes would be required. I'm not sure if we could move the separation so more of the Visualisation Config was within the visualiser too and whether that would reduce API changes.

Other option is to build a dynamic lib and runtime load it, that way you can add soft failure if certain features are unsupported (similar to how OpenGL and GL Extension Wrangler work).

I've now updated hash, will merge when CI is happy.

PR #1049, opened 2023-03-17
Add declare/define versions of agent/host function shims

Robadob:
This makes it easier to split models across multiple compilation units.

Useful to Fujitsu model, I manually did host function declarations in Primage too but never required agent function declarations until now.

This is a relatively advanced feature (and we don't have any documentation on splitting large models), so I feel it requires documentation.

I build and ran the test suite for sanity reasons, so it should be fine.

### DIFF
--- a/include/flamegpu/runtime/AgentFunctionCondition_shim.cuh
+++ b/include/flamegpu/runtime/AgentFunctionCondition_shim.cuh
@@ -24,7 +24,6 @@ namespace flamegpu {
  * SomeAgentFunctionCondition_cdn_impl SomeAgentFunctionCondition;
  * @endcode
  */
-
 #define FLAMEGPU_AGENT_FUNCTION_CONDITION(funcName)\
 struct funcName ## _cdn_impl {\
     __device__ __forceinline__ bool operator()(flamegpu::ReadOnlyDeviceAPI *FLAMEGPU) const;\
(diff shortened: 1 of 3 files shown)


ptheywood (approved):
Changes look fine, current test suite all passes under linux. 

I've not manually tried using them separately across compilation units however, but don't see why it wouldn't work.   

I aggree about it worth being documenting.

Robadob:
> I aggree about it worth being documenting.

I typo'd, OP was supposed to say don't feel worth documenting.

Robadob:
Have now tested it, doesn't device link unless `__forceinline__` is removed from them.

So two choices:
* Check the performance implications of removing `__forceinline__` from agent function implementations.
* Use redundancy, and only remove it from the DECL/DEF versions.

ptheywood:
Makes sense that `__forceinline__` doesn't link if its' effectivley being multiply defined.

If removing it does cost perf, LTO might be a way to reclaim the lost performance, would only know via experimentation.

Duplication seems like the "safer" choice / makes it opt-in to the (potentially slower) non inlined version(s)

ptheywood (approved):
Looks fine and test suite is still happy, though the new definitions are not present in the test suite so haven't actually tested them myself. 

Could probably do with being put into a "we should add tests for these at some point" issue.

Robadob:
> Could probably do with being put into a "we should add tests for these at some point" issue.

Will do, I've tested them locally with fujitsu.

PR #1052, opened 2023-03-25
Visualiser: Begin Paused v2

Robadob:
Tweaks visualiser begin paused behaviour, so that it begins paused on the first step with agents.

The changes so far, cause it to pause the step after agents are created (if created in a step function), need to check that agent creation can be moved so this is affected earlier.

Closes https://github.com/FLAMEGPU/FLAMEGPU2-visualiser/issues/122

### DIFF
--- a/include/flamegpu/visualiser/ModelVis.h
+++ b/include/flamegpu/visualiser/ModelVis.h
@@ -63,6 +63,10 @@ struct ModelVisData {
      */
     bool env_registered = false;
     /**
+     * Marked true when first agent data has been passed to vis
+     */
+    bool has_first_agents = true;
+    /**
      * Updates all agent renders from corresponding
      * @param sc Step count, the step count value shown in visualisation HUD
      */
--- a/src/flamegpu/visualiser/ModelVis.cpp
+++ b/src/flamegpu/visualiser/ModelVis.cpp
@@ -34,12 +34,16 @@ void ModelVisData::registerEnvProperties() {
 }
 void ModelVisData::updateBuffers(const unsigned int& sc) {
     if (visualiser) {
-        bool has_agents = false;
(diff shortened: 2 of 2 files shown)


ptheywood:
From that commit, looks like that change was to maintain the same value as before, while moving the `step_count` incrementing (deleted from line 73), so that the `step_count` is the last thing updated within the `step` method, making the the value is "correct" within an exit condtion.

This shouldn't have changed the value output to during the vis at all. It just means the timestep shown in the UI is the time after the current step, not the start of the current step. Removing the +1 would probably work, but then it would show 0 before step 0 has executed (if this would ever actually be visible) and at the end of step 0.

ptheywood (changes requested):
This does cause the simulation to pause reliably at insertion time (with the count in the bottom right reading +1 of the iteration used internally). 

It still isn't technically `beginPaused` though, so adding a docstring change to `setBeginPaused` noting this might be justified? 

I.e. this is an improvement on the current behaviour so worth merging, but it is arguably not a full fix for `beginPaused`.

Happy for this to be merged with a docstring change, but not close any issues for this (if it's still even open), and will still need the workaround of having atleast one agent being rendered from step 0 for true begin paused.

Robadob:
I honestly don't see the value in pausing before agents. But sure, will get around to it when I have time, need to remind myself of all these changes.

ptheywood:
Yes, grand scheme it's unintersting and I'm being pedantic, but for e.g. recording a video of a rate-limtied simulation it might be of interest.

I don't think it's an urgent need, as there's a known workaround, so just a docblock change so the method is clearly described as pausing on the first iterations agnets are present would be fine for the immediate future.

Unsure if removing the two `+1` being passed to the vis would cause other issues, or if it would just change the number shown.

PR #1055, opened 2023-03-27
BugFix: Message buildIndex() before buffer init.

### LINKED ISSUE #1054: MsgBucket, Optional output before buffer allocated exception.
Exception thrown here, because `getReadPtr()` reported the buffer hadn't been allocated yet.

https://github.com/FLAMEGPU/FLAMEGPU2/blob/c5007d284da9d8efbf0e937bb02034f571a7b0ef/src/flamegpu/runtime/messaging/MessageBucket.cu#L112

This fixes the error, but will need to write a proper test both for valid PBM after zero message and zero message first output.

```c++
    if (!MESSAGE_COUNT) {
        gpuErrchk(cudaMemsetAsync(hd_data.PBM, 0x00000000, (bucketCount + 1) * sizeof(unsigned int), stream));
        gpuErrchk(cudaStreamSynchronize(stream));
        return;
    }
```

I expect this may affect spatial messages too as they have similar code.

Robadob:
I don't think it impacts array messages (they have fixed allocation) or brute force (doesn't use `buildIndex()`).

I added tests because it's quite the edge-case to hit this bug.

Closes #1054


ptheywood (approved):
New (and old) tests pass under linux.

PR #1056, opened 2023-03-28
CUDA 12 stream destruction fix for multi-gpu systems

### LINKED ISSUE #1053: `cleanup` issue in multi-gpu environment with CUDA 12.x
The existing naive `flamegpu::cleanup` method causes cuda errors during stream destruction when using CUDA 12.x in multi-gpu systems. 

The cleanup method is currently quite naive, and resets all devices present in the system, but does not re-select the original device which may cause issues. This is trivial to add but does not resolve this issue. 

A cudaDriver API method is used to guard stream destruction, based on if the primary context for the device is `active`. Previsouly this was sufficient to prevent destruction of a stream when the context was no longer valid. 

There is a cudaDriver API method `cuStreamGetCtx` which returns a pointer to the cudaContext which the stream belongs to. If the CUDASimulation object remembered the cudaContext which it belongs to, we could check this is still correct (and is also still the active primary context?) and check the device is correct and only destroy the streams in that case (which should resolve this issue). 

This will (may) need to copy the context object rather than just the pointer to the context, incase the context pointer is reused. 

+ [x] Cleanup device resumption
+ [x] Adjust stream destruction context guarding to be cuda 12.x compatible

ptheywood:
When using CUDA 12.x in a multi-gpu system, an interaction between `flamegpu::cleanup()` and `cudaDestroyStreams` during `~CUDASimulation()` was casuing exceptions to be thrown.

+ CUDA streams are only valid in the context which was used to create them
+ CUDA runtime and driver API methods to query a streams context / status all include UB when passing an invalid stream handle, leading to segfaults
+ A CUDA 11.x detection for the context no longer being valid by checking the primary context with the driver API does not successfully prevent the UB being triggered in CUDA 12.x in a multi-gpu system.
+ `CUcontext` can/will return the same handle after a reset, so we cannot "remember" the context that way, but CUDA 12 introduces a method which allows us to compare a unique `unsigned long long int` which can be used to check the context is a match.
+ Also ensures cleanup re-selects the original device at the time of the call, just in case (even though any existing contexts would no longer be valid). 

No new tests are required, the existing cleanup tests triggered the issue with multi-gpu + CUDA 12.

Closes #1053 

Tested in a multi-gpu linux system with CUDA 11.0, 11.8 and 12.0 with different `CUDA_VISIBLE_DEVICES` values and on a single GPU system with CUDA 12.0.

### DIFF
--- a/include/flamegpu/detail/cuda.cuh
+++ b/include/flamegpu/detail/cuda.cuh
@@ -8,6 +8,7 @@
  */
 #include <cuda_runtime.h>
 #include <cuda.h>
+#include <limits>
 #include "flamegpu/exception/FLAMEGPUException.h"
 
 namespace flamegpu {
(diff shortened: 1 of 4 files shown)


Robadob (approved):
Changes look reasonable.

Can test on Windows on my office machine if you think necessary (and specify what the Repro is).

Robadob:
Yeah my office machine has the 750ti and titanx on Windows, hence my suggestion of that if we care.

PR #1060, opened 2023-04-05
Add Poisson distribution support to Host/Agent random API.

Robadob:
Note: Only the least robust curand version is supported (others require extra data/host operations).

Max said this would be useful.

Docs needs updating too:
* https://docs.flamegpu.com/guide/agent-functions/random-numbers.html
* https://docs.flamegpu.com/guide/host-functions/random-numbers.html


ptheywood (approved):
Looks senssible, and builds under linux but it is not used anywhere so I don't know if it actually runs or not. 

Would be nice to have atleast one test which calls this and checks it doesn't throw / produce nonsense values (seems to be a well defined what values can be returned by a given mean).

PR #1062, opened 2023-04-21
CMake Fixes

### LINKED ISSUE #1045: CMake: make `lint_flamegpu` optional / not part of `all_lint` when included
When creating a flame gpu executable in isolation, the `all_lint` target includes `lint_flamegpu`, which greatly increases lint time, for files which users will not be interested in linting (99% of the time).

We should make the creation of the `lint_flamegpu` target conditional within `src/CMakeLists.txt`. 

Probably via a CMake option `FLAMEGPU_ENABLE_LINT_FLAMEGPU` or similar, which defaults to `OFF` unless either `CMakeLists.txt` or `src/CMakeLists.txt` are the top level CMakeLists.txt? In general though I'm not sure we even support `src/CMakeLists.txt` being the top level CMakeLists.txt, so it might be an easier fix.

Downstream users could then opt in to this linting, rather than it being default (and not possible to opt out from). 

Alterantively, we could just make it not part of `all_lint`, but I'd lean towards it not being a target.

---

It might also be nice to make the name of tha `all_lint` target configurable, i.e. so it can just be called `lint` (without a duplicate target).

### LINKED ISSUE #1051: CMake `flamegpu_add_executable` `source_group` errors
When using `flamegpu_add_executable`, the use of `source_group` to assign visual studio filters can cause errors when:

+ `${CMAKE_CURRENT_SOURCE_DIR}/src` is not a folder
+ source files are not in or below `${CMAKE_CURRENT_SOURCE_DIR}`, which is passed to the `TREE` argument of `source_group`. 

This should be made more flexible / robust. 

https://cmake.org/cmake/help/latest/command/source_group.html

in `src/CMakeLists.txt`, it uses the `FLAMEGPU_ROOT` relative paths to source group, but that will not work for downstream projects. 

Simplest "fix" would be to allow this feature to be disabled from `flamegpu_add_executable` via an argument, allowing users to create their own filters with known paths. We may be able to first check all passed source files against `CMAKE_CURRENT_SOURCE_DIR` and disable the feature at that point, emitting a warning to the user to pass the new argument, and suggest they add their own filters?

There might be a more complex solution allowing users to provide their own source groups, but disablign the autmatic source groups would be flexible enough.

### LINKED ISSUE #1057: CMake: flamegpu_new_linter_target uses PROJECT_NAME not NAME
The custom CMake function `flamegpu_new_linter_target` uses a project scoped variable, not the function scoped `name` variable, preventing the definition of 2 flamegpu executables within one CMake project.

This is effectively a copy-paste bug in `cpplint.cmake`

I'll fix this when I have time to check for other occurances too.

ptheywood:
A number of fixes and additions to FLAMEGPU2's CMake encountered when working on a more complex tempalte model.

+ [x] `flamegpu_new_linter_target` fix. Closes #1057 
+ [x] Make the `lint_flamegpu` target OFF by default when this repo is not he primary CMake. Closes #1045 
+ [x] Make the `docs` target OFF by default when this repo is not primary
+ [x] Add `flamegpu_add_library`, for creating static libraries which are more than just linked against FLAMEGPU2 (i.e. warning levels MSVC + CUDA fixes etc. This allows the creation of unit-testable FLAMEGPU 2 simulators.
+ [x] Some Refactoring, to reduce duplication
+ [x] Demonstrate `flamegpu_add_library` in an existing example. This may not actually be desirable for merging.
+ [x] Fix `source_group` errors, Closes #1051. Tested on windows.
+ [x] Make sure that abs paths with `++` still work (I.e. check I don't break #1085 during a rebase)

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,6 +179,7 @@ cmake --build . --target all
 | `FLAMEGPU_SHARE_USAGE_STATISTICS`    | `ON`/`OFF`                  | Share usage statistics ([telemetry](https://docs.flamegpu.com/guide/telemetry)) to support evidencing usage/impact of the software. Default `ON`. |
 | `FLAMEGPU_TELEMETRY_SUPPRESS_NOTICE` | `ON`/`OFF`                  | Suppress notice encouraging telemetry to be enabled, which is emitted once per binary execution if telemetry is disabled. Defaults to `OFF`, or the value of a system environment variable of the same name. |
 | `FLAMEGPU_TELEMETRY_TEST_MODE`       | `ON`/`OFF`                  | Submit telemetry values to the test mode of TelemetryDeck. Intended for use during development of FLAMEGPU rather than use. Defaults to `OFF`, or the value of a system environment variable of the same name.|
+| `FLAMEGPU_ENABLE_LINT_FLAMEGPU`      | `ON`/`OFF`                  | Enable/Disable creation of the `lint_flamegpu` target. Default `ON` if this repository is the root CMAKE_SOURCE_DIR, otherwise `OFF` |
 
 <!-- Additional options which users can find if they need them.
 | `FLAMEGPU_BUILD_API_DOCUMENTATION` | `ON`/`OFF` | Build the documentation target. Default `ON` |
--- a/cmake/common.cmake
+++ b/cmake/common.cmake
@@ -172,9 +172,9 @@ function(flamegpu_common_compiler_settings)
 
     # Ensure that a target has been passed, and that it is a valid target.
     if(NOT CCS_TARGET)
-        message( FATAL_ERROR "flamegpu_common_compiler_settings: 'TARGET' argument required")
+        message( FATAL_ERROR "${CMAKE_CURRENT_FUNCTION}: 'TARGET' argument required")
     elseif(NOT TARGET ${CCS_TARGET} )
-        message( FATAL_ERROR "flamegpu_common_compiler_settings: TARGET '${CCS_TARGET}' is not a valid target")
+        message( FATAL_ERROR "${CMAKE_CURRENT_FUNCTION}: TARGET '${CCS_TARGET}' is not a valid target")
     endif()
 
     # Enable -lineinfo for Release builds, for improved profiling output.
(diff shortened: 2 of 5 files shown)


Robadob:
I've no strong preference, removing the redundant `src` seems fine.

ptheywood:
>  removing the redundant src seems fine.

For examples we can't really do that, as the src dir may not exist (which could cause errors / issues with the source grouping before) and would then potentially cause folder conflicts, if say the user had this (unwise) setup

```
├── a
│   └── b.h
└── src
    └── a
        └── b.h
```

Then intentionally removing src would cause both `a`'s to be the same filter, which contain files with the same name which I'm not sure how VS would handle. 

---

We probably could for the flamegpu target, but its probably simpler / more consistent to keep it.

I'll make the change when I'm back in windows at some point. 

---

The CI failure is just from my low-effort testing of the folder structures / CMake, so once I drop the appropraite commit it pass.

Robadob:
> > removing the redundant src seems fine.
> 
> For examples we can't really do that, as the src dir may not exist (which could cause errors / issues with the source grouping before) and would then potentially cause folder conflicts, if say the user had this (unwise) setup
> 
> ```
> ├── a
> │   └── b.h
> └── src
>     └── a
>         └── b.h
> ```
> 
> Then intentionally removing src would cause both `a`'s to be the same filter, which contain files with the same name which I'm not sure how VS would handle.
> 
> We probably could for the flamegpu target, but its probably simpler / more consistent to keep it.
> 
> I'll make the change when I'm back in windows at some point.
> 
> The CI failure is just from my low-effort testing of the folder structures / CMake, so once I drop the appropraite commit it pass.

As I said, no strong preference do what you think best. It's hard to go wrong.

ptheywood:
I've now made the source_group change to the FLAMEGPU target, so that it places things into the CMake default filters of `Header Files` and `Source Files`, but using tree within the `FLAMEGPU_ROOT` directory (must use `FLAMEGPU_ROOT` as `include` is above `src/CMakeLists.txt`). 

This makes it nest deeper, but is consistent.

image

I'm not strongly opinionated either way, so if you don't like this we can drop the final commit and still merge this anyway.

---

I've also removed the example(s) which showed the previous CMake failures, as they were messy and I'm not aware of a nice way to have a CMake test suite. I have pushed them to another branch [`cmake-fixes-rc1-with-example`](https://github.com/FLAMEGPU/FLAMEGPU2/tree/cmake-fixes-rc1-with-example) if you would like to test them.

Otherwise this PR is ready for review.

Robadob (changes requested):
> CMake: Demonstrate use of flamegpu_add_library on the host_fucntions example. Maybe don't merge?

Would require example readme discussing, or delete commit

ptheywood:
I've now dropped the missed "probably don't merge" commit, which was just a test case demonstrating the library use case rather than binary. I'll open an issue to document it on the docs repo and / or use it in a standalone example in the future too.

Robadob (approved):
LGTM

PR #1077, opened 2023-07-05
BugFix: PyNative would fail to translate math fns inside message loops.

Robadob:
This affected the pynative tutorial from the documentation.

The prior code

```py
            # message input iterator arg
            elif self._message_iterator_var:
                if t.value.id == self._message_iterator_var:
```
Would cause the branch to be taken if inside a message loop, but then if the message iterator var was not being used, it would break from that branch and fail to check the other conditions. Hence math and other functions inside message loops would not be translated.

Related: https://github.com/FLAMEGPU/FLAMEGPU2-docs/pull/150

### DIFF
--- a/swig/python/codegen/codegen.py
+++ b/swig/python/codegen/codegen.py
@@ -477,8 +477,7 @@ class CodeGenerator:
                     self.RaiseError(t, f"Message input variable '{self._input_message_var}' does not have a supported function '{t.attr}'") 
 
             # message input iterator arg
-            elif self._message_iterator_var:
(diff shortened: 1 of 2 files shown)


mondus:
Logic change for this looks correct. It is obviously why function calls inside loops where failing to proceed with parsing. Happy to merge this now (although would benefit from a test of expected message format inside a message loop).

mondus (approved):
Logic change for this looks correct. It is obviously why function calls inside loops where failing to proceed with parsing. Happy to merge this now (although would benefit from a test of expected message format inside a message loop).

Issue #998, opened 2022-11-24
Windows pyflamegpu build creates nested venvs (eventually leading to build failure)


Robadob:
Some recent change (probably past month) has led to repeated incremental (and re?) builds of the `pyflamegpu` target on windows, to cause a nested venv to occur (see the logs below). This grows, until the point the path is so deep long (in my case about 4 levels) that a file copy fails and the build breaks.

```
3>copying venv\Lib\site-packages\venv\Lib\site-packages\venv\Lib\site-packages\venv\Lib\site-packages\venv\Lib\site-packages\venv\Lib\site-packages\pip\_internal\utils\distutils_args.py -> build\lib.win-amd64-cpython-39\venv\Lib\site-packages\venv\Lib\site-packages\venv\Lib\site-packages\venv\Lib\site-packages\venv\Lib\site-packages\venv\Lib\site-packages\pip\_internal\utils
3>copying venv\Lib\site-packages\venv\Lib\site-packages\venv\Lib\site-packages\venv\Lib\site-packages\venv\Lib\site-packages\venv\Lib\site-packages\pip\_internal\utils\egg_link.py -> build\lib.win-amd64-cpython-39\venv\Lib\site-packages\venv\Lib\site-packages\venv\Lib\site-packages\venv\Lib\site-packages\venv\Lib\site-packages\venv\Lib\site-packages\pip\_internal\utils
3>copying venv\Lib\site-packages\venv\Lib\site-packages\venv\Lib\site-packages\venv\Lib\site-packages\venv\Lib\site-packages\venv\Lib\site-packages\pip\_internal\utils\encoding.py -> build\lib.win-amd64-cpython-39\venv\Lib\site-packages\venv\Lib\site-packages\venv\Lib\site-packages\venv\Lib\site-packages\venv\Lib\site-packages\venv\Lib\site-packages\pip\_internal\utils
3>copying venv\Lib\site-packages\venv\Lib\site-packages\venv\Lib\site-packages\venv\Lib\site-packages\venv\Lib\site-packages\venv\Lib\site-packages\pip\_internal\utils\entrypoints.py -> build\lib.win-amd64-cpython-39\venv\Lib\site-packages\venv\Lib\site-packages\venv\Lib\site-packages\venv\Lib\site-packages\venv\Lib\site-packages\venv\Lib\site-packages\pip\_internal\utils
[...]

PR #1083, opened 2023-07-06
Workaround: Pyflamegpu incremental builds would fail on Windows.

Robadob:
Build creates nested VMs with repeated builds, incrasing wheel size and leading to failure.

Deleting these two directories works around the problem as a temporary hack

Relates #998


Robadob:
Deprecate by the fix #1084

PR #1085, opened 2023-07-07
CMake: Escape user-provided regex patterns.

### LINKED ISSUE #1071: CMake: Configuration failure with `++` in abs path
When cpplint is found and enabled, configuring cmake within a directory containing `++` character in the path results in a regular expression compilation failure at:

https://github.com/FLAMEGPU/FLAMEGPU2/blob/86197b63efddfa561afd184785a1cc127dd6fb4d/cmake/cpplint.cmake#L52

We will need to somehow escape the offending `++` (and other patterns that woudl need escaping" from `${FLAMEGPU_ROOT}`. 

This would also need applying tyo any other CMake Regex use, which uses a path as the filter, or a user-supplied filter (if any)

Robadob:
Closes #1071 (See issue for fix commentary)

### DIFF
--- a/cmake/EscapeRegex.cmake
+++ b/cmake/EscapeRegex.cmake
@@ -1,0 +1,7 @@
+# Define a function for escaping regex chars from strings
+# This prevents regex failures when user provided paths contain regex special chars
+function(escape_regex IN_STRING OUT_VAR)
+    # https://gitlab.kitware.com/cmake/cmake/-/issues/18580#note_483128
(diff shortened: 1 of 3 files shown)


ptheywood (approved):
Fixes paths including `++` with linting enabled under linux :+1:

Might cause me some conflicts with #1062, but that's a problem for future me (and I just need to make sure I don't undo this fix when that gets merged, hence I'm tagging that PR here)

PR #1087, opened 2023-07-07
Environment Macro Property IO

### LINKED ISSUE #1070: Environment Macro Property IO
It's not currently possible for environment macro properties to be imported/exported via xml or json.

This will need to be host API functionality, as it's currently expected that they zero init.

https://github.com/FLAMEGPU/FLAMEGPU2/discussions/1069

Robadob:
Ended up refactoring much of the internal model state IO interface to make partial import/export possible.

- [x] BugFix: `JSONStateWriter` would create an agent object for each agent state, leading to multiple members with the same key.
- [x] Bugfix: `HostMacroProperty` did not use streams to copy data. (C tests still pass after this change)
- [x] Switch `CUDAMacroEnvironment` from reference to `shared_ptr`
* State
  * Export
    - [x] JSON
    - [x] XML
  * Import
    - [x] JSON
    - [x] XML
  - [x] tests
* HostAPI
  * Export
    - [x] JSON
    - [x] XML
    - [x] binary
  * Import
    - [x] JSON
    - [x] XML
    - [x] binary
  - [x] tests
- [x] Docs update PR (https://github.com/FLAMEGPU/FLAMEGPU2-docs/pull/152)

**Of note, macro properties can still not be initialised via `EnvironmentDescription` or `RunPlan`.**

It's a basic interface change (two new methods added to `HostAPI` with no weird types/templates, so only C API tests have been extended/updated.

Commits are split cleanly, does **not** require squash.

Closes #1070

### DIFF
--- a/include/flamegpu/runtime/environment/HostMacroProperty.cuh
+++ b/include/flamegpu/runtime/environment/HostMacroProperty.cuh
@@ -33,14 +33,26 @@ struct HostMacroProperty_MetaData {
     }
     /**
      * Download data
+     * @note Only works first time
+     * @see force_download()
      */
     void download() {
         if (!h_base_ptr) {
-            h_base_ptr = static_cast<char*>(malloc(elements * type_size));
-            gpuErrchk(cudaMemcpyAsync(h_base_ptr, d_base_ptr, elements * type_size, cudaMemcpyDeviceToHost, stream));
-            gpuErrchk(cudaStreamSynchronize(stream));
-            has_changed = false;
+            force_download();
         }
+    }
+
+    /**
+     * Downloads data from device to host (allocating host buffer if required)
+     * Sets has_changed flag to false, as host is current to device
+     */
+    void force_download() {
+        if (!h_base_ptr) {
+            h_base_ptr = static_cast<char*>(malloc(elements * type_size));
+        }
+        gpuErrchk(cudaMemcpyAsync(h_base_ptr, d_base_ptr, elements * type_size, cudaMemcpyDeviceToHost, stream));
+        gpuErrchk(cudaStreamSynchronize(stream));
+        has_changed = false;
     }
     /**
      * Upload data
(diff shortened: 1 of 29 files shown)


include/flamegpu/runtime/environment/HostMacroProperty.cuh
```diff
@@ -32,13 +33,22 @@ struct HostMacroProperty_MetaData {
...
      */
     void download() {
         if (!h_base_ptr) {
             h_base_ptr = static_cast<char*>(malloc(elements * type_size));
-            gpuErrchk(cudaMemcpy(h_base_ptr, d_base_ptr, elements * type_size, cudaMemcpyDeviceToHost));
-            has_changed = false;
+            force_download();
+        }
+    }
+    void force_download() {
```
ptheywood, line 45: Missing docstring, so the `@see` in download's docstring isn't mega useful.

src/flamegpu/io/XMLStateWriter.cu
```diff
@@ -0,0 +1,435 @@
+
+/**
+ * @file
+ * @author
+ * @date
+ * @brief
+ *
+ * \todo longer description
+ */
```
ptheywood, line 9: This seems pointless? I assume it was in the .cpp version of this file and its just been retained (git's decided they are different enough to not show a useful diff).

ptheywood (changes requested):
A few trivial comments added, that would be nice (and simple) to address, though not the end of the world to merge without. Can wait till you return from leave.

The change to .cu files made diffing less clear (just git issues), and will mean we have to do more of a breaking change later if we have a non cuda build, but that's going to be breaking either way. 

Tests all pass under linux. 

This is an API break, so will need documenting in the changelog as such for the next pre-release.

src/flamegpu/io/XMLStateWriter.cu
Robadob: yes

ptheywood (approved):

PR #1089, opened 2023-07-14
Network message v2

Robadob:
Copied from https://github.com/FLAMEGPU/FLAMEGPU2/pull/883, working in a new branch to avoid breaking FRE project's dependency during development.

Remaining tasks:

- [x] Update current ID code to not enforce contiguous IDs
- [x] Validate IDs
- [x] Fix traversal
  - [x] Validate early, translate at the end
  - [x] Update pbm/ipbm to construct keys/vals with idmap
   - [x] fix tests
- [x] Add device API edge src-dest to index method
- [x] Duplicate host API to offer both index and ID (This will necessitate a host ID map)
- [x] (more) tests
- [x] python/rtc API tests (particularly around no edges)
- [x] native python/codegen tests (api has change a touch since they were setup)
- [x] [Improve `HostAPI` auto sync](https://github.com/FLAMEGPU/FLAMEGPU2/pull/883#issuecomment-1410080268)
- [x] docs (https://github.com/FLAMEGPU/FLAMEGPU2-docs/pull/167)


Robadob:
I've checked that the "Hello World" from this tutorial works: https://mpitutorial.com/tutorials/mpi-hello-world/

Will play with this a bit, to understand it, before trying to update `CUDAEnsemble`.

Robadob:
## Todo

Agree whether vertex IDs can be "sparse" (@ptheywood in favour because everyone else does it, me not in favour because it adds an additional indirection (if not more) when accessing the graph via ID)

Robadob:
Sparse IDs, limited by memory. Users encouraged to keep them contiguous to avoid memory issues.

include/flamegpu/runtime/environment/DeviceEnvironmentDirectedGraph.cuh
```diff
@@ -527,6 +528,17 @@ __device__ __forceinline__ id_t DeviceEnvironmentDirectedGraph::getVertexID(cons
 #endif
     return getVertexProperty<id_t>("_id", vertex_index);
 }
+
+__device__ __forceinline__ unsigned int DeviceEnvironmentDirectedGraph::getVertexIndex(id_t vertex_id) const {
+#if !defined(FLAMEGPU_SEATBELTS) || FLAMEGPU_SEATBELTS
+    const unsigned int VERTEX_COUNT = detail::curve::DeviceCurve::getVariableCount("_index_map", graph_hash ^ detail::curve::Curve::variableHash("_environment_directed_graph_vertex"));
```
Robadob, line 534: As discussed in meeting, this limit will nolonger be enforced. Instead IDs will have a separate max.

src/flamegpu/simulation/detail/CUDAEnvironmentDirectedGraphBuffers.cu
```diff
@@ -82,6 +82,7 @@ void CUDAEnvironmentDirectedGraphBuffers::allocateVertexBuffers(const size_type
     gpuErrchk(cudaMalloc(&d_pbm, sizeof(unsigned int) * (count + 1)));
     gpuErrchk(cudaMalloc(&d_pbm_swap, sizeof(unsigned int) * (count + 1)));
     gpuErrchk(cudaMalloc(&d_ipbm, sizeof(unsigned int)* (count + 1)));
+    gpuErrchk(cudaMalloc(&d_vertex_index_map, sizeof(unsigned int) * count));
```
Robadob: count will be a different var.

src/flamegpu/simulation/detail/CUDAEnvironmentDirectedGraphBuffers.cu
```diff
@@ -290,6 +295,12 @@ struct CustomMin {
         return (b < a) ? b : a;
     }
 };
+__global__ void buildIDMap(const unsigned int *IDsIn, unsigned int *indexOut, const unsigned int count) {
```
Robadob: this will need a rethink

include/flamegpu/exception/FLAMEGPUException.h
```diff
@@ -420,6 +420,8 @@ DERIVED_FLAMEGPUException(InvalidDependencyGraph, "Agent function dependency gra
  * This should not occur if the shared ID matches ID_NOT_SET
  */
 DERIVED_FLAMEGPUException(AgentIDCollision, "Multiple agents of same type share an ID");
+DERIVED_FLAMEGPUException(IDCollision, "Generic ID collision");
+DERIVED_FLAMEGPUException(IDOutOfBounds, "Generic ID range failure");
```
Robadob, line 424: ID range will now be reported as CUDA error (though we could catch and rethrow).

Robadob:
Hard stuff is all done (unless i find more edge cases along the way).

I've agreed with @ptheywood that we should offer both index and ID access API, and just document index as being preferable for performance.

**Remaining tasks:**
* device/host id&index api duplication+ tests
* rtc device api tests
* docs

Should be achievable by the end of next week.

Robadob:
Current discussion is for `HostAPI` to emulate C++ `std::map`. Albeit with a fixed size

`HostStaticVertexMap HostEnvironmentDirectedGraph::vertices()`

User first must resize it with existing `HostEnvironmentDirectedGraph::setVertexCount()`.

They can then insert and update elements, using `operator[]`. Similar to a regular `std::map`, or use iteration mechanics.

`insert()`, `erase()`, `emplace()` would not be supported to avoid the need to define a host vertex structure (which would essentially be a wrapper for `std::map<std::string, std::any>` 🤮).

Robadob:
HostAPI refactor should be complete, with all tests working on C++/Py.

Going to make a start on the documentation.

Robadob:
Final changes to do next week are adding python tests around device API edge cases and checking whether there needs to be some IO documentation added.

include/flamegpu/runtime/detail/SharedBlock.h
```diff
@@ -19,8 +19,8 @@ struct SharedBlock {
 #if !defined(FLAMEGPU_SEATBELTS) || FLAMEGPU_SEATBELTS
     unsigned int curve_type_size[curve::Curve::MAX_VARIABLES];
     unsigned int curve_elements[curve::Curve::MAX_VARIABLES];
-    unsigned int curve_count[curve::Curve::MAX_VARIABLES];
 #endif
+    unsigned int curve_count[curve::Curve::MAX_VARIABLES];
```
ptheywood, line 23: This feels like a bug that was fixed to do with seatbelts elsewhere? Not sure if this is the correct value, mid review. Should double check this works with and without seatbelts just incase.

(this is just a note mid-review so feel free to ignor for now, but doubt I'll get through all of this in one go).

include/flamegpu/runtime/detail/curve/DeviceCurve.cuh
```diff
@@ -231,6 +233,69 @@ class DeviceCurve {
...
 
+
+    /**
+     * Retrieve the address of the specified environment directed graph's pbm (CSR) for edges outgoing from each vertex
+     *
+     * @param graphHash The hash of the affected graph
+     * @return The requested address
+     * @throws exception::DeviceError (Only when SEATBELTS==ON) If the specified graph is not found in the cuRVE hashtable, or it's details are invalid.
+     */
+    __device__ __forceinline__ static unsigned int *getEnvironmentDirectedGraphPBM(VariableHash graphHash);
```
ptheywood, line 244: I'd prefer CSR/CSC in the names not PBM / IPBM, but prolly not worth the effort to fix / just my bias from doing more graph stuff.

include/flamegpu/runtime/environment/DeviceEnvironmentDirectedGraph.cuh
```diff
@@ -0,0 +1,653 @@
...
+             */
+            const OutEdgeFilter&_parent;
+            /**
+             * This is the index of the currently accessed edge, relative to the full edge list
+             */
+            unsigned int edge_index;
+
+         public:
+            /**
+             * Constructs a message and directly initialises all of it's member variables
```
ptheywood, line 47:
```suggestion
             * Constructs a edge and directly initialises all of it's member variables
```

include/flamegpu/runtime/environment/DeviceEnvironmentDirectedGraph.cuh
```diff
@@ -0,0 +1,653 @@
...
+                return this->edge_index == rhs.edge_index;
+            }
+            /**
+             * Inequality operator
+             * Returns inverse of equality operator
+             * @see operator==(const Message&)
+             */
+            __device__ bool operator!=(const Edge& rhs) const { return !(*this == rhs); }
+            /**
+             * Updates the message to return variables from the next message in the message list
```
ptheywood, line 68:
```suggestion
             * Updates the edge to return variables from the next message in the message list
```

include/flamegpu/runtime/environment/DeviceEnvironmentDirectedGraph.cuh
```diff
@@ -0,0 +1,653 @@
...
+             */
+            unsigned int ipbm_index;
+            /**
+             * This is the index of the currently accessed edge, relative to the full edge list
+             */
+            unsigned int edge_index;
+
+         public:
+            /**
+             * Constructs a message and directly initialises all of it's member variables
```
ptheywood, line 233:
```suggestion
             * Constructs a edge and directly initialises all of it's member variables
```

include/flamegpu/runtime/environment/DeviceEnvironmentDirectedGraph.cuh
```diff
@@ -0,0 +1,653 @@
...
+                return this->ipbm_index == rhs.ipbm_index;
+            }
+            /**
+             * Inequality operator
+             * Returns inverse of equality operator
+             * @see operator==(const Message&)
+             */
+            __device__ bool operator!=(const Edge& rhs) const { return !(*this == rhs); }
+            /**
+             * Updates the message to return variables from the next edge in the message list
```
ptheywood, line 256:
```suggestion
             * Updates the edge to return variables from the next edge in the list of incoming edges
```

include/flamegpu/runtime/environment/DeviceEnvironmentDirectedGraph.cuh
```diff
@@ -0,0 +1,653 @@
...
+/**
+ * Read-only accessor to a specific directed graph
+ * Directed graphs can only be modified via host functions
+ */
+class DeviceEnvironmentDirectedGraph {
+    const detail::curve::Curve::VariableHash graph_hash;
+
+ public:
+    /**
+    * This class is created when a vertex is provided to DeviceEnvironmentDirectedGraph::edgesLeaving()(id_t)
```
ptheywood, line 17:
```suggestion
    * This class is created when a vertex is provided to DeviceEnvironmentDirectedGraph::outEdges()(id_t)
```

Think this is the correct mapping? There are other occurences too that I might not have caught all of them in comments / docstrings.

include/flamegpu/runtime/environment/DeviceEnvironmentDirectedGraph.cuh
```diff
@@ -0,0 +1,653 @@
...
+            /**
+             * The edge returned to the user
+             */
+            Edge _edge;
+
+         public:
+            /**
+             * Constructor
+             * This iterator is constructed by DeviceEnvironmentDirectedGraph::OutEdgeFilter::begin()(id_t)
+             * @see DeviceEnvironmentDirectedGraph::edgesLeaving()(id_t)
```
ptheywood, line 118:
```suggestion
             * @see DeviceEnvironmentDirectedGraph::outEdges()(id_t)
```

include/flamegpu/runtime/environment/DeviceEnvironmentDirectedGraph.cuh
```diff
@@ -0,0 +1,653 @@
...
+         * Search bucket bounds
+         */
+        unsigned int bucket_begin, bucket_end;
+        /**
+         * Graph hash for accessing graph data
+         */
+        const detail::curve::Curve::VariableHash graph_hash;
+    };
+    /**
+     * This class is created when a vertex is provided to DeviceEnvironmentDirectedGraph::edgesLeaving()(id_t)
```
ptheywood, line 198:
```suggestion
     * This class is created when a vertex is provided to DeviceEnvironmentDirectedGraph::inEdges()(id_t)
```

include/flamegpu/runtime/environment/DeviceEnvironmentDirectedGraph.cuh
```diff
@@ -0,0 +1,653 @@
...
+            /**
+             * The edge returned to the user
+             */
+            Edge _edge;
+
+         public:
+            /**
+             * Constructor
+             * This iterator is constructed by DeviceEnvironmentDirectedGraph::InEdgeFilter::begin()(id_t)
+             * @see DeviceEnvironmentDirectedGraph::edgesJoining()(id_t)
```
ptheywood, line 309: Same applies for edges joining as edges leaving, a few occurences in comments in dif files that I may not mark all of, so worth a search. 

```suggestion
             * @see DeviceEnvironmentDirectedGraph::inEdges()(id_t)
```

include/flamegpu/runtime/environment/DeviceEnvironmentDirectedGraph.cuh
```diff
@@ -0,0 +1,653 @@
...
+        };
+        /**
+         * Constructor, takes the search parameters required
+         * Begin key and end key specify the [begin, end) contiguous range of bucket. (inclusive begin, exclusive end)
+         * @param _graph_hash Graph hash for accessing data via curve
+         * @param vertex_index The index of the vertex to retrieve leaving edges for
+        */
+        inline __device__ InEdgeFilter(detail::curve::Curve::VariableHash _graph_hash, const id_t &vertex_index);
+        /**
+       * Returns an iterator to the start of the message list subset about the search origin
```
ptheywood, line 357:
```suggestion
       * Returns an iterator to the start of the in edge list subset leaving the vertex
```

include/flamegpu/runtime/environment/HostEnvironmentDirectedGraph.cuh
```diff
@@ -0,0 +1,440 @@
...
+    size_type getEdgeCount();
+    /**
+     * Returns a map for accessing vertices
+     * The vertices are accessed via their ID, and will be created if they don't already exist
+     * It is necessary to first call setVertexCount() to initialise the storage for vertices
+     */
+    VertexMap vertices();
+    /**
+     * Returns a map for accessing edges
+     * The edges are accessed via their ID, and will be created if they don't already exist
```
ptheywood, line 94: Edges don't have an id? they have a vertex id pair?

include/flamegpu/util/StringPair.h
```diff
@@ -14,26 +14,44 @@ namespace util {
...
 
 /**
- * Hash function so that StringPair can be used as a key in a map
+ * Compare function so that StringPair can be used in a map
  */
-struct StringPairHash {
-    size_t operator()(const std::pair<std::string, std::string>& k) const {
-        return std::hash<std::string>()(k.first) ^
-            (std::hash<std::string>()(k.second) << 1);
+template<typename T1, typename T2>
```
ptheywood, line 19: Techinically these changes might be an api break, even though its a bit no one is intended to directly use. 

Might be worth mentioning but not in much detail in a cahngelog?

src/CMakeLists.txt
```diff
@@ -100,6 +100,10 @@ unset(FLAMEGPU_ENABLE_LINT_FLAMEGPU_DEFAULT)
 option(FLAMEGPU_ENABLE_GLM "Experimental: Make GLM available to flamegpu2 projects on the include path" OFF)
 mark_as_advanced(FLAMEGPU_ENABLE_GLM)
 
+# Option to make put glm on the include path
```
ptheywood, line 103:
```suggestion
# Option to enable the "advanced API" - access to high performance internals that may break at any time 
```

src/flamegpu/io/JSONGraphReader.cpp
```diff
@@ -0,0 +1,383 @@
...
+            mode.push(Nop);
+        }
+        return true;
+    }
+    bool EndArray(rapidjson::SizeType) {
+        mode.pop();
+        return true;
+    }
+};
+    /**
```
ptheywood, line 79:
```suggestion
/**
```

include/flamegpu/runtime/detail/SharedBlock.h
Robadob: I think there was a bug related to this, that you couldn't build with seatbelts turned off. It was probably in fujitsu and has been fixed here (not enough context in 2 lines of code shown to check off-hand).

include/flamegpu/runtime/detail/curve/DeviceCurve.cuh
Robadob: Doesn't bother me if you wish to refactor, so long as "PBM" is in the doc comment for my comprehension.

ptheywood (changes requested):
As discussed I've reviewed this focussing on the headers, and only quickly reviewing the implemntation.

Generally looks good, just some docstrings haven't been updated to reflect changes to the API. 

There's an absence of any python tests of the host portion of the graph api, with only codegen tests.

Could do with some python tests mirroring atleast some of the c++ tests just to make sure the interface is relatively nice to use in python. 

Also would prefer CSR/CSC used in place of PBM/IPBM, but it's not important enough to change.

include/flamegpu/runtime/environment/HostEnvironmentDirectedGraph.cuh
Robadob: Yup `Edge operator[](SrcDestPair source_dest_vertex_ids);`

ptheywood:
seatbelts=off builds fail without a rebase, due to the `strings` bug fixed elsewhere. 

seatbelts=on builds pass all tests (c++ and python) under linux as it stands.


I *think* the conflict wants to be resolved as follows (and changed to avoid the py3.8? bug that has been fixed in master since) but I am not 100% on this, so not force pushing.

```python
        if t.targets[0].id not in self._locals :
            # Special case, catch message.at() where a message is returned outside a message loop
            if hasattr(t.value, "func") and isinstance(t.value.func, ast.Attribute) and t.value.func.attr == 'at' :
                if t.value.func.value.id == self._input_message_var :
                    self._standalone_message_var.append(t.targets[0].id)
            # Special case, track which variables hold directed graph handles
            elif hasattr(t.value, "func") and isinstance(t.value.func, ast.Attribute) and t.value.func.attr == 'getDirectedGraph' :
                if t.value.func.value.value.id == "pyflamegpu" and t.value.func.value.attr == "environment" :
                    self._directed_graph_vars.append(t.targets[0].id)
            # Special case, definitions outside of agent fn are made const
            if self._indent == 0:
                self.write("constexpr ")
```

include/flamegpu/util/StringPair.h
Robadob: It's not a breaking change.

`StringPairHash` is redefined below as a `typedef`, I just made the implementation more generalised via templating.

Robadob:
Windows Release SEATBELTS=OFF C & Python tests now pass.

ptheywood:
Linux release seatbelts=ON C tests all pass post-rebase.

ptheywood (approved):

PR #1097, opened 2023-07-28
Update 'latest' CI to CUDA 12.2

ptheywood:
Update CI and windows installation script to CUDA 12.2.

Local linux builds behave as intended, although the included driver update may break the context creation timing test (see #1096)


mondus (approved):
Approved subject to CI passing

PR #1100, opened 2023-07-28
Draft Release CI: Replace Deprecated ::set-output with GITHUB_OUTPUT.

### LINKED ISSUE #1017: Draft-Release CI set-output deprecation warnings
The final step of the CI workflow which produces the wheels and attached them to the draft release uses some now deprecated commands.

This CI runs incredibly infrequently so had been missed.

They should be upgraded (and probably tested independently of the main repo / with a much lighter weight equivalent

image

https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter

ptheywood:
Replace Deprecated ::set-output with GITHUB_OUTPUT in the draft release workflow.

Closes #1017

### DIFF
--- a/.github/workflows/Draft-Release.yml
+++ b/.github/workflows/Draft-Release.yml
@@ -609,7 +609,7 @@ jobs:
         done
         echo "files: ${files}"
         # Set the list of files as the output for this step
-        echo "::set-output name=files::${files}"
+        echo "files=${files}" >> "$GITHUB_OUTPUT"
 
     # Extract information from the tag which is required for the draft github release
     - name: Process Tag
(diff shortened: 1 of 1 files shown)


Robadob (approved):

Robadob:
You could trigger a draft release on this branch?

ptheywood:
> You could trigger a draft release on this branch?

Edits to the draft release workflow do trigger it, but the part this file editied is only triggered when it was caused by a tag push (otherwise the tag value would be empty). 

I've edited the top post with a linked example showing the tag variable being set the old way and the new way, with the same relative change as performed here, so I am pretty confident this is correct.

PR #1102, opened 2023-08-18
Add post-release CI which triggers wheelhouse creation via gh

### LINKED ISSUE #645: Self-hosted (ghpages) pip archive html file
If we are going to release on pypi, we will likely only be able to host `pyflamegpu` packages built with a single cuda architecture, which Ideally should be the oldest version with broad support at the time of release.  I.e. CUDA 11.2 would be my preference, as this will work with CUDA 11 releases >= 11.2, so wide support. Alternatively CUDA 11.0 for the oldest support, but this does not have forwards compatibility.

In some cases,we might want to provide multiple version. Ie CUDA 11.0 (collab) and 11.2 (broader support), as well as in the future CUDA 11.x and CUDA 12.x. In which case we use the python local package version number to embed this, i.e. `+cuda112` for cuda 11.2.

We could support users installing these via pip, withtout them being uploaded to pypi by leveraging the github release artifacts, github actions, and github pages.

Users could then install a potential 2.0.0 release built with CUDA 11.2  via:
```
python3 -m pip install -f https://download.flamegpu.com/whl/stable.html pyflamegpu==2.0.0+cuda112
``` 

Or alternatively pre-release versions using the `--pre` flag

```
python3 -m pip install  --pre -f https://download.flamegpu.com/whl/prerelease.html pyflamegpu==2.0.0-rc1+cuda112
``` 

` --no-index` may also be useful in the above commands?

This is how pytorch handle pip installation of alternate versions.

As with pip distribution in general, it has downsides re: alterante builds. In this case you cannot ask for the latest pyflamegpu built with CUDA XY, instead you must specify a specific minor version and the local version.
Poetry / other python packaging might also not be able to resolve this if they do not support the -f flag for pip subpackages.

One way to implement this would be:
+ Github action, which on release publication:
  + Queries github for the assets of the release (they may be available in the gha context already)
  + Insert html links to these assets into a .html file, commited to the gh_pages branch of an alternate repo, which is hosted via ghpages on a subdomain of flamegpu.com. 

Some of the following commands may be useful to implemetn this.

Using the `gh` command line tool which is installed into actions by default, we can:

get a list of the last `10` releases using:

```bash
gh release list --limit 10 
```

get a json object containing all of the .whl assets for a specific release tag using: 

```bash
gh release view v2.0.0-alpha --json assets --jq ".assets.[] | select(.name|endswith(\".whl\"))"
```

For more information, see comments on #605.

ptheywood:
Adds a CI workflow which is triggered by release publication, which will then trigger a different workflow on another repository.

This requires a PAT to be created and stored in the appropraite repository secret. 

We could extend this to also trigger docs builds for us, but would need to maintain 2 PATs, and potentially the website if we were to make the website build fetch information about the wheelhouse for instance.

# Todo 

+ [ ] Finish setting up the other repository
+ [ ] Create PAT and store in the relevant secret on this repository
+ [ ] Update readme with wheelhouse information / instructions. 
  + I might add a basic html page to the wheelhouse with instructions on use, pointing at each of the generated sub-wheel houses? It could potentially just go in the root wheelhouse but not 100% if that is valid or not.

Closes #645

### DIFF
--- a/.github/workflows/Post-Release.yml
+++ b/.github/workflows/Post-Release.yml
@@ -1,0 +1,23 @@
+name: Post Release
+# On publication of a release, or when manually triggerered
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+jobs:
+  # Trigger creation of the wheelhouse via an api call. PAT required.
+  trigger-wheelhouse-ci:
+    runs-on: ubuntu-latest
+    steps:
+      # Use gh to trigger the workflow_dispatch event on the appropriate workflow in another repository.
+      # This step may fail if the PAT is invalid or has expired
+      # If so, create a fine-grained pat with actions: write and permissions on the appropriate repository, storing it in the relevant secret on this repository.
+      - name: Trigger FLAMEGPU2-wheelhouse
+        env:
+          GH_TOKEN: ${{ secrets.WHEELHOUSE_TOKEN }}
+          REPO: "FLAMEGPU/FLAMEGPU2-wheelhouse"
+          WORKFLOW: "ci.yml"
+        run: |
+          gh workflow run ${WORKFLOW} -R ${REPO}
+          sleep 5
+          gh run watch -R ${REPO} $(gh run list -R ${REPO} -w ${WORKFLOW} -L1 --json databaseId --jq .[0].databaseId)


.github/workflows/Post-Release.yml
```diff
@@ -0,0 +1,23 @@
...
+  # Trigger creation of the wheelhouse via an api call. PAT required.
+  trigger-wheelhouse-ci:
+    runs-on: ubuntu-latest
+    steps:
+      # Use gh to trigger the workflow_dispatch event on the appropriate workflow in another repository.
+      # This step may fail if the PAT is invalid or has expired
+      # If so, create a fine-grained pat with actions: write and permissions on the appropriate repository, storing it in the relevant secret on this repository.
+      - name: Trigger FLAMEGPU2-wheelhouse
+        env:
+          GH_TOKEN: ${{ secrets.WHEELHOUSE_TOKEN }}
```
mondus, line 17:
```suggestion
          GH_TOKEN: ${{ secrets.WHEELHOUSE_TOKEN }}  # Note that this will periodically need to be renewed
```

mondus (approved):

ptheywood:
Manually triggerd the new post-release CI post merge:

https://github.com/FLAMEGPU/FLAMEGPU2/actions/runs/5974872939

Which successfully triggered a run of the wheelhouse deployment ci:

https://github.com/FLAMEGPU/FLAMEGPU2-wheelhouse/actions/runs/5974874888

So that's behaved as intended, and will auto rebuild the wheelhouse on the next github release

PR #1105, opened 2023-09-14
Examples readme

Robadob:
Also noticed diffusion example wasn't using orthographic vis.

### DIFF
--- a/examples/cpp/diffusion/src/main.cu
+++ b/examples/cpp/diffusion/src/main.cu
@@ -137,7 +137,6 @@ int main(int argc, const char ** argv) {
 
     /**
      * Create visualisation
-     * @note FLAMEGPU2 doesn't currently have proper support for discrete/2d visualisations
      */
 #ifdef FLAMEGPU_VISUALISATION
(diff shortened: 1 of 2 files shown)


ptheywood (approved):
Text is great. 

The images when set to 100% width look a bit blurry though, as the linked resources seem to be 480x360 px, so with the `style="width:100%` they get pretty blurry. 

Could probably just do them as standard  markdown embeds instead and not worry about the width etc? 

i.e. instead of:

```
<div align="left">
    <a href="https://www.youtube.com/watch?v=eE646_6NfqQ">
        <img src="https://img.youtube.com/vi/eE646_6NfqQ/0.jpg" style="width:100%;">
    </a>
</div>
```

```
[![Boids screenshot](https://img.youtube.com/vi/eE646_6NfqQ/0.jpg)](https://www.youtube.com/watch?v=eE646_6NfqQ)
```

[Boids screenshot](https://www.youtube.com/watch?v=eE646_6NfqQ)

Unless there's a way to get a higher res youtube still?

PR #1116, opened 2023-09-26
DeviceAPI::isAgent() DeviceAPI::isState()

### LINKED ISSUE #1108: Access/Check agent state inside an agent function
### Discussed in https://github.com/FLAMEGPU/FLAMEGPU2/discussions/1107

<div type='discussions-op-text'>

<sup>Originally posted by **daniele-baccega** September 18, 2023</sup>
Hi,
is it possible to access to the agent state inside an agent function?

Thank you</div>

-------------

Need to decide a suitable API, but otherwise should be a simple PR.

Can we make the agent state comparison compile time? aka free?

Robadob:
With new C/Python tests

Closes #1108

## Todo

* C/non-RTC
  - [x] Implementation
  - [x] Tests
* Python/RTC
  - [x] Implementation
  - [x] Tests
* Agent Python
  - [x] Implementation?
  - [x] Tests
- [x] Docs (https://github.com/FLAMEGPU/FLAMEGPU2-docs/pull/164)

## Note

* `strcmp()` is not defined in device code, so ripped a small correct looking implementation off stack overflow (found in `util/dstring.h`).

**Branch should be squashed**

### DIFF
--- a/include/flamegpu/util/dstring.h
+++ b/include/flamegpu/util/dstring.h
@@ -1,0 +1,24 @@
+#ifndef INCLUDE_FLAMEGPU_UTIL_DSTRING_H_
+#define INCLUDE_FLAMEGPU_UTIL_DSTRING_H_
+
+#include <cuda_runtime.h>
+
+/**
+ * Device implementations of required <cstring> functionality
+ */
+
+
+/**
+ * strcmp() - Compare two strings, return 0 if equal, otherwise return suggests order
+ *
+ * @note Implementation based on https://stackoverflow.com/a/34873763/1646387
+ */
+__device__ __forceinline__ int dstrcmp(const char *s1, const char *s2) {
+    const unsigned char *p1 = (const unsigned char *)s1;
+    const unsigned char *p2 = (const unsigned char *)s2;
+
+    while (*p1 && *p1 == *p2) ++p1, ++p2;
+
+    return (*p1 > *p2) - (*p2  > *p1);
+}
+#endif  // INCLUDE_FLAMEGPU_UTIL_DSTRING_H_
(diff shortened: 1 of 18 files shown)


src/flamegpu/simulation/detail/CUDAAgent.cu
```diff
@@ -132,7 +132,7 @@ void CUDAAgent::setPopulationData(const AgentVector& population, const std::stri
         if (state_name == ModelData::DEFAULT_STATE) {
             THROW exception::InvalidAgentState("Agent '%s' does not use the default state, so the state must be passed explicitly, "
                 "in CUDAAgent::setPopulationData()",
-                state_name.c_str(), population.getAgentName().c_str());
```
Robadob, line 135: Had this exception thrown whilst writing test, noticed it was wrong.

include/flamegpu/util/dstring.h
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * strcmp() - Compare two strings, return 0 if equal, otherwise return suggests order
+ *
+ * @param s1 First string to be compared
+ * @param s2 Second string to be compared
+ *
+ * @note Implementation based on https://stackoverflow.com/a/34873763/1646387
+ */
+__device__ __forceinline__ int dstrcmp(const char *s1, const char *s2) {
```
mondus, line 21: This will not be resolved at compile time for RTC due to the use of the while loop. If we really wanted to have this resolve at compile time we would have to find a way for this to work with `constexpr` or a through template recursion. I don't think this is necessary though and we should flag that string comparisons are not particularly fast in the docs.

mondus (approved):

PR #1127, opened 2023-10-16
BugFix: Agent Python did not account for variable scope.

### LINKED ISSUE #1125: agent native does not account for scope
```python
for edge in fgraph.outEdges(vertex_index):
    foo = edge.getPropertyInt("foo")

for edge in fgraph.inEdges(vertex_index):
    foo = edge.getPropertyInt("foo")
```

translates to 

```c++
for (const auto& edge : fgraph.outEdges(vertex_index)){
    auto foo = edge.getProperty<int>("foo");
}
for (const auto& edge : fgraph.inEdges(vertex_index)){
    foo = edge.getProperty<int>("foo");
}
```

This could be fixed by updating `self._locals` to instead store tuples that log the value of `self._indent` at declaration. Then `self._locals` could be purged of out of scope vars as `leave()` is triggered which decrements `self._indent`.

I may try to address this after network message has been merged (rather than complicate my existing changes).

Robadob:
Closes #1125

*Note: Current merges into #1089, as there was a possibility of conflict that I wished to avoid.*

### DIFF
--- a/swig/python/codegen/codegen.py
+++ b/swig/python/codegen/codegen.py
@@ -118,7 +118,7 @@ class CodeGenerator:
         self.future_imports = []
         self._indent = 0
         # dict of locals used to determine if variable already exists in assignments
-        self._locals = ["pyflamegpu"]
+        self._locals = {"pyflamegpu": 0}
         self._device_functions = []
         self._message_iterator_var = None             # default
         self._input_message_var = 'message_in'        # default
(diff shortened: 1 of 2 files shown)


mondus (dismissed):
This seems eminently sensible so happy for this to be merged. I cant think of any scenario where this approach would not work.

Robadob:
Thanks
Will probably merge it after static graph

ptheywood (approved):

Issue #1137, opened 2023-10-27
Incorrect singletons->string guarding


ptheywood:
#1116 introduced bug(s) related to the guarding of `singletons->strings` in `CUDASimulation.cu`.

`strings` is only defined as a member for builds with `FLAMEGPU_SEATBELTS` defined to a truthy value (i.e. not defined to 0). 

However, it is used in places where those guards are not present as of 7a7944d.

This needs fixing (and we should add a `FLAMEGPU_SEATBELTS=OFF` build to CI to prevent this from occurring in the future, #1136).

---

Discovered by @daniele-baccega in #1135

PR #1138, opened 2023-10-27
CI: Add FLAMEGPU_SEATBELTS=OFF CI via matrix include:

### LINKED ISSUE #1136: CI `FLAMEGPU_SEATBELTS=OFF`
We don't appear to have any CI builds which compile with FLAMEGPU_SEATBELTS=OFF, which is an oversight.

We should have atleast one build with and without SEATBELTS, ideally on each OS. 

---

Exposed by a SEATBELTS=OFF only bug.

https://github.com/FLAMEGPU/FLAMEGPU2/discussions/1135

ptheywood:
CI: Add FLAMEGPU_SEATBELTS=OFF CI via matrix include:

Using include is very verbose, but wanted to not multiply the exisiting matrix by 2, or trial and error expand the exclude region.

Switching to dynamic build matrices via a 2 step job would be nice but likely significant effort

Closes #1136


ptheywood:
This should fail currently, until #1137 is resolved.

ptheywood:
It seems that my existing, working use of complex objects within a matrix is not strictly correct/supported even though it works. 

In the web editor, the cudacxx objects are reported as an error. 

> Matrix options must only contain primitive values

I might have to figure out the "correct" way to do this in general (I.e. where using a specific cuda version means a specific value of cuda_arch/hostcxx should be used). 

It might be that we have to do a fully `include` based matrix for that to be possible, which will quickly become incredibly verbose for wheel generation...

Robadob:
Would a simpler compromise be to configure ALL standalone example builds to be `SEATBELTS=OFF`?

That would force a main library build, but wouldn't catch any device API template code (only going to fully get that with the test suite).

It's possibly a suitable compromise, if release flow ensures tests are built `SEATBELTS=OFF` too

ptheywood:
That would be an option to quickly get a `FLAMEGPU_SEATBELTS=OFF` build covered, but the matrix yaml I've used everywhere for the cuda compiler would still be invalid. 

From some github issues / feature requests it seems that lists can be used as matrix elements, so potentially could use something like:


```
` - compilerstuff: ["12.3", "50", gcc-12, ubuntu-22.04]
```

instead of 

```
- cuda: "12.3"
  cuda_arch: "50"
  hostcxx: gcc-12
  os: ubuntu-22.04
```

Though its a bit less readable / easy to make mistakes. 

Thankfully I did map matrix values to env vars where they are used, so the changes should only need to be prior to jobs starting. I'll mock up the matrix stuff in a separate repo when I have some time.

ptheywood:
This is now ready for review, assuming that CI passes. 

There are now 6 builds for regular Ubuntu and Windows CIs:

+ Latest CUDA 12, Release, Console
+ Latest CUDA 12, Release, Vis
+ Latest CUDA 12, beltsoff, Console
+ Latest CUDA 11, Release, Console
+ Latest CUDA 11, Release, Vis
+ Oldest CUDA 11, Release, Console

We could probably drop the CUDA 11.8 Release Vis build, as the vis shouldn't be significantly tied to the CUDA version, but it might be worth keeping.

Adding a single GLM build might be worthwhile, but with the current matrix syntax that will be a lot of additional exclude statements.

Robadob (approved):
The changes look sensible. Can merge assuming CI passes.

Robadob:
> Adding a single GLM build might be worthwhile, but with the current matrix syntax that will be a lot of additional exclude statements.

Useful sure, but not a priority whilst GLM remains experimental and not fully supported by Python.

PR #1147, opened 2023-11-08
AgentPython capture external constants.

### LINKED ISSUE #1144: AgentPython does not capture external variables/constants
As raised in #1142, where a user is attempting to port the Sugarscape example to agent python.

The below code does not generate something which can be compiled, as `AGENT_START_COUNT`'s declaration is not included if passed to codegen via the identifier of the agent function.

```py
AGENT_START_COUNT = int(2)

@pyflamegpu.agent_function
def movement_request(message_in: pyflamegpu.MessageArray2D, message_out: pyflamegpu.MessageArray2D):
    foo = AGENT_START_COUNT 
    return pyflamegpu.ALIVE
```

The agent python boid's example however uses device functions, which do appear to be dynamically pulled in. Hence, the reason a user would expect this to work.

------------

This does however create concern that the included `AGENT_START_COUNT` would exist as an implicit `__device__` variable, which could lead to race conditions if misused by a user.
It make make more sense to convert them to either `#define` or `constexpr`.

Robadob:
```py
import pyflamegpu
import typing
TEST: typing.Final = 5
TEST2: typing.Final = 12
TEST2: typing.Final = math.sqrt(12 * 36)
TEST3: pyflamegpu.constant = 12
TEST4: int = 14
```
converts to
```c++
constexpr auto TEST = 5;
constexpr auto TEST2 = 20.784609690826528;
constexpr auto TEST3 = 12;
```

**todo**

- [x] tests
- [x] Do something inside actual `codegen()` to make global vars `constexpr`? (if a user passes in a string)
- [x] Replace `inspect.get_attributes()` (Its python 3.10+, can be emulated via `inspect.get_members()`)

Closes #1144

### DIFF
--- a/swig/python/codegen/__init__.py
+++ b/swig/python/codegen/__init__.py
@@ -40,6 +40,20 @@ def translate(function: Union[str, Callable]) -> str:
         # get source for function and preprend device functions
         function_source = prepend_source + inspect.getsource(function)
         tree = ast.parse(function_source)
-        return codegen(tree)
+        # Filter constants
+        module_annontations = inspect.get_annotations(module) #  requires python 3.10
+        module_members = inspect.getmembers(module);
+        print(module_annontations)
+        print(module_members)
+        prepend_c_source = ""
+        # Find all annotated variables
+        for key, val in module_annontations.items():
+            print(val.__name__)
+            if val.__name__ == "Final" or val.__name__ == "constant":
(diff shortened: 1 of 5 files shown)


mondus (changes requested):
Can we have a test for 

`TEST2: typing.Final = some_user_defined_func()`

The expected result of this is that it should inspect the function and assign a value.

ptheywood (approved):
New tests all pass under linux (and existing codegen tests)

PR #1150, opened 2023-11-15
Migrate to Jitify2

Robadob:
Migrate from Jitify (1) to Jitify2

- Jitify1 is no longer recieving updates
- Significantly improves rtc kernel runtime with CUDA 12.3+

---

- [x] RTC and Execution (Works with CUDA12.0, Windows/Linux)
- [x] Serialisation/Deserialisation (jitify2 pre-processed serialised objects are 2-3x larger)
- [x] CUDA 12.3 support
  - [x] Works with https://github.com/NVIDIA/jitify/pull/128, hence wait for merge.
  - [x] Same branch deprecates the `launch()` method (used in `CUDASimulation`), says to replace with `launch_raw()`.
- [x] Investigate access violation from `cuda().ModuleUnload()` during sim shutdown, when `CUDAAgent` map is cleared by `CUDASimulation` destructor.
    * Has occurred on Windows & Linux, but its inconsistent.
    * gdb log: https://gist.github.com/Robadob/3cb0e93014d56f05587f4ea1ea581203 (with `jitify2-misc-fixes2` branch)
- [x] Reimplement jitify1 demangle used in `curve_rtc.cpp`
- [x] Optimise serialisation load time
    * Jitify2 serialises pre-NVRTC, this is 50x slower and produces 2x larger serial blob
    * Waiting for clarity: https://github.com/NVIDIA/jitify/issues/133
* Optimise compile time
   - [ ] Rework of old header hack? (would require loading headers from file)
       * Preloading fgpu headers only cuts agent fn time from 6.8s to 4.1s.
       * Loadings CUDA headers too makes a big difference, but these may not be particularly stable between version
   - [ ] Offline pre-process FLAMEGPU2 include hierarchy into a single header file? (using jitify tools?)
   - [x] Wait for [this PR](https://github.com/NVIDIA/jitify/pull/131) to be merged?
 - ~Visual Studio 2019 support (we may be able to drop this)~
 - [x] ManyLinux2_28 support


.github/scripts/install_cuda_centos.sh
```diff
@@ -91,9 +92,12 @@ do :
     if [[ ${package} == libcu* ]] && version_lt "$CUDA_VERSION_MAJOR_MINOR" "11.0" ; then
         package="${package/libcu/cuda-cu}"
     fi
-    # CUDA < 11, -devel- packages were actually -dev
-    if [[ ${package} == *devel* ]] && version_lt "$CUDA_VERSION_MAJOR_MINOR" "11.0" ; then
-        package="${package//devel/dev}"
+    # libnvjitlink not required prior to CUDA 12.0
+    if [[ ${package} == libnvjitlink-dev* ]] && version_lt "$CUDA_VERSION_MAJOR_MINOR" "12.0" ;then
```
Robadob, line 96: Not sure if this is as narrow as it could be, but I wanted to make minimal changes to get it working.

The wildcard on `libnvjitlink-dev` can probably be removed.

Robadob:
Update regarding header pre-loading with Jitify2/CUDA 12.3

**Windows/CUDA 12.0**
````
No preload
Millis: 6822.000000
Millis: 6853.000000

Preloading FLAMEGPU headers
Millis: 4045.000000
Millis: 4277.000000

Preload FLAMEGPU + CUDA headers
Millis: 1296.000000
Millis: 1667.000000

From cache
Millis: 446.000000
Millis: 774.000000
```

**Linux/CUDA 12.3**

```
Jitify 2 from scratch (Waimu)
Millis: 25318.000000
Millis: 24143.000000

Preload FLAMEGPU + CUDA headers
Millis: 1376.000000
Millis: 2218.000000

From cache
Millis: 685.000000
Millis 1372.000000
```

CUDA 12.0 has [~30 CUDA headers to preload](https://gist.github.com/Robadob/396198a91d62d6e52d44f8d5609d871b).
CUDA 12.3 has [~257 CUDA headers to preload](https://gist.github.com/Robadob/d5e37a9153b96fb567b3a07e456a8376). (List contains some dupes)

Not clear whether we would want to generalise this code, to better handle different CUDA versions, because we could be potentially needing to update it with each CUDA update.

Robadob:
Current issue holding back the Jitify2 preprocesor branch is that it expects our flamegpu headers to be included as system header `<>` rather than `" "`. Waiting to here back from the dev (Ben) before I try to correct that on our side.

Robadob:
Did three full test runs last night, all passed

Linux/CUDA12.3/Seatbelts ON/GLM ON/Release
Linux/CUDA12.3/Seatbelts OFF/GLM ON/Release
Windows/CUDA12.0/Seatbelts ON/GLM OFF/Debug

ptheywood:
The updated jitify2 branch uses c++ features which error on windows with CUDA 11.0 and 11.1, but 11.2+ is fine. 

This was a planned change anyway for CCCL support (see #1155), so will just need dealing with here too/instead.

Another option (to discuss) would be to drop 11.x more widely (although I'd rather not yet) as the plan is to only support 2 major cuda versions and I expect 13 to turn up this year, but would rather not do that yet. 

For wheel production on github actions, we do need to maintain vs 2019 support, with the recent change for vs2022 updates requireing 12.4, even if we drop cuda 11.0.
(when the vs2019 image dissapears we will have to find a workaround for that, or just tell windows users to use WSL). Mainstream window server 2019 support has ended, but extended support is until 2029. windows server 2026 was dropped by github actions ~2 months after mainstream supported ended, but that was before Microsoft purchased github.

ptheywood:
The required changes to jitify2 have now been merged into the `jitify2` branch, as commit `46936132050a6035ff62e1aa050b50d849f1991b`. 

Our `jitify2_rebase_2025_05` branch is the most recent rebase of this branch, which needs a few tweaks then should be force pushed over this? 

- [ ] Update pinned jitify commmit hash / branch
- [ ] Remove/Adjust the header log filtering from preprocessing.
- [ ] Fix/suppress VS2019 warnings (in our code)
- [ ] Any remaininig tasks from the top of this branch

ptheywood:
I've included the jitify2 fixes as `.patch` files applied when jitify is fetched, similar to current rapidjson patching.

This might require a fresh build dir, not 100%. 

Also squashed the commits to get rid of fixup commits etc, probably could have kept some of them but it was a bit conflicty when I gave it a quick go. Can pull it from the reflog if we want to split it back up.

Behaves fine in CUDA 12.0 linux, should probably double check it runs ok on Windows too but otherwise this can go in (with a new issue being created to bump the hash and remove the patches when merged upstream). 

CUDA 13 requires subsequent chages to test, so might need to ammend the patch in #1302

PR #1151, opened 2023-11-20
Expose JitifyCache via Python interface.

Robadob:
Required by https://github.com/FLAMEGPU/ABM_Framework_Comparisons/pull/9

To allow

```py
if "--disable-rtc-cache" in sys.argv:
    rtc_cache = pyflamegpu.JitifyCache.getInstance()
    rtc_cache.useMemoryCache(False)
    rtc_cache.useDiskCache(False)
```

The changes to `JitifyCache.h/.cu` are required to suppress a SWIG warning, "Nothing known about `jitify::experimental::KernelInstantiation`".

This will cause conflicts with #1150


ptheywood (approved):
Added a couple of test cases to programatically check this is exposed, which it is.

I'll let you (@Robadob) merge this once CI passes, assuming you are happy with my tests.

PR #1152, opened 2023-11-23
Adds RunPlanVector::setPropertyStep()

Robadob:
This has behaviour similar to Python's range(), more appropriate than lerp in many cases with integers.

Providing this will hopefully help people avoid using lerp wrongly for integers.

https://github.com/FLAMEGPU/FLAMEGPU2-docs/pull/175

alt possible names:
* `setPropertyUniformStep()`
* `setPropertyInterval()`
* `setPropertyUniformInterval()`
* `setPropertyRange()` (reminds me of Python's `range()` although it's not 1-1 as the length is predefined).


ptheywood (approved):
Looks fine, not tested locally but test looks sane. 

(`step` might be getting a bit overloaded, but there should be enough context that it's not a problem)

Robadob:
I've ran both tests Windows.

PR #1153, opened 2023-11-27
BugFix: AgentPython was missing ID->id_t type conversion.

Robadob:
Improved detail in various codgen exceptions too.


Robadob:
3 codegen tests failing with these changes

Issue #959, opened 2022-11-03
Double Precision Spatial Messaging


ptheywood:
The current spatial messaging implementation uses `float` for the implicit `x`/`y`/`z` members (and associated methods such as `getVirtualX`). 

Some users might want double precision spatial coordinates for e.g. large ranges but some dense regions, so we might want to provide a `double` variant.

Performance will be worse due to reduced FP64 units and incresed memory requirementsm but may be neccesary.

We could implement this in the future without breaking the API by either creating new copies of the single precision classes, with `Double` in the name.
Alternatively we could templating the current classes, then typedef concrete versions, using the current name for single precision and a new name for double. This would be an ABI break, but not an API break (and we don't maintain a stable ABI).

Robadob:
> Alternatively we could...

This would be the cleaner approach.

PR #1160, opened 2023-12-12
BugFix: MsgSpatial interaction radius correct when not a factor of en…

### LINKED ISSUE #1157: Spatial messaging shrinks interaction radius to a factor of width, when not a factor
https://github.com/FLAMEGPU/FLAMEGPU2/blob/3801865978db1f1dd4d0652792501ef3f298e0d1/src/flamegpu/runtime/messaging/MessageSpatial2D.cu#L44

Consider:
Width 105
Radius 10

This currently creates 11 bins.
Therefore the 11th bin would only be half occupied (covering 5 width, rather than 10).
Therefore any agents in the 10th bin, will only access the 11th bin and not wrap to the 1st bin.
Therefore missing out on messages in that direction, creating bias.

The solution would be to `floor()` rather than `ceil()` and accept that the 10th bin will be denser as it covers any overflow width.

Robadob:
## todo
- [x] Fix bug
- [x] Seatbelts error if wrapped is used when radius is not a factor of environment bounds
    * Happy for someone to update my `fmod()` solution to something cleaner, floating point isn't great.
- [x] Test that bug is fixed
- [x] Test that wrapped is blocked under `SEATBELTS=ON`
- [x]  document wrapped limitation (https://github.com/FLAMEGPU/FLAMEGPU2-docs/pull/176)

Closes #1157

### DIFF
--- a/include/flamegpu/runtime/messaging/MessageSpatial2D/MessageSpatial2DDevice.cuh
+++ b/include/flamegpu/runtime/messaging/MessageSpatial2D/MessageSpatial2DDevice.cuh
@@ -500,6 +500,13 @@ class MessageSpatial2D::In {
              // Return iterator at min corner of env, this should be safe
              return WrapFilter(metadata, metadata->min[0], metadata->min[1]);
          }
+         if (fmodf(metadata->max[0] - metadata->min[0], metadata->radius) > 0.00001f ||
+             fmodf(metadata->max[1] - metadata->min[1], metadata->radius) > 0.00001f) {
+             DTHROW("Spatial messaging radius (%g) is not a factor of environment dimensions (%g, %g),"
+                 " this is unsupported for the wrapped iterator, MessageSpatial2D::In::wrap().\n", metadata->radius,
+                 metadata->max[0] - metadata->min[0],
+                 metadata->max[1] - metadata->min[1]);
+         }
 #endif
          return WrapFilter(metadata, x, y);
      }
--- a/include/flamegpu/runtime/messaging/MessageSpatial2D/MessageSpatial2DDevice.cuh
+++ b/include/flamegpu/runtime/messaging/MessageSpatial2D/MessageSpatial2DDevice.cuh
@@ -599,8 +599,8 @@ T MessageSpatial2D::In::WrapFilter::Message::getVariable(const char(&variable_na
 __device__ __forceinline__ MessageSpatial2D::GridPos2D getGridPosition2D(const MessageSpatial2D::MetaData *md, float x, float y) {
     // Clamp each grid coord to 0<=x<dim
     int gridPos[2] = {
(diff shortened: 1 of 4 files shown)


Robadob:
I did run full test suite of a Debug build, just closed it without thinking to copy results.

include/flamegpu/runtime/messaging/MessageSpatial2D/MessageSpatial2DDevice.cuh
```diff
@@ -500,6 +500,13 @@ class MessageSpatial2D::In {
              // Return iterator at min corner of env, this should be safe
              return WrapFilter(metadata, metadata->min[0], metadata->min[1]);
          }
+         if (fmodf(metadata->max[0] - metadata->min[0], metadata->radius) > 0.00001f ||
```
ptheywood, line 503: Any particular reason/justification for this particular epsilon value?.

Though our current spatial comms is fp32 only anyway, so a magic value is fine, just why this one / is it sufficient for different values

E.g. if the comm rad is a small positive number, will this work? (though if its too small then large env widths will cause other problems anyway, so can prolly get away with a fixed arbitrary epsilon)

Tests only appear to cover a single radii/width combo.

ptheywood (approved):
Tests all pass under linux. 

Code looks sane, other than maybe the comment about the epsilon value selected.
Prolly fine to merge but might be worth thinking about the epsilon / testing with some edge-casey (but still sensible, allocations will fail for very high bin counts) comm radiuses / env widths to make sure the epsilon check still behaves. 

If / when we have a fp64 spatial comm (#959) that would need a different epsilon due to increased precision.

Approving anyway, but might warrent a chagne.

include/flamegpu/runtime/messaging/MessageSpatial2D/MessageSpatial2DDevice.cuh
Robadob: (I'd rather make wrapped work at a tiny cost to perf for non-wrapped to avoid this floating point crap)

ptheywood: I agree that hurting wrapped more and not hurting non-wrapped is the better choice. 

Given the radii and width are known on the host, we could maybe do the check there once, and then set a value on the device somewhere marking if wrapped is possioble for that message list or not? for less perf loss to wrapped usage. An extra global memory read is probably cheaper than a couple of fmodf's?  

Could do it entirley on the host if we made wrapped opt-in on the message list, but I assume that's not desirable? 

Not sure there's a better option than fmodf, jsut the particular epsilon might need some thought / broader testing.

Robadob:
>  An extra global memory read is probably cheaper than a couple of fmodf's?

It's inside SEATBELTS, so perf doesn't matter.

Issue #1164, opened 2023-12-14
Remove unnecessary astpretty dependency


ptheywood:
Astpretty is currently a pyflamegpu dependency in `setup.py.in`. 

It is not actually used within pyfalemgpu's codegen module, only in a test.

It therefore is not a dependency of pyflamegpu, so should be removed from setup.py.in. 

~Usage within the test suite is also condition, guarded by an undefined variable `DEBUG_OUT`. This should be removed too, and can be added back in with the import if/when required.~

Usage within the test suite is guarded by a file-scoped variable `DEBUG_OUT`, which is set to true.
We can make this an optional dependency installed in the test environment alongside pytest.

Robadob:
This impact #1131 too

ptheywood:
Already closed as part of #1166, but I misspelt `closes` in the PR so this didn't auto close :laughing: 

> Closses https://github.com/FLAMEGPU/FLAMEGPU2/issues/1164

PR #1166, opened 2023-12-14
Remove astpretty as a dependency of pyflamegpu

ptheywood:
astpretty is only used for prettyprint debugging in codegen testing - it is not a pyflamegpu dependency.

This commit also makes it an optional test_codegen.py dependency (but installs it in the test venv by default)

Closses #1164


ptheywood:
pytest codegen tests still pass:

but if ran without astpretty installed, a message is output to the debug info

```
python3 -m pytest ../tests/python/codegen/ -s
... 
collecting ... Install astpretty if ast tree printing required for test debugging
...
=========== 57 passed in 0.30s ==========
```

swig/python/CMakeLists.txt
```diff
@@ -413,8 +413,8 @@ if(FLAMEGPU_BUILD_PYTHON_VENV)
     # Install the wheel from the dist dir. Specify a version incase the same build dir has been used for a previous build
     # Relaxed the no-index option to allow setup.py to install requirements
     COMMAND ${VENV_PIP} install --force-reinstall --find-links=dist -U ${PYTHON_DISTRIBUTION_NAME}==${FLAMEGPU_VERSION_PYTHON}
-    # install pytest
-    COMMAND ${VENV_PIP} install pytest
+    # install pytest and astpretty 
+    COMMAND ${VENV_PIP} install pytest astpretty
```
Robadob, line 417: Is there a reason to remove it from the venv, if it's used by the codegen tests?

ptheywood: this is adding it to the venv if pytest is also being installed. 

Previously it would be implicitly installed as a dependency of pyflamegpu.

The changes to test_codegen.py itself making it optional within the test is not really neccesarry, just adds a bit more flexibility, it makes it so that just pyflamegpu and pytest are required, but if codegen debugging is wanted then astpretty might help.

PR #1168, opened 2024-01-04
Cmake version deprecations fix

### LINKED ISSUE #1156: CMake 3.27 Deprecation warnings (dependencies)
Cmake 3.26?27? deprecates compatibility with CMake < 3.5 in `cmake_minimum_required` calls. 

These are triggered by sevarl of our dependencies. We should update versions of our depenenceies to ensure futre CMake support.

- [ ] GoogleTest
  - Fixed by  https://github.com/google/googletest/commit/4fed5f285030085b096c930ff03e42c7814739b2, included in the `v1.14.0` release. 
- [ ] RapidJSON
    - Not fixed upstream, an open issue from May. Could patch ourselves during fetch content?
- Visualisation:
  - [ ] CMakeRC 
    - This is vendored into our cmake rather than fetched, so we can update our committed copy. Upstream has fixed this in https://github.com/vector-of-bool/cmrc/commit/91c9522ee59654e2f4d9701947a4600a4436d076. It is not included in a tagged release.

ptheywood:
Fixes deprecation notice of removal of `cmake_minimum_required` <= 3.5 in dependencencies with CMake >= 3.27 (or 3.26.4?). 

+ [x] Updates Google test to the latest tagged release 
+ [x] Updates rapid json to the most recent version which does not break our method of fetch content + find_package without cmake install, and patches that commit to avoid the cmake_minimum_required error
+ [x] https://github.com/FLAMEGPU/FLAMEGPU2-visualiser/pull/132 fixes the same issue with CMakeRC there. The commit hash here will need updating post merge, and then again for the next release.

Closes #1156

### DIFF
--- a/cmake/dependencies/flamegpu2-visualiser.cmake
+++ b/cmake/dependencies/flamegpu2-visualiser.cmake
@@ -7,7 +7,7 @@ include(FetchContent)
 cmake_policy(SET CMP0079 NEW)
 
 # Set the visualiser repo and tag to use unless overridden by the user.
-set(DEFAULT_FLAMEGPU_VISUALISATION_GIT_VERSION "02097b5ca1bf38a52693e926d0d750d972a2ae80")
+set(DEFAULT_FLAMEGPU_VISUALISATION_GIT_VERSION "5de60dc004d524ce9dd07898daa3c84ea1e32203")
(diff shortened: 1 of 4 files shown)


ptheywood:
vis pr merged, need to update hash then can merge this.

PR #1169, opened 2024-01-04
Fix docstring(s) attributed to the flamegpu namespace

### LINKED ISSUE #1149: Incorrectly located docstrings attribtuted to namespaces
From the API docs, some docstrings are being associated to the namespace rather than the class / file they are for. 

e.g. on https://docs.flamegpu.com/api/namespace_flamegpu.html

> Detailed Description
> Collection of cuda related utility methods for internal use. Mostly to allow for graceful handling of device resets when cuda is called by dtors
> 
> This header exists to allow a convenient way to switch between curand implementations
> 
> The main cuRVE header file for the CUDA Runtime Struct Variable Environment (cuRVE) Based off the following article http:// www.gamasutra.com/view/news/127915/InDepth_Quasi_CompileTime_String_Hashing.php This file contains definitions common to HostCurve and DeviceCurve
> 
> CUDA does not implement atomicAdd(double*, double) for pre-pascal GPUs, which do not have the underlying hardware instruction. A CAS based software implementation is required instead, which will be much slower. This implementation is based on the reference implementation prodived by the CUDA toolkit documentation.
> 
> THIS CLASS SHOULD NOT BE INCLUDED DIRECTLY Include File AgentVector.h instead Use AgentVector::CAgent instead of Class AgentVector_CAgent Use AgentVector::Agent instead of Class AgentVector_Agent
> 
> Prodvides a utility method to cleanup after flamegpu. Currently for the only implementation (CUDA) this resets all devices.
> 
> Utility namespace for handling of NVTX profiling markers/ranges, uses if constexpr to avoid runtime cost when disabled
> 
> Macro FLAMEGPU_USE_NVTX is defined via CMake to set the member constexpr
> 

E.g. the first part of that comes from 

https://github.com/FLAMEGPU/FLAMEGPU2/blob/e606d4c2bd3db24d548729339896d8e1a7196fd1/include/flamegpu/detail/cuda.cuh#L4-L8

Which is prior to the `flamegpu` namespace, rather than the nested `cuda` namspace. 

Moving that docstring to before the `namespace cuda` declaration would correctly associate it with the namespace.

https://github.com/FLAMEGPU/FLAMEGPU2/blob/e606d4c2bd3db24d548729339896d8e1a7196fd1/include/flamegpu/detail/cuda.cuh#L15-L16. 

We probably do want to have a docstring somewhere for the flamepgu namspace though to populate that page with a general comment. Could be in any file but flamepgu.h would be the best place?

---

There will likely also be other occurences of this, and other docstring issues, which we might want to resolve. Might just need to do a pass of the api docs at some point if time allows.

ptheywood:
Replaces docstring comments with regular multline comments where appropriate (`/**` -> `/*`)

Moves docstring comments to be correctly associated with the namespace where appropriate.

Adds a single explicit docstring comment for `flamegpu`, in `flamegpu/flamegpu.h`

Closes #1149

### DIFF
--- a/include/flamegpu/simulation/AgentVector_Agent.h
+++ b/include/flamegpu/simulation/AgentVector_Agent.h
@@ -1,8 +1,8 @@
 #ifndef INCLUDE_FLAMEGPU_SIMULATION_AGENTVECTOR_AGENT_H_
 #define INCLUDE_FLAMEGPU_SIMULATION_AGENTVECTOR_AGENT_H_
 
-/**
- * THIS CLASS SHOULD NOT BE INCLUDED DIRECTLY
+/*
+ * THIS HEADER SHOULD NOT BE INCLUDED DIRECTLY
  * Include flamegpu/simulation/AgentVector.h instead
  * Use AgentVector::CAgent instead of AgentVector_CAgent
  * Use AgentVector::Agent instead of AgentVector_Agent
--- a/include/flamegpu/util/cleanup.h
+++ b/include/flamegpu/util/cleanup.h
@@ -1,8 +1,8 @@
 #ifndef INCLUDE_FLAMEGPU_UTIL_CLEANUP_H_
 #define INCLUDE_FLAMEGPU_UTIL_CLEANUP_H_
 
-/**
- * Prodvides a utility method to cleanup after flamegpu. Currently for the only implementation (CUDA) this resets all devices.
+/*
+ * Provides a utility method to cleanup after flamegpu. Currently for the only implementation (CUDA) this resets all devices.
  */
 
 namespace flamegpu {
--- a/include/flamegpu/detail/cuda.cuh
+++ b/include/flamegpu/detail/cuda.cuh
@@ -1,11 +1,6 @@
 #ifndef INCLUDE_FLAMEGPU_DETAIL_CUDA_CUH_
 #define INCLUDE_FLAMEGPU_DETAIL_CUDA_CUH_
 
-/**
- * Collection of cuda related utility methods for internal use.
- * Mostly to allow for graceful handling of device resets when cuda is called by dtors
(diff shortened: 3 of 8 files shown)


ptheywood:
I've not thoroughly checked the full docs, there are too many pages, but I skipped all the explcitly documented namespaces in doxygen html output.

include/flamegpu/simulation/AgentVector_Agent.h
```diff
@@ -1,8 +1,8 @@
 #ifndef INCLUDE_FLAMEGPU_SIMULATION_AGENTVECTOR_AGENT_H_
 #define INCLUDE_FLAMEGPU_SIMULATION_AGENTVECTOR_AGENT_H_
 
-/**
```
Robadob, line 4: Not sure why you're removing the second `**`, I always put them back in.

include/flamegpu/util/cleanup.h
```diff
@@ -1,8 +1,8 @@
 #ifndef INCLUDE_FLAMEGPU_UTIL_CLEANUP_H_
 #define INCLUDE_FLAMEGPU_UTIL_CLEANUP_H_
 
-/**
```
Robadob, line 4: Same here, also inconsistent with other changes in this PR.

Robadob (approved):
Grand scheme the stars are inconsequential, fixes are positive.

include/flamegpu/simulation/AgentVector_Agent.h
ptheywood: ** makes it a docstring, not a multiline comment, which doxygen associates with the next namespace / class / mehtod it gets to, unless told otherwise. 

This is a file level comment, so not a docstring that should be assocated with a particular class nor namespace.

include/flamegpu/util/cleanup.h
ptheywood: its a file level multline comment, not a docstring about a particular namespace nor class (there isn't a cleanup specific namespace or class, just method(s)

Robadob:
Fair, wasn't aware the difference!

PR #1171, opened 2024-01-08
Included cstdint in DiscreteColor.h (it is necessary for gcc 12.3)

daniele-baccega:
I discovered that with gcc 12.3 it is necessary to include cstdint in DiscreteColor.h to use the types int32_t, uint32_t, etc. (the compiler itself suggested me to do this).


Robadob (approved):
Makes sense as `int32_t` is used at line 28, possibly missed as that may be included by the other flame headers included.

Will merge after CI passes. (Hmm CI hasn't popped up, will chat with @ptheywood).

Thanks

PR #1181, opened 2024-02-02
Fix Rapidjosn CMake patching for windows

### LINKED ISSUE #1180: CMake FetchContent patch apply error on reconfigure on Windows
#1168 added a fetch content patch to the rapidjson dependency to resolve a deprecation warning.

Patch commands are non-trivial, and my implementation which behaves well on linux does not behave well on windows on a reconfigure that re-triggers the patch command.

When git apply is triggered when it has previously been applied already, it will fail with a non-zero error code. 

Under linux this can be fixed with `|| true`, but this does not work under windows.

ptheywood:
+ [x] Add re-configuration to CI to check for fetch content apply bugs
+ [x] fix rapidjson CMake patch command for windows

Closes #1180


ptheywood:
Unable to get PATCH_COMMAND to behave on windows. 

Instead calling execute_process to manually check if the patch is applicable, and only apply it if it is. Seems to work. 

First configuration output will now include:

```
-- Found Jitify: C:/Users/ptheywood/code/flamegpu/FLAMEGPU2/build/_deps/jitify-src (found version "0.9")
-- Patching RapidJSON
-- Patching RapidJSON - done
-- RapidJSON found. Headers: C:/Users/ptheywood/code/flamegpu/FLAMEGPU2/build/_deps/rapidjson-src/include
-- Could NOT find Doxygen (missing: DOXYGEN_EXECUTABLE)
```

Second output (i.e. when the patch can't be applied so isn't attempted)

```
-- CUDA Architectures: 50-real;60-real;70-real;80-real;90-real;90-virtual
-- RapidJSON found. Headers: C:/Users/ptheywood/code/flamegpu/FLAMEGPU2/build/_deps/rapidjson-src/include
-- Could NOT find Doxygen (missing: DOXYGEN_EXECUTABLE)
```

This does mean that if the patch becomes stale it will silently not be applied, but that will only occur if the rapidjson hash is changed and the relevant lines are changed too. The patch is only fixing a cmake deprecation notice, and there are PRs upstream to fix that (but they've been open for a while and rapidJSON have also broken their find_package logic anyway, so larger changes will beneeded if we bump the hash any further). 

If there were a race on disk, where between the patch being checked and the patch being applied on disk, the output would instead be as follows, but it would not prevent cmake from progressing. 

```
-- Patching RapidJSON - patching failed
```

Need to check this still behaves on linux prior to marking as ready.

ptheywood:
Linux configure, reconfigure and build are all behaving, so this is good to go subject to CI passing.

Robadob (approved):
Trusting that the rapidjson changes work, because CI passes.
Rest looks sensible.

PR #1182, opened 2024-02-02
Fix Spatial Wrapped invalid env/radii detection

### LINKED ISSUE #1177: Wrapped MsgSpatial interaction radius factor bug
#1160 added checks that comm radii are a factor of the environment when using spacial comms with seatbelts enabled.

This seems to be incorrectly triggering in some cases, including the `python_native` example.

```
python3 ../examples/python_native/boids_spatial3D_wrapped/boids_spatial3D.py -t  -v
FLAME GPU 2.0.0-rc.2+25a37160
Simulation configuration:
        Random Seed: 1706546446
        Steps: 1
RTC Initialisation Processing time: 26.107000 s
Init Function Processing time: 0.000000 s
Processing Simulation Step 0
Traceback (most recent call last):
  File "/users/pheywood/flamegpu/FLAMEGPU2/build-gh-123/../examples/python_native/boids_spatial3D_wrapped/boids_spatial3D.py", line 430, in <module>
    cudaSimulation.simulate()
  File "/users/pheywood/flamegpu/FLAMEGPU2/build-gh-123/lib/Release/python/venv/lib64/python3.9/site-packages/pyflamegpu/pyflamegpu.py", line 9255, in simulate
    return _pyflamegpu.CUDASimulation_simulate(self, *args)
pyflamegpu.pyflamegpu.FLAMEGPURuntimeException: (DeviceError) Device function 'inputdata' reported 40000 errors.
First error:
flamegpu/runtime/messaging/MessageSpatial3D/MessageSpatial3DDevice.cuh(545)[52,0,0][0,0,0]:
Spatial messaging radius (0.05) is not a factor of environment dimensions (1, 1, 1), this is unsupported for the wrapped iterator, MessageSpatial3D::In::wrap().
```

Probably due to inexact floating point stuff.

This needs fixing, and the test suite(s) expanding to cover a broader range of cases / edge cases.

Additionally, this check is currenrlty doing expensive ops per thread on device (when seatbelts enabled) which could instead be done once on the host, then a single bool / bit being set on device and read when the iterator is requested.

ptheywood:
+ [x] Reproduce python-native edge case in c++ test suite
+ [x] Expand edge case tests coverage to exposes a broader range of floating point issues
+ [x] Fix tests
+ [x] Optional: Migrate float logic to the host, making in-device checks a single global read.
+ [x] Update the example back to using a radius of `0.05`

Closes #1177

### DIFF
--- a/.github/workflows/Draft-Release.yml
+++ b/.github/workflows/Draft-Release.yml
@@ -414,16 +414,11 @@ jobs:
         make install
 
     - name: Install setuptools and wheel
-      if: ${{ env.PYTHON >= 3.12 }}
       run: |
         python${{env.PYTHON}} -m pip install setuptools wheel
+
     - name: Add custom problem matchers for annotations
       run: echo "::add-matcher::.github/problem-matchers.json"
-
-    - name: Install setuptools and wheel
-      if: ${{ env.PYTHON >= 3.12 }}
-      run: |
-        python${{env.PYTHON}} -m pip install setuptools wheel
 
     # This patches a bug where ManyLinux doesn't generate buildnumber as git dir is owned by diff user
     - name: Enable git safe-directory
--- a/.github/workflows/Manylinux_2_28.yml
+++ b/.github/workflows/Manylinux_2_28.yml
@@ -134,7 +134,6 @@ jobs:
         make install
 
     - name: Install setuptools and wheel
-      if: ${{ env.PYTHON >= 3.12 }}
       run: |
         python${{env.PYTHON}} -m pip install setuptools wheel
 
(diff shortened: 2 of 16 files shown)


ptheywood:
The following appears to work for values tested so far. Not the cheapest method so worth only doing this once on the host and storing if wrapped is available for a given spatial message list on the device, rather than doing this per call to the wrapped iterator on the device

```c++
template <typename T>
bool approxExactlyDivisible(T a, T b) {
    // Scale machine epsilon by the magnitude of the larger value
    T scaledEpsilon = std::max(std::abs(a), std::abs(b)) * std::numeric_limits<T>::epsilon();
    // Compute the remainder
    T v = std::fmod(a, b);
    // approx equal if the remainder is within scaledEpsilon of 0 or b (fmod(1, 0.05f) returns ~0.05f)
    return v <= scaledEpsilon || v > b - scaledEpsilon;
}
```

Issue #1193, opened 2024-04-02
Python wheel `libnvrtc-builtins.so` 


ptheywood:
Originally identified as part of #1191, although nvrtc sonames are now major version only (since 11.3, using `.so.11.2` or `.so.12.0`), `libnvrtc-builtins.so` is still explicitly versioned, and is dependended upon by the generic `libnvrtc.so` which we are linking against.

In practice, this means that currently you must have the exact version of the CTK installed and available at runtime as was used to build the python wheel, rather than a "compatible" version. 

This was not noticed when testing locally, as CMake causes a `RPATH`/`RUNPATH` to be set pointing at the exact location of the shared object, so if installed in the same location it is found even if `LD_LIBRARY_PATH` does not point to it. 

--- 

A workaround has been added to the google-colab notebooke in https://github.com/FLAMEGPU/FLAMEGPU2-tutorial-python/commit/47cc1c119654c79509934164228f52b4bf59713b, which installs the `nvidia-cuda-nvrtc-cu12==12.0.140` python package to bring in the matching version of `libnvrtc-builtins.so`, and then explicitly imports it via `ctypes.CDLL` to ensure it is loaded.

A robust verison of this fix, would be to add the exact version of `nvidia-cuda-nvrtc-cu12` (or `nvidia-cuda-nvrtc-cu11`) required to our python packages' `extra_install_requires` in `setup.py`, when the packages are intended to be distributed (i.e. only in our CI, not in all local builds to avoid binary bloat). 

In `__init__.py`, we would then need to ensure that the appropraite .so is loaded at runtime if it is not found implicilty, i.e. try and import it system wide, catch errors and if the error is appropraite explicitly load the exact `libnvrtc.so` provided by the python package. 

Unfortunately, `nvidia-cuda-nvrtc-cu11` is only provided on pypi for `11.7.99` and `11.8.89`, so this does not work for our 11.2 wheels...

Alternative options include: 

+ Static linking `libnvrtc`. Bigger wheels, requires CMake >= 3.26 and CUDA >= 11.5 to do robustly. 
+ Package `libnvrtc.so` and `libnvrtc-builtins.so` in the wheel (like vis dlls) and ensure that version is used if needed. Bloated wheels, but avoids pip conflicts. Related to #647 (but not exact) 
+ Actually dlopen things (#647).
[...]

Issue #647, opened 2021-08-18
Manylinux2014 compliant issues (dlopen)


ptheywood:
#644 Enables the production of python wheels built using the manylinux2014 image to provide support for linux installations using glibc 2.17 or newer.

These wheels are still suffixed `linux_x86_64` however, rather than `manylinux2014_x86_64` as they are not manylinux compliant, so the `auditwheel repair` step is disabled.

Manylinux compliance requires that there are is no to external shared objects (i.e. `.so` files which are  not packages inside the wheel).
In some cases `auditwheel repair` would be able to handle this for us, however with CUDA this is not the case.

Key shared objects which are currently linked against are 

+ CUDA Driver API (`libcuda.so`)
+ CUDA Runtime API and libraries (`libcudart.so`, `libcurand.so`, etc)
  + `cudart` may actually be a static link on some platforms a the moment 
+ Visualisation dependenceis

We **cannot** package `libcuda.so`, as this is against the EULA. Instead current practice (tensorflow etc) is to achieve compliance by `dlopen`ing the library ([gitter source](https://gitter.im/tensorflow/sig-build?at=603d5752e8267a46f2e663bc),  [discuss.python.org](https://discuss.python.org/t/what-is-the-recommended-way-of-distributing-cuda-dependent-packages/9335))

This can also be applied to other cuda libraries to avoid inflating our already too large for standard pypi wheels significantly.

Other shared dependencies such as the visualisation we could package within the wheel, with some visualisation related libraries being allowed as external .so's, such as libGL.so.1 ([source](https://www.python.org/dev/peps/pep-0599/#the-manylinux2014-policy)).
These could be packaged in the wheel for many linux compliance, and / or we could (in some caes) dlopen these libraries to allow a single, visualisation enabled python distribution which will run on headless machines such as HPC systems (with added cost in terms of file size).

The use of dlopen should probably be opt-in via a CMake option, defaulting to `OFF`, and only used when building manylinux compliant wheels.

ptheywood:
Currently, the only dynamic links that are not allowed bymanlinux are:

```
 libcuda.so.1
 libnvrtc.so.11.2
```
This confirms that libcudart.so is being statically linked (which is default in CMake when using the CUDA language, and can be controlled with a global cmake variable/target property)


Visualiasiton builds hwoever link against `70` .so's of which `~60` are probably not allowed, which may be a decent amount of effort to resolve.

ptheywood:
[...]

PR #1194, opened 2024-04-02
Add Manylinux_2_28 workflow

ptheywood:
Add Manylinux_2_28 workflow to test compatibility prior to expected manylinux2014 EOL (due to cent7 EOL)

+ [x] add standalone manylinux 2_28 workflow for regular testing
+ [ ] add / prep deployment workflow for 2_28. Potentially this can wait till nearer the time, and merging the regular ci workflow sooner might be a worthwhile change.
    + [ ] Ensure this respects changes in #1200, if/when it's possible to use artifact uploads post Node 16 removal. 
+ [ ] Ensure 2_28 workflows are using patchelf to remove the libnvrtc-builtins.so dependency (see #1193)

---

It is not clear that Manylinux_2014 will be removed/deprecated when Cent7 is, but I expect it will / should be. 

2_28 is the oldest still supported manylinux (2_24 is already EOL). 

Produced wheels will not be manylinux compliant, due to linking against non-packaged libs, which still requires #647 and the `dlopen`ing of libcuda.so.

### DIFF
--- a/.github/scripts/install_cuda_el8.sh
+++ b/.github/scripts/install_cuda_el8.sh
@@ -1,0 +1,173 @@
+# Install CUDA on Alma8/manylinux_2_28. 
+
+## -------------------
+## Constants
+## -------------------
+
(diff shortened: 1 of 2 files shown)


ptheywood:
Replaced by #1228

Issue #440, opened 2021-02-11
Ensemble Enhancements


Robadob:
These were left over from the PR, but none-blocking

* Import/export `RunPlanVec` to file ability.
* There was originally a plan to have code that estimates max concurrent models by estimating memory usage.
* If `CUDASimulation` errors out during ensemble, currently no information is captured. Might want a nice way of logging these errors.

Robadob:
Was thinking about `CUDAEnsemble` error reporting or lack thereof, may have thought of a suitable approach:
New `CUDAEnsemble::EnsembleConfig` parameter `error_level` or `fail_mode`.

* `0` or `None`: Runs which fail do not cause an exception to be raised. Failed runs must be probed manually via checking the return value of calls to `CUDAEnsemble::simulate()`. If `quiet` is set to true, runs will not be printed to `stderr`. (This is the current behaviour, however `CUDAEnsemble::simulate()` currently returns void.)
* `1` or `Slow`: If any runs fail, an `EnsembleException` will be raised after all runs have been attempted, before `CUDAEnsemble::simulate()` returns.
* `2` or `Fast`: An `EnsembleException` will be raised as soon as a failed run is detected, cancelling remaining runs.

Would need a decision on the most appropriate default error level, probably `1` imo.

Robadob:
> Was thinking about `CUDAEnsemble` error reporting or lack thereof, may have thought of a suitable approach: New `CUDAEnsemble::EnsembleConfig` parameter `error_level` or `fail_mode`.
> 
> * `0` or `None`: Runs which fail do not cause an exception to be raised. Failed runs must be probed manually via checking the return value of calls to `CUDAEnsemble::simulate()`. If `quiet` is set to true, runs will not be printed to `stderr`. (This is the current behaviour, however `CUDAEnsemble::simulate()` currently returns void.)
> * `1` or `Slow`: If any runs fail, an `EnsembleException` will be raised after all runs have been attempted, before `CUDAEnsemble::simulate()` returns.
> * `2` or `Fast`: An `EnsembleException` will be raised as soon as a failed run is detected, cancelling remaining runs.
> 
> Would need a decision on the most appropriate default error level, probably `1` imo.

This was addressed by https://github.com/FLAMEGPU/FLAMEGPU2/pull/839

PR #1230, opened 2024-09-21
RunPlanVector file IO

Robadob:
Currently just an (untested) minimal viable product

Still requires
- [ ] XML version
- [ ] Integration with `CUDAEnsemble` config/arg parsing
- [ ] Testing
- [ ]  Documentation

Relates #440 #1229


Robadob:
MVP now works, two issues to resolve
* Pretty print does not work (this is a bug that also requires addressing in `JSONStateWriter`)
* It needs adding to the SWIG interface, although I won't check that works beyond compiling.

Issue #1050, opened 2023-03-20
Missing Tests


Robadob:
Features that are not currently tested via the test suite:

* Agent Function (Condition) Macro DECL/DEF versions #1049
* Visualisation
* Random quality https://github.com/FLAMEGPU/FLAMEGPU2/issues/330

Robadob:
Doesn't appear we have full Graph IO tests (atleast not within the IO suite), although I've just added one to catch a bug found in #1361

PR #1232, opened 2024-09-21
BugFix: JSONStateWriter prettyprint output is nolonger minified.

### LINKED ISSUE #1231: [Bug] `JSONStateWriter` pretty print feature prints minified.
### Bug Description

`JSONStateWriter` casts the pretty writer pointer to a regular writer, I'm not totally sure why this compiles given `rapidjson::PrettyWriter` extends `rapidjson::Writer`, however none of the methods are virtual. Hence, once the pointer is cast, all calls hit `rapidjson::Writer` and skip the pretty print prefixes.

This bug is not present in `JSONLogger`, but due to how `JSONStateWriter` is structured for partial outputs, it will need a more complex solution (e.g. factory and templated class, or two internal writer pointers being tested at use 🤮).

### To Reproduce

Run `IOTtest.JSON_WriteRead` and breakpoint it after the file has been output (pretty print is enabled by default) and check `test.json` found in `tests/`.

### Expected Behaviour

Pretty print output should be across multiple lines with indentation.

### Additional Information

I'll probably write a fix tonight/tomorrow.
I'm not going to write a dedicated test for this, because that would require manually parsing the json and checking for newlines/tab, but I'm not that keen.

Robadob:
rapidjson's `PrettyWriter` extends `Writer`, but methods are non-virtual, leading to the bug which has been resolved.

This fix is the best solution of the bad options (e.g. heavy templating or dual internal writers/`std::variant` with an if/cast every time a method is called) given this implementation's modular approach to writing.

I have tested this works manually, I have no plans to write a test that parses the output to check whether it's minified or not as it would essentially be an inverse of the parsing rules I've written unless you had ground truths to compare against for a specific input. But you can if that's your prerogative.

`JSONLogger` does not suffer from the same bug, I have a feeling it was noticed when that was implemented and worked around.

Closes #1231

### DIFF
--- a/include/flamegpu/io/JSONStateWriter.h
+++ b/include/flamegpu/io/JSONStateWriter.h
@@ -61,9 +61,13 @@ class JSONStateWriter : public StateWriter {
     bool environment_written = false;
     bool macro_environment_written = false;
     bool agents_written = false;
+    // Dirty workaround for PrettyWriter overloads not being virtual
+    bool newline_purge_required = false;
     std::string outputPath;
     rapidjson::StringBuffer buffer;
-    std::unique_ptr<rapidjson::Writer<rapidjson::StringBuffer, rapidjson::UTF8<>, rapidjson::UTF8<>, rapidjson::CrtAllocator, rapidjson::kWriteNanAndInfFlag>> writer;
+    // Typedef because the template is too long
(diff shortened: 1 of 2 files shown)


ptheywood (approved):
A test would be prefferred but understandable why you haven't with our current time limitations. 

Maybe add it to #1050 so it can be added in the future if time allows? 

Otherwise looks sensible and appears to behave as intended under linix via a manual check of test.json.

mondus:
This might not support single quotes but is less broken than before so suggest to merge and raise a new bug if it comes up.

PR #1243, opened 2024-10-26
Graph visualisation would only work if CUDASimulation was explicitly init()

Robadob:
`applyConfig_derived()` is not implicitly called, I assumed otherwise. Hence, `CUDAEnvironmentDirectedGraphBuffers` was not receiving it's ptr to the visualisation.

Also noticed an edge-case with invalid characters in filepaths on Windows so improved the exception thrown.


Robadob:
Thinking about this, technically it's possible for a user to activate a visualisation after the simulation has been init, e.g. if manually stepping the model. So that call should also be updated to hook if the sim has been init I guess.

ptheywood (approved):
Looks good to me (I've not tested it myself).

Nice spot of the incorrect exception message in LoggerFactory too.

PR #1245, opened 2024-11-05
Cpplint 2 fixes

### LINKED ISSUE #1241: cpplint 2.x
cpplint 2.x has been released, with new rules and improvments to existing rules causing our lint CI to now fail. 

From a glance at the failures we should: 

- [ ] Prevent the <filesystem> being banned rule via CPPLINT.cfg, filesystem is useful.
- [ ] Prevent the intend in namespace rule via CPPLINT.cfg
- [ ] Probably fix the new include what you use errors

This will need applying to many repositories.

We could also pin to cpplint 1.x instead, but we'd also have to ammend our readme's to tell users its only 1.x compatible.

Alternatively we could switch to clang-format and see if it's any more stable (buit that would take more time and effort and installation used to be more of a faff than cpplint)

ptheywood:
+ [x] Suppress `<filesystem>` warnings
+ [x] Suppress namespace indenting warnings which are incorrect
+ [x] Fix some whitespace warnings
+ [x] Suppress whitespace warnings in macros that would result in a change of `__LINE__` behaviour in macros 
+ [x] Fix missing include warnings

Closes #1241

### DIFF
--- a/CPPLINT.cfg
+++ b/CPPLINT.cfg
@@ -1,3 +1,3 @@
 set noparent 
 linelength=320
-filter=-legal/copyright,-readability/todo,-runtime/references,-build/c++11
+filter=-legal/copyright,-readability/todo,-runtime/references,-build/c++11,-build/c++17
--- a/CPPLINT.cfg
+++ b/CPPLINT.cfg
@@ -1,3 +1,3 @@
 set noparent 
 linelength=320
-filter=-legal/copyright,-readability/todo,-runtime/references,-build/c++11,-build/c++17
+filter=-legal/copyright,-readability/todo,-runtime/references,-build/c++11,-build/c++17,-whitespace/indent_namespace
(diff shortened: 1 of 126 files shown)


mondus:
Added missing includes but error on my local machine 

`FLAMEGPU2\include\flamegpu/detail/cuda.cuh(121): error : identifier "cuCtxGetId" is undefined`

Which is odd as I am on CUDA 12.3. Pushing for now as I have run out of time. Might just be my system.

ptheywood:
12.0 and 12.3 on linux locally for me are fine with your changes, so assuming it's an msvc/your machine issue. 

More importantly `all_lint` passes now with cpplint 2, so assuming windows CI passes this is good to merge IMO.

@Robadob (or @mondus) will have to do the formal review on GH, as I'm unable to as the original PR author (though I can skip the checks anyway)

Robadob (approved):
Example builds for me locally (Windows, Debug, Vis=OFF, Seatbelts=ON, CUDA 12.1)

Lint flamegpu/tests both pass locally too.

PR #1262, opened 2024-11-26
Fix the FLAMEGPU_TELEMETRY_SUPPRESS_NOTICE CMake option.

plietar:
The option was defined using `cmake_dependent_option`, making it available if and only if `FLAMEGPU_SHARE_USAGE_STATISTICS` is true. Otherwise, its value is forced to false (or whatever was found in the environment).

This is the opposite of the expected behaviour. `SUPRESS_NOTICE` only makes sense when telemetry is disabled, i.e. `SHARE_USAGE_STATISTICS` is false.

This issue was making it impossible to suppress the notice through CMake configuration (other avenues still worked). Based on the examples in the [documentation for `cmake_dependent_option`][cmake_dependent_option], the list of conditions allows for negations, therefore using `NOT FLAMEGPU_SHARE_USAGE_STATISTICS` as the dependency fixes this.

[cmake_dependent_option]: https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html


Robadob (approved):
Thanks for noticing this and putting in a PR.

I've tested it, and confirmed that it now works as expected.

PR #1266, opened 2024-12-04
BugFix: DeviceAgentVector::setVariable() array element did not sync

Robadob:
Trivially, the _changed() flag was not being called like in other `setVariable()` methods

Test case added which would have caught this bug and for full arrays.

Thanks @ptheywood for spotting this bug, I don't think an issue was created.


ptheywood (approved):
Fixes the issue I encountered and mentioned on slack (and hand't made the issue yet.

Manylinux CI will fail until I get around to fixing it, so going to merge regardless

PR #1267, opened 2024-12-04
Fix error message when adding conflicting agent functions.

plietar:
The error message was attempting to print the name of both the agent and of the function being added, however it was accidentally using the agent's name twice instead.


Robadob (approved):

mondus (approved):

Robadob:
Thanks again!

PR #1268, opened 2024-12-04
Test Fix: Fix DependencyGraphTest.UnattachedFunctionWarning

ptheywood:
Fix test `DependencyGraphTest.UnattachedFunctionWarning`.

This test was broken by changing the exception message to include a newline char in 0a7456eacc2a755e41f4a3d608dae71a265cf252


Robadob (approved):
Not tested it, but looks reasonable.

(I only ran the specific suite when fixing that bug earlier)

PR #1277, opened 2025-03-05
Replace RapidJSON with nlohmann::json

Robadob:
RapidJSON has become outdated, which has led to increasing maintenance issues.

*Haven't yet had time to run test suite, but the static lib builds fine on Windows/Debug


ptheywood:
@mondus suggestion - check what values are used when loading a file which included NaN/inf etc or add a special case & document.

Robadob:
CI failing after rebase, appears to be a CUDA compiler issue which also affected GunRock.

https://github.com/gunrock/gunrock/issues/1059

Robadob:
I think I've addressed the requirements from our discussion a couple of months ago

- Resolved test, so it expects JSON reader to convert Inf to NaN
- Added a stderr warning when JSON being read contains NULL

Assigned to @ptheywood to decide whether we can drop pre CUDA 11.6 support on CI.

ptheywood:
> CI failing after rebase, appears to be a CUDA compiler issue which also affected GunRock's use of nhlomann (https://github.com/kokkos/mdspan/issues/97#issuecomment-1367262711).

> Assigned to @ptheywood to decide whether we can drop pre CUDA 11.6 support on CI. (See previous comment)

Surely this would mean dropping pre 11.6 support in general rather than just CI? 


If CI only, this would impact python users, changing the minimum CUDA (for non data center GPUs which have drivers that support enhanced compatability) to 11.6 rather than 11.2. 

I'd rather maintain 11.2+ support for now, but we do have some options to discuss

- Try and find a workaround (i.e. we don't use nhloman on device, so surely we can hide it from nvcc?)
- Bump to 11.6 everywhere
- Increse minimum / 11 builds to 11.8, so Hopper support is included (in wheels), without the need for PTX jitting
- Drop CUDA 11 support. 
  - We are broadly only planning to support 2 major CUDA versions anyway, i.e. when 13 releases (I'd guess within 6mo?) we would start thinking about dropping 11 anyway
    - But this would highlight the jitify perf issue more

Robadob:
Discussed, goal is to try to hide nlohmann from headers so CUDA never sees it.

Robadob:
Yeah the issue is the other way round that's prevented this before, cpp file including core flamegpu headers which includes cuda runtime API.

I've not tried ifdeffing that out everywhere, not sure if worth the effort.

ptheywood:
I've rebased this branch onto #1302, and removed the unsuccessful attempt to add CUDA <= 11.6 support (as it's being dropped in #1302 anyway). 

I've also tidied up a few references / comments about rapidjson which are had been missed.

Subject to CI, this should now be ready for a propper review (from me? including double checking the new nan/inf behavour) and then allow full CMake 4.0 support to be implemeneted on top. This will then join our chain of PRs waiting for #1150 to be mergable.

Tests pass locally under linux.

This will need to be included as a breaking change (along with all the other breaking changes in our next release)

src/flamegpu/io/JSONStateReader.cu
```diff
@@ -214,21 +213,42 @@ class JSONStateReader_impl : public rapidjson::BaseReaderHandler<rapidjson::UTF8
...
-    bool Double(double d) { return processValue<double>(d); }
-    bool String(const char*, rapidjson::SizeType, bool) {
+    bool null() {
+        bool is_array = false;
+        if (mode.top() == VariableArray) {
+            mode.pop();
+            is_array = true;
+        }
+        if (mode.top() == Environment) {
+            fprintf(stderr, "Warning: JSON Environment property '%s' contains NULL, this has been interpreted as NaN (but may represent Inf).\n", lastKey.c_str());
```
ptheywood, line 223: Ideally these warnings should only be emitted if verbosity is `> Quiet`. 

Should be fine, but will need to check when I implement it.

Issue #1276, opened 2025-03-05
CMake 4.0 Support


ptheywood:
CMake 4.0 (currently in a pre-release phase) removes compatibilty with CMake older than 3.5. 

Our minimum CMake is 3.18, however, RapidJSON which we fetch has an older minimum supported cmake (in the version we fetch) which is no longer supported, resulting in configuration errors (as @Robadob encountered)

![Image](https://github.com/user-attachments/assets/9040481a-6448-4597-b6d1-cf43f9ec32d2)

We could:

- Patch Rapidjson after we fetch it, bumping the minimum verison to 3.5. 
- Update the version of rapidjson we fetch, but that was non trivial last time we looked. 
- Replace Rapidjson with [nlohmann/json](https://github.com/nlohmann/json) or similar which has made an actual release since 2016 

There may be other CMake 4.0 changes required, and vis support should be checked too as we have a larger number of vis dependencies

ptheywood:
Manylinux images have now update to CMake 4.0, resulting in CI failures due to this warning. 

We can either temporarily fix ci by downgrading CMake / trying `-DCMAKE_POLICY_VERSION_MINIMUM=3.5` (or whichever version would work) or finish and merge #1277

ptheywood:
The installation of DevIL inside centors via camke also emits the similar warning, so will need to potentially bump / patch devil / install another way  / make sure older cmake is used for devil building. 

There have been no commits since 2021, so unlikely that we will be able to rely on this being fixed upstream any time soon.

PR #1280, opened 2025-04-22
Manylinux temporary fix: Downgrade CMake to 3.x via pipx in manylinux

ptheywood:
This avoids CMake errors introduced with CMake 4.0 related to 3rd party dependencies with cmake_minimum_required(3.5) or older

This is needed for rapdijson at configure time, and for visualiastion builds the el8 installation of DevIL from source via CMake.

This should be replaced by a CMake >= 4 compatible fix in the future.

See #1276


Robadob (approved):

Robadob:
Looks good assuming CI passes.

PR #1333, opened 2025-11-04
CI:  to improve the robustness of installing cuda on windows v…

### LINKED ISSUE #1332: CI: Make `redist` cuda windows installer more robust
The faster `redist` based CUDA installation script for windows CI runners occasionally fails during zip extraction, presumably due to the download being incomplete/corrupt.

This should be made more robust by:

- get the SHA for each zip from the json manifest
- After the download has completed, compare the SHA of file to the expected SHA.
- If they do not match, redownload.
- If extraction fails, redownload.
- If redownloading doesn't work, try again with exponetial backoff a couple of times. If it still fails, give up and error.

ptheywood:
Improve the robustness of installing CUDA on windows via redist

Installing indiviudal packages from the cuda redist would occasionally fail during zip extraction, most likely due to a corrupted or cancelled download.

This adds:

- Check the SHA256 is as expected before extraction
- Retry up to 3 times per download, if download, sha256 checking, or extraction fail
- Sleep for 10 seconds before trying again

Closes #1332

### DIFF
--- a/.github/scripts/install_cuda_windows_redist.ps1
+++ b/.github/scripts/install_cuda_windows_redist.ps1
@@ -111,15 +111,51 @@ Foreach ($package in $CUDA_PACKAGES) {
         echo "Warning: $package.$PLATFORM.relative_path is missing, unable to download"
         continue
     }
-    # Download the package.
+    # Download the package, retrying if the download or etraction fails, or the SHA did not match, up to a maximum number of times
+    $attempt = 0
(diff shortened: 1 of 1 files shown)


Robadob:
Worth testing then 👍

Robadob (approved):
No glaring issues (powershell isn't really my thing), and you've tested it thoroughly based on your comment.

mondus (approved):

PR #1334, opened 2025-11-05
Jitify2 occupancy API fix

ptheywood:
Jitify2 `jitify2::KernelData::function` returns a `jitify2::CudaFunction`, which is either a `CUkernel` or a `CUfunction` depending on if `JITIFY_USE_CONTEXT_INDEPENDENT_LOADING` is being used.

If `JITIFY_USE_CONTEXT_INDEPENDENT_LOADING` is not defined in advance, for CUDA >= 12 a `CUkernel` will be returned, otherwise a `CUfunction` is returned.

In our case, given a `jitify2::KernelData` for an agent function or condition, we want to pass the `CUfunction` to `cuOccupancyMaxPotentialBlockSize` to get the block size for the kernel launch. Jitify2 includes a method for this, but it configures the kernel for a grid-stride loop using the full device, where as we want to launch the minimum gridsize possible that includes at least one thread per agent (not using a grid stride loop).

In the ugprade from Jitify 1 to Jitify 2, we did not update out use of this to reflect the change in type. We missed this, as passing a `CUkernel` to `cuOccupancyMaxPotentialBlockSize` works as expected on recent CUDA drivers (i.e R575 and R580).

However, on systems with older CUDA drivers (R550) such as Google Colab and TUoS Stanage, uncaught cuda errors within `cuOccupancyMaxPotentialBlockSize` were leading to a division by zero floating point exceptions.

This is because `cuOccupancyMaxPotentialBlockSize` expects a `CUfunction`, but we were providing a `CUkernel` cast to a `CUfunction`. The documentation says that `cuLibraryGetKernel` should be used, to get a `CUkernel` from a `CUfunction`: 

> the API can also be used with context-less kernel CUkernel by querying the handle using cuLibraryGetKernel() and then passing it to the API by casting to CUfunction. Here, the context to use for calculations will be the current context.

To resolve this, I have:

- Added mising return value checks on occupancy API calls, for the runtime and driver API
  - This would have provided us with the error nicely, avoiding the difficult google colab debugging
  - More checks are probably missing for other CUDA API calls 
- Added an unnamed namespace function which given a `jitify2::KernelData` returns a `CUfunction` regardless of the value of `JITIFY_USE_CONTEXT_INDEPENDENT_LOADING`
  - This uses `if constexpr` in a c++20 templated lambda (to avoid both sides of a non-templated `if constexpr` needing to compile at the same time)

### DIFF
--- a/src/flamegpu/simulation/CUDASimulation.cu
+++ b/src/flamegpu/simulation/CUDASimulation.cu
@@ -57,6 +57,33 @@ namespace {
             return std::unique_ptr<detail::Timer>(new detail::SteadyClockTimer());
         }
     }
+
+    /**
+     * Given a jitify2::KernelData, gets the CUfunction for the current CUDA context, regardless of if context-independent loading is being used or not
+     *
+     * Jitify 2 may (or may not) be using context-independent loading, leading to `jitify2::KernelData::function` returning a `CUkernel` or a `CUfunction`.
+     * The Driver API method `cuOccupancyMaxPotentialBlockSize` expects a CUfunction, but is documented as "the API can also be used with context-less kernel CUkernel by querying the handle using cuLibraryGetKernel() and then passing it to the API by casting to CUfunction. Here, the context to use for calculations will be the current context."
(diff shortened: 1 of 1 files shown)


ptheywood:
Confirmed that with this fix the FLAME GPU tutorial on colab is now functional, and compilation error messages are visible.

PR #1340, opened 2025-11-11
Breaking Change Fix: JSONStateWriter writes scalars as scalars.

### LINKED ISSUE #1341: JSON state writer broken for scalar agent variables
Agent variables being written to json are always output as arrays rather than scalar values. E.g. "x": [123] rather than "x": 123. This may occur in other places within the state writing.

This may occur in other places within the JSOSN state writer.

Robadob:
Agent variables and environment properties have been fixed.

Macro environment properties are still always written as arrays.

This behaviour was accidentally changed when JSON library was replaced with nlohmann::json

Example output after fix: [end.json](https://github.com/user-attachments/files/23484967/end.json)

Fixes #1341.

### DIFF
--- a/src/flamegpu/io/JSONStateWriter.cu
+++ b/src/flamegpu/io/JSONStateWriter.cu
@@ -135,33 +135,34 @@ void JSONStateWriter::writeEnvironment(const std::shared_ptr<const detail::Envir
                     THROW exception::JSONError("Model contains environment property '%s' of unsupported type '%s', "
                         "in JSONStateWriter::writeEnvironment()\n", a.first.c_str(), a.second.type.name());
                 }
-            }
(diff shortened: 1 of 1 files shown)


Robadob:
> Changes look sensible, builds and no errors on linux.
> 
> We're clearly missing a test that checks the JSON (and probaly XML) produced are as intended seeing as this snuck through + no test change needed in fixing it either.
> 
> Probably best to get the fix in ASAP, and the missing tests can be added to #1050.

As noted in #1050, we have a test. However, our JSON reader is happy to parse scalars regardless of whether they're presented as scalar or an array length 1. Paul's tutorial mini parser was merely less robust and hence could not handle that change.

PR #1342, opened 2025-11-12
CUDA Architecture fixes and improvements

### LINKED ISSUE #1043: Improved CUDA ARCH error messages for CUDA 11.5+
[`flamegpu/detail/compute_capability.cuh`](https://github.com/FLAMEGPU/FLAMEGPU2/blob/master/include/flamegpu/detail/compute_capability.cuh) and associated .cpp and CMake code uses the minimum CUDA version selected at CMake configuration time to provide a more user friendly runtime error when a too old GPU is used. 
This is required as `__CUDA_ARCH__` is only available during the appropriate nvcc compilation phase for the single `compute_xy` arch.

CUDA 11.5 introduced a new virtual architecture macro, `__CUDA_ARCH_LIST__` which is available on host compilation passes (when invoked through nvcc, so .cu files), which contains a comma separated list of `xy0` for the `compute_xy` gencodes passed.

This could be used to improve error messages in use-cases where the minimum cuda arch is not sufficient (or atleast improve the error message itself when CUDA 11.5+ is used). It does not include the real archi (`code_xy`) however so is still not perfect. 

As it is a string defined in a macro without any quotes, need to stringify the macro to pass it to `printf` for instance

> For example, the following nvcc compilation command line will define `__CUDA_ARCH_LIST__` as `500,530,800` :
> 
> ```
> nvcc x.cu \
>    --generate-code arch=compute_80,code=sm_80 \
>    --generate-code arch=compute_50,code=sm_52 \
>    --generate-code arch=compute_50,code=sm_50 \
>    --generate-code arch=compute_53,code=sm_53
> ```

https://docs.nvidia.com/cuda/archive/11.5.0/cuda-compiler-driver-nvcc/index.html#virtual-architecture-macros

This is super low priority, just a feature I was previously unaware of.

### LINKED ISSUE #1284: Support letter-suffexed CUDA compute capabilities
In several places we (previously correctly) assume that device compute capabilities are numeric. 

Since Hopper, there are compute capabilities suffixed with `a` for architecture specific compilation. i.e. `sm_90a`. 

With Blackwell, family specific compute capibilities are introduced, suffixed with `f`. 

We should ensure that compiling for these targets is possible and behaves as intended (within reason, we don't have any blackwell to run on yet). 

In some places, the CUDA api assumes they are always integers however, i.e. `nvrtcGetSupportedArchs` which we use in `compute_capability::getNVRTCSupportedComputeCapabilties`· 

----

We may not need to make any changes, but should check if these are usable.

### LINKED ISSUE #1339: CUDA >= 12.8 incorrect `flamegpu::exception::InvalidCUDAComputeCapability` error with `all-major` (default)
By default, if `CMAKE_CUDA_ARCHITECTURES` is not defined (or populated from the `CUDAARCHS` env var) then `all-major` is used. 

For CUDA >= 12.8, if `all-major` is used and the code is executed on an SM < 100 device then users will encounter an error when running simulations:

```
terminate called after throwing an instance of 'flamegpu::exception::InvalidCUDAComputeCapability'
  what():  /home/ptheywood/code/flamegpu/FLAMEGPU2/src/flamegpu/simulation/CUDASimulation.cu(1552): Error application compiled for CUDA Compute Capability 100 and above. Device 0 is compute capability 86. Rebuild for SM_86.
```

This was first identified by @Robadob when doing a fresh build on a machine. 

The workaround for FLAME GPU <= 2.0.0rc.3 is to explicitly set `CMAKE_CUDA_ARCHITECTURES` or `CUDAARCHS` to a specific value (i.e. `86`), `native` or a list of architectures in ascending order to replicate `all-major` such as `"75-real;80-real;90-real;100-real;110-real;120-real;120-virtual"` for CUDA 13.0.

---

This is a bug in our (my) CMake for determining the `min_cuda_arch` value when `all-major` is used, which is returning `100` instead of `50`, as it is sorting lexicographically rather than numerically.

This will have been missed as I always set `CUDAARCHS` on my machines for faster build times and in #1274 I only checked explicit values not with `all` or `all-major`, and CI uses explicit lists of major architectures (although we could switch to all-major` for wheel production now we have dropped CUDA 11).

We cannot exclusively treat compute capabilities as numbers as `a` and `f` varaints exist, and `100-real` or `100-virtual` are also valid. 

Providing explciit values out of order behaves as intended (`100;90` correctly returns `90`).

Separately however `a` and `f` variants of architectures are explcitiyl blocked by our CMake due to the use of regular expressions against the nvcc help output, and then would also break any sorting used in that area too, so coudl also do with being fixed (I thought I had checked this but #1284 is still open so must not have done).

```
CMake Error at cmake/CUDAArchitectures.cmake:178 (message):
   CMAKE_CUDA_ARCHITECTURES value `90a` is not supported by nvcc 13.0.88.
   Supported architectures based on nvcc --help: 
     100;103;110;120;121;75;80;86;87;88;89;90
```

---

> [!IMPORTANT]
> **We might be able to avoid setting our own minimum via cmake processing now we have dropped CUDA < 11.5 support via #1043, but some investigation into `__CUDA_ARCH_LIST__` would be required for edge cases (it's documented as being sorted, but would need to check a/f variants and ptx-only builds)**

--- 

We don't have a way to write programatic tests for our CMake, but test cases to check would include:

| CMAKE_CUDA_ARCHITECTURES | CUDA | retval | 
|--------------------------|------|--------|
| `native`                 | any  | `0` (can't detect this without compiling and running code) |
| `all-major` \| `all` | 13.x | `75` | 
| `all-major` \| `all` | 12.x | `50` |
| `100;80` | >= 12.8 | `80` | 
| `80;100` | >= 12.8 | `80` |
| `90a;`   | >= 12.0? | `90`? (this cannot be ran on anything other than 90, so may not be valid as a 'minimum' anyway) |
| `100-real; 90-real;90-virtual`| >= 12.8 | `90` |

ptheywood:
Use `__CUDA_ARCH_LIST__` to fix/improve handling of cuda architectures

- CMake: Fix `all`/`all-major` `CMAKE_CUDA_ARCHITECTURES` for CUDA >= 12.8
  - Uses the first value from `__CUDA_ARCH_LIST__` as the minimum, rather than a value determined via CMake which was previosuly incorrect for all/all-major with CUDA >= 12.8
 - The new approach fixes the min cuda arch checking for `native`.
 - Closes #1339
- CMake: use `all-major` as the default all the time as minimum cmake supports it
- CMake: Allow `a`/`f` post-fixed cuda architectures
  - Fixes regular expressions in our custom cmake handling of cuda architectures 
  - Relaxes this from an error to a warning, in case of future changes
  - Emits the warning(s) once and only once per cmake invocation
  - configuration and compilation checked only, not executed on an 90a/10Xa/f device
  - Closes #1284
- Improve `InvalidCUDAComputeCapability` exception messages using `__CUDA_ARCH_LIST__`
  - Exception string now of the form: `Error application compiled for CUDA Compute Capabilities "90;100". Rebuild including compute capability <= 86 for device 0`
  - Closes #1043

### DIFF
--- a/include/flamegpu/detail/compute_capability.cuh
+++ b/include/flamegpu/detail/compute_capability.cuh
@@ -30,6 +30,13 @@ int getComputeCapability(int deviceIndex);
 int minimumCompiledComputeCapability();
 
 /**
+ * Get the compute capability integers used when building compute_capability.cu in a string for use in error messages
+ *
+ * @return string of compute capabilties used when compiling compute_capability.cu
+ */
+std::string compiledCompiledComputeCapabilitiesString();
+
+/**
  * Check that the current executable has been built with a low enough compute capability for the current device.
  * This assumes JIT support is enabled for future (major) architectures.
  * If __CUDA_ARCH_LIST__ could not be used to extract a minimum compiled architecture, no decision can be made so it is assumed to be successful.
--- a/src/flamegpu/detail/compute_capability.cu
+++ b/src/flamegpu/detail/compute_capability.cu
@@ -65,6 +65,30 @@ int compute_capability::minimumCompiledComputeCapability() {
     return 0;
 }
 
+std::string compute_capability::compiledCompiledComputeCapabilitiesString() {
+    // Get a std::array<int> containign the values from the macro, via another macro
+    #if defined(__CUDA_ARCH_LIST__)
+        // Macro wrapper for getting the __CUDA_ARCH_LIST__ as a std::array of int
+        #define MACRO_TO_ARRAY_WRAPPER() macro_to_array<int>(__CUDA_ARCH_LIST__)
+        auto archs = MACRO_TO_ARRAY_WRAPPER();
+        #undef MACRO_TO_ARRAY_WRAPPER
+        // Build a semi-colon separated string (as CMAKE_CUDA_ARCHITECTURES must be space-separated)
+        bool first = true;
+        std::string result = "";
+        for (const int& arch : archs) {
+            if (!first) {
+                result += ";";
+            }
+            // div by 10, as __CUDA_ARCH_LIST values have an extra 0 appended to them
+            result += std::to_string(arch / 10);
+            first = false;
+        }
+        return result;
+    #else
+        return std::array<int, 1>{0};
+    #endif
+}
+
 bool compute_capability::checkComputeCapability(int deviceIndex) {
     // If the compile time minimum architecture is defined, fetch the device's compute capability and check that the executable (probably) supports this device.
     if (getComputeCapability(deviceIndex) < minimumCompiledComputeCapability()) {
--- a/src/flamegpu/simulation/CUDASimulation.cu
+++ b/src/flamegpu/simulation/CUDASimulation.cu
@@ -1548,8 +1548,9 @@ void CUDASimulation::applyConfig_derived() {
     // Check the compute capability of the device, throw an exception if not valid for the executable.
     if (!detail::compute_capability::checkComputeCapability(static_cast<int>(config.device_id))) {
         int min_cc = detail::compute_capability::minimumCompiledComputeCapability();
+        std::string compiled_ccs = detail::compute_capability::compiledCompiledComputeCapabilitiesString();
         int cc = detail::compute_capability::getComputeCapability(static_cast<int>(config.device_id));
-        THROW exception::InvalidCUDAComputeCapability("Error application compiled for CUDA Compute Capability %d and above. Device %u is compute capability %d. Rebuild for SM_%d.", min_cc, config.device_id, cc, cc);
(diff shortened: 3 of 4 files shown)


Robadob (dismissed):
Works on the machine where bug was found.

Haven't reviewed the detailed changes.

ptheywood:
@Robadob encountered compiler warnings on windows, where getting the first item from a macro of comma separated architectures that works on linux does not work on windows :cry: 

Needs windows fixes.

image

Robadob:
> I don't believe I've done anything to force a 64-bit toolset or simlar, and it seems that vs2022 should use that by default anyway?

I think it's been that way for a long-time, though I still always manually set it out of habit within cmake.

PR #1350, opened 2025-11-18
BugFix: Resolve race condition during graph visualisation init.

Robadob:
Missing mutex lock when a graph is first linked to visualisation.

Reported by @DavidFletcherShef in https://github.com/FLAMEGPU/FLAMEGPU2-docs/pull/189#issuecomment-3543650107

*CI failures appear to be due to GitHub outage, 500 internal server error during repo fetches.*


ptheywood (approved):
Looks sensible

(I've not got the problem case to test this, and we don't have automated testing of vis)

PR #1361, opened 2026-02-23
BugFix: Graph loader would error if graph contained unknown array

Robadob:
Warning was not resetting array status in mode stack, leading to lost mode.

e.g. the below snippet, would cause the JSON loading to crash as the mode stack would be read when empty, if `UNEXPECTED_ARRAY` was not expected.

```json
...
  "links": [
    {
      "source": "1",
      "target": "2",
      "UNEXPECTED_ARRAY": [0, 1]
    },
...
```
Same applies to nodes too.

Possibly want to add a test case for this.


Robadob:
I've added a test case, it's a bit of a weird one as I think it's standard library undefined behaviour. So I'm testing it catches the bug on windows/linux in debug/release.

- Win Debug: Debug assertion
- Win Release: SEH exception
- Linux Debug & Release: Segfault (crashing the test suite)

Robadob:
I've tested the test-case pre and post-test on windows/linux debug/release, so it should be good to go.

PR #1366, opened 2026-03-17
WIP: Visualisation metadata sharing with simulation.

Robadob:
*In development as part of CrisisTrack, interface needs review.*

In order for simulations to appropriately scale models, this feature passes back limited visualisation metadata (e.g. orthographic projection zoom value, and view position/direction). It would be possible to do this in shader, but allowing it to be handled at a model level enables more complex scaling behaviours to be represented.

Current implementation works, and will suffice for CrisisTrack however I'm regretting the decision to hide it inside the host environment interface, as it won't be consistent with device environment.

- [ ] Protect against race conditions (they seem unimportant in this context, worst case a single simulation step sees a junk value)
- [ ] Finalise/refactor interface
- [ ] Fix CI (the current error seems spurious at first glance)
 

Matching visualiser PR: https://github.com/FLAMEGPU/FLAMEGPU2-visualiser/pull/152


Robadob:
I've also now added a bugfix that allows JSON Graph files to have links before nodes, as I hit this limitation exporting a graph from `nlohmann::json`.

Issue #1368, opened 2026-03-31
AMD RTC & pyflamegpu


ptheywood:
#1367 is explicitly not adding support for RTC or pyflamegpu support.

Once this is implemented, RTC agent function support and subsequently pyflamegpu support should be added. This will be non trivial due to the use of Jitify preventing a direct port.

- C++ RTC
    - [ ] Remove relevant compile-time guards for API functions, tests etc
    - [ ] Minimal replacment of Jitify for AMD, so the happy-path works
    - [ ] Environment variables equicvalent to `CUDA_HOME`/`CUDA_PATH`?
    - [ ] Better replacement of Jitify for AMD, with nice error messages etc
- Pyflamegpu
    - [ ] Remove `message(FATAL ERROR` preventing this combination
    - [ ] CMake so that the swig wrapper is compiled as hip
    - [ ] Choose how to handle hip wheels, `+hip70` or `+roc70` for a 7.0 build? etc
    - [ ] Test CI produced wheel on local AMD machine(s), 
    - [ ] Add AMD to manylinux CI workflow
    - [ ] Add AMD to draft release workflow

PR #1379, opened 2026-04-13
AMD GPU Support via HIP/ROCm

### LINKED ISSUE #1367: Initial AMD GPU Support (ROCm/HIP)
Add support for AMD GPUs via ROCm/HIP.

Initial AMD GPU support will be limited in scope:

- Linux only (Windows may work, but will not be tested)
- ROCM/HIP 7.x only (possibly expanding to include 6.x)
- Offline C++ only (no c++ RTC, no python)
- Limited host compiler support (amdclang shipped with ROCm only?)
- No visualisation support
- No MPI support
- No GLM support (may work, unliekly to be tested)
- ...

This will (almost certainly) result in breaking changes, both to the FLAME GPU API, and to CMake configuration options.

- [ ] Required CMake changes for general HIP support
- [x] CMake fatal errors on not-implemented optional features
- [ ] CI support 
    - if viable for free GitHub action runners, given the full rocm toolkit is 20GB, targetted sub-package installation should be possible for 7.x but unsure on the size of a minimum viable ROCm cmake workflow.
    - [Standard runners](https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories) have 14GB of storage available. 
    - [rocm/dev-ubuntu24.04](https://hub.docker.com/r/rocm/dev-ubuntu-24.04) is 6.9GB, which might be viable (but would not be suitable for future pythong support) 
    - Edit: This appears to be viable and relatively light to just install subpackages, but will need to test with the full FLAMEGPU build. Also possible to install on EL8 (via the RHEL instructions for alma) for ManyLinux_2_28 builds, although this will be left for the subsequent work to add pyflamegpu support for AMD
- [ ] Initial predominanlty macro-based CUDA/Rocm/hip abstraction layer, with as few changes as possible.
- [ ] Clang warning suppressions / fixes (I'll not fully test this with CUDA as the host compiler / include visualisation support, so not a full replacement for #936 but should cover a lot of it)

Subequent AMD GPU support would then include (to be defferred to separate issues at a later date):

- A more refined abstraction of CUDA/HIP differences
- Wider support of OS/Compilers
- Wider support of optional FLAME GPU features, including python wheels.

### LINKED ISSUE #1376: AMD + GLM
#1367 is not adding support for the experimental GLM support (via `FLAMEGPU_ENABLE_GLM`). 

This should be fine, but is being exlcuded from the inital AMD port to minimise complexity

### LINKED ISSUE #1377: AMD ROCTx support
#1367 is shaping up to include ROCTX support, but as it is a non-default feature there is a good chance I forget to test it actually behaves at runtime. 

This issue is mostly to ensure this does not get completely forgotten about. 

It may also be a good opportunity to improve the use and names of ranges, add a domain etc.

The initial port will also leave the CMake variable as `FLAMEGPU_ENABLE_NVTX`. This ideally should be renamed / deprecated with a generic name used instead. We can probably also remove support for NVTXv1, and only support the v3 interface (which should be the only version available for our supported CUDA versions)

ptheywood:
Adds AMD GPU support via HIP / ROCm

> [!CAUTION]
> 
> This is still a WIP - do not review, merge or expect CI to be happy
> 
> It will be rebased many times, and may become the base branch for other AMD/HIP/ROCm/clang related PRs so they can all be merged into `master` in a single go.

- Todo: Add the actual PR text
- Abstracts (most, some is not portable) HIP/CUDA macro usage into `flamegpu::detail::gpu` namespace.
    - This is also a step towards #647, as dlopen'd functions should only need to be within that file (excluding vis interop, but vis is CUDA only currently)
- Closes #1367
- Closes #1376 
- Closes #1377

## Decisions to make during review

- [x] `FLAMEGPU_GPU=CUDA/HIP/OFF` - Best I've come up with, but not super keen on this cmake option name.
    - [x] support lower case values as well (i.e. make it upper case before validation)?
   - [x] BACKEND, VENDOR (prolyl not), PLATFORM, API
- [x] `FLAMEGPU_CURAND_ENGINE` - deprecate and rename? Its a bit grim to do with it being a cache var not an option.
- [ ] Ensure consistency in use of new macros? i.e. `defined(FLAMEGPU_USE_CUDA)` vs `ifdef FLAMEGPU_USE_CUDA`. Maybe do HIP first and fallback to the CUDA version so intellisense doesn't get offended? 
- [ ] Update Docs repo accordingly
- [x] `TODO: DO NOT MERGE` in test suite, where wrapper function changes happened but I didn't update the test suite. 
- [ ] Still don't understand why the wrapper helper is needed for agent functions but not for agent function conditions.... 
- [ ] Fully udpated changelog from this PR (more breaking changes / deprecations not yet included)
- [x] Telemetry NVCCVersion. Or maybe use a diff field name anyway? 

--- 

As of 2026-06-10 tests pass with:

- CUDA 13.2 on linux
- HIP 7.0 on linux
- CUDA 12.4 on windows


jeffdaily:
*Note: this reply was drafted by an AI assistant.*

We have been building and testing this branch on a few AMD GPUs and would like to offer the results, plus two small fixes, in case they are useful.

At 5e42a64 the Linux HIP build is clean out of the box and the test suite passes on an AMD Instinct MI250X (gfx90a, CDNA2, wavefront 64) with ROCm 7.14: 1070 passed, 64 skipped, 0 failed. The 64 skips are the runtime-compilation tests. boids_bruteforce runs to completion.

An earlier revision of the branch was also tested on three more GPUs, which between them cover both wavefront sizes and both operating systems:

- AMD Radeon Pro W7800 (gfx1100, RDNA3, wavefront 32), Linux, ROCm 7.2.1: 1069 passed, 64 skipped, 0 failed.
- AMD Radeon RX 9070 XT (gfx1201, RDNA4), Windows, ROCm 7.14 nightly: 1058 passed, 64 skipped, 11 failed.
- AMD Radeon PRO V710 (gfx1101, RDNA3), Windows, ROCm 7.14 nightly: 1049 passed, 59 skipped, 11 failed.

The 11 Windows failures are all in TestCUDASimulationConcurrency, which asserts a 1.5x speedup from running agent functions concurrently across streams; the measured speedup there was about 1.0x. Every functional test passes on both Windows GPUs. We have not characterised the cause and are not drawing any broader conclusion from it, but we are happy to look into it separately if that would help.

We have two fixes on a branch cut from 5e42a64, at https://github.com/AMD-Ecosystem/FLAMEGPU2/tree/amdgpu-fixes:

The first deletes the copy and move operations on CUDAEventTimer. It creates two events in its constructor and destroys them in its destructor, but is implicitly copyable, so a copy would leave two objects owning the same handles and the second destructor would throw. Nothing in the tree copies one today, so this is latent; deleting the four operations turns any future copy into a compile error.

The second adds a guarded windows.h include to CUDAEnsemble.cu. It calls SetThreadExecutionState in three `#ifdef _MSC_VER` blocks without including the header. The CUDA build picks the declaration up transitively, but the HIP build on Windows does not and the file will not compile. AbstractSimRunner.cu, SimLogger.cu and MPISimRunner.cu all include it directly already.

Would you prefer these as a PR against amdgpu, or would you rather just cherry-pick them? Either suits us. We are also glad to rerun the suite on any of these GPUs whenever it would be useful to you.

ptheywood:
@jeffdaily Thank you for testing this out on your systems, especially on Windows as we had ruled that out supporting that initially due to not having a machine to test that on. It's good to know that it only requires the missing include to be functional. 
We also haven't tested 7.14 yet locally/in CI due to the installation process change (#1402) so that is very helpful too. 

The 64 skipped tests are expected, predominantly as we have not implemented runtime compilation on AMD devices yet (#1368, we will be adding this in the future), and some tests depend on other CMake configuration options being enabled.

> The 11 Windows failures are all in TestCUDASimulationConcurrency, which asserts a 1.5x speedup from running agent functions concurrently across streams; the measured speedup there was about 1.0x. Every functional test passes on both Windows GPUs. We have not characterised the cause and are not drawing any broader conclusion from it, but we are happy to look into it separately if that would help.

As you've highlighted, these tests are to check that execution using multiple streams do actually run concurrently, as we've previously unintentionally broken this functionality. Given the linux AMD tests pass, I assume this is related to the WDDM driver implementation? Previously we've had to work around issues with the NVIDIA WDDM driver and how the command buffer flushing interacts with compute and data transfer. For now I think we're OK with this being a known issue for windows / AMD and we can look into this further once we have time and a Windows AMD system set up.


> Would you prefer these as a PR against amdgpu, or would you rather just cherry-pick them? Either suits us. We are also glad to rerun the suite on any of these GPUs whenever it would be useful to you.

Those fixes both sound useful. A PR against amdgpu woudl be great. As FLAME GPU 2 is duel licencesed (AGPL + commercial) we do have a [CLA](https://github.com/FLAMEGPU/FLAMEGPU2/blob/master/CONTRIBUTING.md#contributor-license-agreement-cla) which a bot should prompt you about. Hopefulyl that is not an issue? 

--- 

Lastly, the only real issue we've had with the HIP/ROCm implementation is that for our agent function kernels (launching the user-defined device function) `hipOccupancyMaxPotentialBlockSize` would hang in debug or sigtrap in release, so we have fixed the block size to `128` for all kernel launches on AMD for now to get something out which runs.

In a standalone repository (and for some of the more typical kernels / launches in FLAME GPU 2) this behaves as you would expect, but would not work for the main place the occupancy API would be useful (user-defined code where resource limitations may prevent launches at a given block size). 
The kernel is templated with a user-specified device function (which is also templated) all wrapped up in a macro for user-convenience, so it is pretty grim. I tried mitigating this a few ways (storing the address of a host function which launches the kernel rather than the kernal address, moving the use of the occupancy api into the macro so its in the same compilation unit, etc.) but could not find a workaround. Do you have any ideas about this? There's more information and a branch which still exhibits the hangs in #1387, so it may be better to discuss there.

PR #1390, opened 2026-05-06
Sub models

IreneTerrone:
- [x] Create a generic submodel for movement
- [x] test it on a sugarscape model 
- [x] test it on beesandflower model
- [x] create unit tests for c++
- [x] create unit tests for python
- [x] export the new functionalities to python
- [x] create a new sugarscape model in python and test it


examples/cpp/beesandflowers/src/main.cu
```diff
@@ -0,0 +1,275 @@
+#include <iostream>
+#include <fstream>
+#include <vector>
+#include <algorithm>
+#include <numeric>
+#include <random>
+#include "flamegpu/flamegpu.h"
+#include "flamegpu/stockAgent/subModels/SingleAgentDiscreteMovement.h"
```
ptheywood: Do we want to update `flamegpu/flamegpu.h` to implicitly include the stock submodels header so this is not neccessary? Convinient but will slightly increase compile time (when unused).

examples/cpp/beesandflowers/src/main.cu
```diff
@@ -0,0 +1,275 @@
...
+    bee.newVariable<int>("last_resources_x", -1);
+    bee.newVariable<int>("last_resources_y", -1);
+    bee.newVariable<float>("hunger_level");
+    bee.newVariable<int>("wait", 0);
+    bee.newVariable<float>("priority", 0.0f);
+    bee.newVariable<float>("current_cell_score", 0.0f);
+
+
+    // Declared on the stack - uses empty constructor
+    flamegpu::stockAgent::submodels::SingleAgentDiscreteMovement move_sub_logic;
```
ptheywood: I'd lean towards a slightly different set of namespaces for this? 

`stock` instead of `stockAgent`?

This would need updating in many places (filepaths, include statments and namespace statements), so worth agreeing on the name before making any changes to avoid duplicate effort

examples/cpp/beesandflowers/src/main.cu
```diff
@@ -0,0 +1,275 @@
...
+
+    // Initialize the submodel
+    move_sub_logic.addSingleAgentDiscreteMovementSubmodel(model, ENV_DIM, ENV_DIM);
+
+    // Bind parent agents to submodel
+    move_sub_logic.setMovingAgent("bee",
+        {
+            {"x", "x"},
```
ptheywood, line 191: For automapped variables, might be nice to include that in the example but still show it without. 

Need to more thoroughly review the underlying code (to follow up).

examples/cpp/beesandflowers/src/main.cu
```diff
@@ -0,0 +1,275 @@
...
+            {"y", "y"},
+            {"last_x", "last_x"},
```
ptheywood: Potentially could make `last_x`/`last_y` internal to the submodel. Internal submodel variables might (probably?) be reset on subsequent steps of the outer model, which may be a problem (needs investigation, we will follow up)

src/flamegpu/stockAgent/subModels/SingleAgentDiscreteMovement.cu
```diff
@@ -0,0 +1,427 @@
...
+
+flamegpu::SubModelDescription SingleAgentDiscreteMovement::addSingleAgentDiscreteMovementSubmodel(ModelDescription &model, int ENV_SIZE_X, int ENV_SIZE_Y) {
+    ModelDescription sub_model_move("movement_submodel");
+
+    sub_model_move.Environment().newProperty<int>("submodel_env_width", ENV_SIZE_X);
+    sub_model_move.Environment().newProperty<int>("submodel_env_height", ENV_SIZE_Y);
+
+
+    AgentDescription movingAgent = sub_model_move.newAgent(INTERNAL_MOVING_AGENT_NAME);
+    movingAgent.newState("active");
```
ptheywood: Only a single state is used in the submodel, so simpler/safer to just leave this out and use the default state.

examples/cpp/beesandflowers/src/main.cu
```diff
@@ -0,0 +1,275 @@
...
+    bee.newVariable<int>("wait", 0);
+    bee.newVariable<float>("priority", 0.0f);
+    bee.newVariable<float>("current_cell_score", 0.0f);
+
+
+    // Declared on the stack - uses empty constructor
+    flamegpu::stockAgent::submodels::SingleAgentDiscreteMovement move_sub_logic;
+
+    // Initialize the submodel
+    move_sub_logic.addSingleAgentDiscreteMovementSubmodel(model, ENV_DIM, ENV_DIM);
```
ptheywood: `addSingleAgentDiscreteMovementSubmodel` could be renamed to something more generic, or could be moved onto the Ctor (as long as no exceptions are raised in this method etc, and we don't need a default ctor for these)

examples/cpp/beesandflowers/src/main.cu
```diff
@@ -0,0 +1,275 @@
+#include <iostream>
```
Robadob, line 1: I realised you've been developing this on a headless box, but we probably want to include an option for a live visualisation as with all other serious examples.

examples/cpp/beesandflowers/src/main.cu
```diff
@@ -0,0 +1,275 @@
...
+    agents_log << "step,id,x,y,hunger_level,wait" << std::endl;
+}
+
+FLAMEGPU_STEP_FUNCTION(stepLogger) {
+    auto bees = FLAMEGPU->agent("bee");
+    auto& bee_pop = bees.getPopulationData();
+    unsigned int step = FLAMEGPU->getStepCounter();
+
+    int count = 0;
+    for (const auto& bee : bee_pop) {
```
Robadob: I would prefer this to be using FLAMEGPU API, albeit you would be exporting entire agent state each step based on current limitations (perhaps a suggested future feature if this is evidence that it would be useful).

examples/cpp/beesandflowers/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+
+# Set the location of the ROOT flame gpu project relative to this CMakeList.txt
+get_filename_component(FLAMEGPU_ROOT ${CMAKE_CURRENT_SOURCE_DIR}/../../.. REALPATH)
+
+# Handle CMAKE_CUDA_ARCHITECTURES gracefully
+include(${FLAMEGPU_ROOT}/cmake/CUDAArchitectures.cmake)
+flamegpu_init_cuda_architectures(PROJECT beesandflowers)
+
+# Name the project and enable required languages
+project(beesandflowers CXX CUDA)
```
ptheywood, line 12: CMake changes will be required play nicely with #1379 (AMD GPU support). 

Right now it's correct, but depending on merge order we'll need to update this. 

For future reference this is probaly:

- no more CUDAArchitectures.cmake
- project should just depend on CXX
- need to call `flamegpu_enable_languages` to conditionally enable CUDA or HIP (after including common.cmake)
- Maybe some other bits.

examples/cpp/sugarscape_submodel_api/src/main.cu
```diff
@@ -0,0 +1,260 @@
+#include <iostream>
```
ptheywood, line 1: Probabyl want to rename the example directory to be consistent withour other exampels which are all lower snake case? 

I.e. `sugarscape_submodel_api` or similar?

examples/cpp/sugarscapeSubModelAPI/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+
+# Set the location of the ROOT flame gpu project relative to this CMakeList.txt
+get_filename_component(FLAMEGPU_ROOT ${CMAKE_CURRENT_SOURCE_DIR}/../../.. REALPATH)
+
+# Handle CMAKE_CUDA_ARCHITECTURES gracefully
+include(${FLAMEGPU_ROOT}/cmake/CUDAArchitectures.cmake)
+flamegpu_init_cuda_architectures(PROJECT sugarscapeAI)
+
+# Name the project and enable required languages
+project(sugarscapeAI CXX CUDA)
```
ptheywood: `sugarscape_submodel_api` or `sugarscape_api` instead? (i.e. match the directory name when changed, or add the missing `p`)

examples/cpp/sugarscape_submodel_api/visualizesimulationsugar.py
```diff
@@ -0,0 +1,156 @@
+import matplotlib.pyplot as plt
```
ptheywood, line 1: Given this has non-pyflamegpu dependencies (matplotlib), do we need to setup a requirements.txt for these? Thoguh not sure the best way to add pyflamegpu to that? 

Or do we just add matplotlib to the venv created during build? which is not neccessary a lot of the time? 

Unsure on the best way to handle this.

include/flamegpu/stockAgent/subModels/SingleAgentDiscreteMovement.h
```diff
@@ -0,0 +1,74 @@
+#ifndef INCLUDE_FLAMEGPU_STOCKAGENT_SUBMODELS_SINGLEAGENTDISCRETEMOVEMENT_H_
```
ptheywood: Path probably wants to be changed, which will then cause a re-lint to be required  for due to the include guarding (suggested rename in another comment)

examples/cpp/beesandflowers/src/main.cu
```diff
@@ -0,0 +1,275 @@
...
+    model.addExitFunction(exitLog);
+}
+
+int main(int argc, const char ** argv) {
+    ModelDescription model("OneAgentMovingModel");
+
+    define_model(model);
+
+    CUDASimulation simulation(model);
+    simulation.SimulationConfig().steps = SIMULATION_STEPS;
```
Robadob: Model should probably be init with argv to control the steps.

examples/cpp/beesandflowers/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+
+# Add the executable and set required flags for the target
+flamegpu_add_executable("${PROJECT_NAME}" "${ALL_SRC}" "${FLAMEGPU_ROOT}" "${PROJECT_BINARY_DIR}" TRUE)
+
+# Also set as startup project (if top level project)
+set_property(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"  PROPERTY VS_STARTUP_PROJECT "${PROJECT_NAME}")
+
+# Set the default (visual studio) debug working directory and args
+set_target_properties("${PROJECT_NAME}" PROPERTIES VS_DEBUGGER_WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
+                                                   VS_DEBUGGER_COMMAND_ARGUMENTS "-s 10")
```
Robadob: As model stands (although i've suggested above it changes), passing default args is redundant.

examples/cpp/sugarscapeSubModelAPI/src/main.cu
```diff
@@ -0,0 +1,260 @@
...
+            {"x", "x"},
+            {"y", "y"},
+            {"last_x", "last_x"},
+            {"last_y", "last_y"},
+            {"last_resources_x", "last_resources_x"},
+            {"last_resources_y", "last_resources_y"},
+            {"current_cell_score", "current_cell_score"}
+        },
+        {
+            {"active", flamegpu::ModelData::DEFAULT_STATE}
```
Robadob: If this is redundant, perhaps include it but commented out, with an additional comment to state "Default state is automatically used" or similar.

include/flamegpu/stockAgent/subModels/SingleAgentDiscreteMovement.h
```diff
@@ -0,0 +1,74 @@
...
+    SingleAgentDiscreteMovement() = default;
+
+    /**
+     * Defines the movement submodel and adds it to the provided parent model.
+     * @param model The parent ModelDescription
+     * @param ENV_SIZE_X Width of the environment
+     * @param ENV_SIZE_Y Height of the environment
+     * @return The created SubModelDescription
+     */
+    flamegpu::SubModelDescription addSingleAgentDiscreteMovementSubmodel(flamegpu::ModelDescription &model, int ENV_SIZE_X, int ENV_SIZE_Y);
```
ptheywood: `addSingleAgentDiscreteMovementSubmodel` only exists for the `SingleAgentDiscreteMovement` class, so it does not need as verbose of a name? 

Could this just be the constructor instead of having a defaulted constructor? Although the ctor for each class derived from AbstractSubmodel would then be a little bit different (which is probably fine)

examples/cpp/sugarscape_submodel_api/src/main.cu
```diff
@@ -0,0 +1,260 @@
...
+    }
+
+    model.addStepFunction(step_logger);
+
+    /**
+     * Simulation Setup
+     */
+    flamegpu::CUDASimulation cudaSimulation(model);
+    cudaSimulation.initialise(argc, argv);
+    cudaSimulation.SimulationConfig().steps = 100;
```
Robadob, line 177:
```suggestion
```

include/flamegpu/stock/subModels/SingleAgentDiscreteMovement.h
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * Defines the movement submodel and adds it to the provided parent model.
+     * @param model The parent ModelDescription
+     * @param ENV_SIZE_X Width of the environment
+     * @param ENV_SIZE_Y Height of the environment
+     * @return The created SubModelDescription
+     */
+    flamegpu::SubModelDescription addSingleAgentDiscreteMovementSubmodel(flamegpu::ModelDescription &model, int ENV_SIZE_X, int ENV_SIZE_Y);
+
+    /**
```
ptheywood, line 32: We probably want full docstrings prior to merging (`@param` etc)? but maybe leave that till we've finished proposing changes to the API itself to avoid redundant work.

examples/cpp/sugarscape_submodel_api/src/main.cu
```diff
@@ -0,0 +1,260 @@
...
+        }
+
+        cudaSimulation.setPopulationData(bug_pop);
+        cudaSimulation.setPopulationData(cell_pop);
+    }
+
+    cudaSimulation.simulate();
+
+    return 0;
+}
```
Robadob, line 308: I would have preferred this model to be a minimal rewrite of the existing sugarscape (or vice versa), to allow a more direct comparison of usage with/without the portable submodel.

src/flamegpu/stock/subModels/SingleAgentDiscreteMovement.cu
```diff
@@ -0,0 +1,427 @@
+#include "flamegpu/stockAgent/subModels/SingleAgentDiscreteMovement.h"
+
+#include <map>
+#include <string>
+#include <memory>
+#include <vector>
+
+#include "flamegpu/flamegpu.h"
+
+using std::string;
```
ptheywood, line 10: I'm surprised cpplint is not offended by these, I'm sure it used to be.

src/CMakeLists.txt
```diff
@@ -299,6 +299,8 @@ SET(SRC_INCLUDE
     ${FLAMEGPU_ROOT}/include/flamegpu/detail/Timer.h
     ${FLAMEGPU_ROOT}/include/flamegpu/detail/TestSuiteTelemetry.h
     ${FLAMEGPU_ROOT}/include/flamegpu/detail/JitifyCache.h
+    ${FLAMEGPU_ROOT}/include/flamegpu/stockAgent/subModels/AbstractSubmodels.h
+    ${FLAMEGPU_ROOT}/include/flamegpu/stockAgent/subModels/SingleAgentDiscreteMovement.h
```
ptheywood: Do we want a header which includes all the library submodels as well? (though there's onlt the one currently)

I.e. `include/stock/submodels.h` or somethign along those lines? 

This probably depends if we include it in the core flamegpu.h or not implicitly. 

Convinience is probably to include it in `flamegpu.h`, but for compile time keeping the total amount of header included in every flamegpu model might be nice

swig/python/flamegpu.i
```diff
@@ -42,10 +42,13 @@
 %{
 // Include the main library header, that should subsequently make all other required (public) headers available.
 #include "flamegpu/flamegpu.h"
-// Also include TestSuiteTelemetyr header, which is not intended to be public. 
+// Also include TestSuiteTelemetyr header, which is not intended to be public.
 #include "flamegpu/detail/TestSuiteTelemetry.h"
+#include "flamegpu/stockAgent/subModels/AbstractSubmodels.h"
```
ptheywood: Probably don't need to wrap the abstract class? (I don't think we're currently planning to support python-defined reusable submodels which are part of the core library, / a pure python version would probably be better anyway for that case). 

Though deleting this include might also break things / be implicit anyway from the derived class header.

.gitignore
```diff
@@ -500,3 +500,13 @@ MigrationBackup/
 
 # Ionide (cross platform F# VS Code tools) working folder
 .ionide/
+
+
+#Temporary for this branch
```
ptheywood: Do the temporary bits still need to be kept? 

For reference, `.git/info/exclude` is essentially a per-clone `.gitignore` where you can have your own ignore rules that won't be comitied. 

Images and csv's being ignored makes sense though.

CMakeLists.txt
```diff
@@ -30,9 +30,9 @@ set(DOCUMENTATION_ONLY_BUILD OFF)
 
 # Check compiler functionailty, as there are known issues in some cases, but version checks are not always sufficient.
 include(./cmake/CheckCompilerFunctionality.cmake)
-# If this returned a negative result, set the docs only build. 
+# If this returned a negative result, set the docs only build.
 if(NOT FLAMEGPU_CheckCompilerFunctionality_RESULT)
-    set(DOCUMENTATION_ONLY_BUILD ON)      
+    set(DOCUMENTATION_ONLY_BUILD ON)
```
ptheywood, line 35: White space trims are going to cause conflicts with #1379 for atleast this line.

ptheywood (changes requested):

examples/python_rtc/sugarscapeModel/sugarscape.py
```diff
@@ -0,0 +1,246 @@
...
+
+    my_step = step_logger()
+    model.addStepFunction(my_step)
+
+    # Register the random initialisation
+    model.addInitFunction(random_initialisation())
+
+    # Set up and run the simulation
+    cudaSimulation = pyflamegpu.CUDASimulation(model)
+    cudaSimulation.SimulationConfig().steps = 100
```
Robadob, line 216: Init from argv instead.

Robadob (changes requested):
Big Stuff:
- Examples should init from argc/argv by default.
- Namespace should be `stock` rather than `stockAgent`
- Ideally examples should provide the option of FLAMEGPU visualisation

Otherwise broadly good. I haven't had a chance to nitpick the API yet, but it's close regardless.

examples/cpp/beesandflowers/src/main.cu
ptheywood: ignore this,

examples/cpp/sugarscape_submodel_api/visualizesimulationsugar.py
ptheywood: add a `examples/cpp/sugarscapeSubModelAPI/requirements.txt` with matplotlib in the example (changed for renaemed dir)

src/CMakeLists.txt
ptheywood: no.

examples/cpp/beesandflowers/src/main.cu
ptheywood: e.g. 

https://github.com/FLAMEGPU/FLAMEGPU2/blob/1a11aea7f7620825233810f2410b08bca0fd6e4e/examples/cpp/sugarscape/src/main.cu#L366-L394

https://github.com/FLAMEGPU/FLAMEGPU2/blob/1a11aea7f7620825233810f2410b08bca0fd6e4e/examples/cpp/sugarscape/src/main.cu#L470-L474

https://github.com/FLAMEGPU/FLAMEGPU2/blob/1a11aea7f7620825233810f2410b08bca0fd6e4e/examples/cpp/sugarscape/src/main.cu#L491-L493

examples/cpp/beesandflowers/src/main.cu
IreneTerrone: The variables internal to the submodel are reset at every iteration. We could leave the behaviour optional: if the user wants to use last_y and last_x, they can explicitly map them. Otherwise, he doesn't and uses the default behaviour (doesn't track the last position). For now, I'm doing it like this.

Robadob: That seems fine, long-term we might want to make the reset behaviour optional if re-usable submodels create good use-cases.

fleming0:
I've had a chance to play with some of the subModel functionality while porting some agent-based models from NetLogo, specifically ones focused on discrete movement. It would be useful if I could control how the cell scores are calculated / broadcast and the submodel handles movement resolution _only_. This way, there is room for extension to other discrete movement models besides Sugarscape.

Essentially:

- Make [`envAgent_broadcast_status`](https://github.com/FLAMEGPU/FLAMEGPU2/blob/subModels/src/flamegpu/stock/subModels/SingleAgentDiscreteMovement.cu#L25) broadcast a custom message type supplied by the caller to the submodel.
- Separate [`movingAgent_request_move`](https://github.com/FLAMEGPU/FLAMEGPU2/blob/subModels/src/flamegpu/stock/subModels/SingleAgentDiscreteMovement.cu#L40) so, _instead_, callers supply an agent function that broadcasts the desired neighbouring cell the agent wishes to move to, and the submodel handles movement bidding and resolution.

And the steps of the movement submodel become:

1. Environment agents broadcast a custom data message (supplied by the caller)
2. Moving agents read this message and broadcast their desired cell (function supplied by caller)
3. Environment agent resolves movement conflicts based on each moving agents desired cell
4. Moving agents check to see if they won their target cell and moves if successful
5. Environment agents update their occupancy status

For example, the current submodel functionality looks like:
```python
model = pyflamegpu.ModelDescription("sugarscape")

bug = model.newAgent("bug")

bug.newVariableInt("x")
bug.newVariableInt("y")
bug.newVariableFloat("sugar")
bug.newVariableFloat("metabolism")
bug.newVariableInt("last_x")
bug.newVariableInt("last_y")
bug.newVariableInt("last_resources_x")
bug.newVariableInt("last_resources_y")
bug.newVariableFloat("current_cell_score")

sugar_cell = model.newAgent("sugar_cell")

sugar_cell.newVariableInt("x")
sugar_cell.newVariableInt("y")
sugar_cell.newVariableFloat("sugar") # mapped to 'cell_score' in submodel
sugar_cell.newVariableFloat("max_sugar")
sugar_cell.newVariableInt("is_occupied")

submodel = pyflamegpu.SingleAgentDiscreteMovement()
move_sub_desc = submodel.addSingleAgentDiscreteMovementSubmodel(model, GRID_HEIGHT, GRID_WIDTH)
```

Whereas an alternative API could possibly look like:

```python
model = pyflamegpu.ModelDescription("discrete_agents")

moving_agent = model.newAgent("agent")

moving_agent.newVariableInt("x")
moving_agent.newVariableInt("y")
moving_agent.newVariableInt("last_x")
moving_agent.newVariableInt("last_y")
moving_agent.newVariableInt("last_resources_x")
moving_agent.newVariableInt("last_resources_y")
moving_agent.newVariableFloat("current_cell_score")

cell = model.newAgent("cell")

cell.newVariableInt("x")
cell.newVariableInt("y")
cell.newVariableFloat("cell_score")
cell.newVariableFloat("max_cell_score")
cell.newVariableInt("is_occupied")

@pyflamegpu.agent_function
def pickCell(
    message_in: pyflamegpu.MessageArray2D, message_out: pyflamegpu.MessageArray2D
):
    # Code that handles which cell the moving agent wants to move to dependent on 
    # the custom data received from the environment cell
    return pyflamegpu.ALIVE

# Messaging Protocol for Cell Broadcasts
cell_broadcast_msg_arr2d = model.newMessageArray2D("cell_broadcast_msg_arr2d")
# Custom data in here:
# cell_broadcast_msg_arr2d.newVariableInt(...)
# cell_broadcast_msg_arr2d.newVariableChar(...)
# cell_broadcast_msg_arr2d.setPersistent(...)
cell_broadcast_msg_arr2d.setDimensions(GRID_HEIGHT, GRID_WIDTH)

# Messaging Protocol for Agent Broadcasts
agent_broadcast_msg_arr2d = model.newMessageArray2D("agent_broadcast_msg_arr2d")
# Broadcasts the desired target cell
agent_broadcast_msg_arr2d.newVariableInt("target_x")
agent_broadcast_msg_arr2d.newVariableInt("target_y")
agent_broadcast_msg_arr2d.setDimensions(GRID_HEIGHT, GRID_WIDTH)

# Configure cell selection fn
agent_cell_selection_fn = buyer.newRTCFunction(
    "agent_cell_selection_fn", pyflamegpu.codegen.translate(pickCell)
)

agent_cell_selection_fn.setMessageInput("cell_broadcast_msg_arr2d")
agent_cell_selection_fn.setMessageOutput("agent_broadcast_msg_arr2d")

submodel = pyflamegpu.SingleAgentDiscreteMovement()
submodel.setCellBroadcastMessage(cell_broadcast_msg_arr2d)
submodel.setAgentBroadcastMessage(agent_broadcast_msg_arr2d)
submodel.setAgentCellSelectionFunction(agent_cell_selection_fn)

move_sub_desc = submodel.addSingleAgentDiscreteMovementSubmodel(
    model, GRID_HEIGHT, GRID_WIDTH
)
```

By moving the cell selection functionality and message types outside of the submodel it should be possible to capture discrete movement models that are different to Sugarscape. I.e the *act* of discrete movement is common across NetLogo (and possibly many other) models but deciding *where* to move is not. This alternative API might help with this.

tl;dr: Essentially make [`envAgent_broadcast_status`](https://github.com/FLAMEGPU/FLAMEGPU2/blob/subModels/src/flamegpu/stock/subModels/SingleAgentDiscreteMovement.cu#L25) and [`movingAgent_request_move`](https://github.com/FLAMEGPU/FLAMEGPU2/blob/subModels/src/flamegpu/stock/subModels/SingleAgentDiscreteMovement.cu#L40) parametrisable by a caller. Rest of movement functionality opaque and handled internally by submodel.

fleming0:
Just to add, here's the MWE that illustrates how a different subModel API could be called. It followed naturally while porting the **Segregation** model from NetLogo to FLAME.

#### Existing API
The current Segregation movement subModel is ~320 LoC and uses logic inside the `movement_request()` function to move to cells that have more neighbours of type `A` or `B` depending on the current agents type.

```python
...
    # if occupied then look for empty cells with a good number of neighbours
    if status == pyflamegpu.environment.getPropertyInt(
        "agent_status_movement_unresolved"
    ):
        for current_message in message_in.wrap(agent_x, agent_y):
            # if location is unoccupied then check for empty locations
            if current_message.getVariableInt(
                "status"
            ) == pyflamegpu.environment.getPropertyInt("agent_status_unoccupied"):
                # if the neigbour level at current location is better than currently stored then update
                message_env_neighbour_level = -1
                if type == pyflamegpu.environment.getPropertyInt("A"):
                    message_env_neighbour_level = current_message.getVariableInt(
                        "A_neighbours"
                    )
                elif type == pyflamegpu.environment.getPropertyInt("B"):
                    message_env_neighbour_level = current_message.getVariableInt(
                        "B_neighbours"
                    )
                message_priority = pyflamegpu.random.uniformFloat()
                if (message_env_neighbour_level > best_neighbour_level) or (
                    message_env_neighbour_level == best_neighbour_level
                    and message_priority > best_neighbour_random
                ):
                    best_neighbour_level = message_env_neighbour_level
                    best_neighbour_random = message_priority
                    best_location_id = current_message.getVariableInt("location_id")
...
```

The logic above also requires a custom message type, which is defined in the subModel:

```python
message = movement_model.newMessageArray2D("cell_status")
message.newVariableID("location_id")
message.newVariableInt("status")
message.newVariableInt("A_neighbours")
message.newVariableInt("B_neighbours")
message.newVariableInt("type")
message.setDimensions(segregation_common.DIM, segregation_common.DIM)

message = movement_model.newMessageArray2D("movement_request")
message.newVariableID("location_id")
message.newVariableID("agent_id")
message.newVariableInt("type")
message.setDimensions(segregation_common.DIM, segregation_common.DIM)
```

This requires a modeller to define their own movement subModel or alter an existing one to successfully resolve movement conflicts across the 2D grid. The alternative API would allow modellers to pass the custom message type and cell selection function _to the existing subModel_, which may look like:

#### Alternative API
```python
...
cell_status_msg = model.newMessageArray2D("cell_status")
cell_status_msg.newVariableID("location_id")
cell_status_msg.newVariableInt("status")
cell_status_msg.newVariableInt("A_neighbours")
cell_status_msg.newVariableInt("B_neighbours")
cell_status_msg.newVariableInt("type")
cell_status_msg .setDimensions(segregation_common.DIM, segregation_common.DIM)

agent_status_msg = model.newMessageArray2D("agent_status_msg")
agent_status_msg.newVariableID("location_id")
agent_status_msg.newVariableID("agent_id")
agent_status_msg.newVariableInt("type")
agent_status_msg.setDimensions(segregation_common.DIM, segregation_common.DIM)

@pyflamegpu.agent_function
def pickCell(
    message_in: pyflamegpu.MessageArray2D, message_out: pyflamegpu.MessageArray2D
):
    # Logic to make a movement request to a cell with the highest number of neighbouring cells of the same type
    return pyflamegpu.ALIVE

# Configure cell selection fn
agent_cell_selection_fn = buyer.newRTCFunction(
    "agent_cell_selection_fn", pyflamegpu.codegen.translate(pickCell)
)

agent_cell_selection_fn.setMessageInput("cell_status_msg")
agent_cell_selection_fn.setMessageOutput("agent_status_msg")

submodel = pyflamegpu.SingleAgentDiscreteMovement()
submodel.setCellBroadcastMessage(cell_status_msg)
submodel.setAgentBroadcastMessage(agent_status_msg)
submodel.setAgentCellSelectionFunction(agent_cell_selection_fn)

move_sub_desc = submodel.addSingleAgentDiscreteMovementSubmodel(
    model, segregation_common.DIM, segregation_common.DIM
)
```

The latter does not require the modeller to create a bespoke subModel to implement discrete movement. Instead, the caller supplies the core decision logic that is needed for a particular behaviour and the messages that facilitate it.

